PR #53, opened 2014-11-05
Add json_encode()ded endpoint to PHP releases list + content type adjustments

marcioAlmada:
We usually warn PHP users to avoid unserialize remote or untrusted data but at the same time php.net only offers a serialized endpoint for PHP releases. This commit:

- adds a JSON endpoint that can be an option to avoid the serialized endpoint
- sets a proper Content-Type response header for both serialized - `text/plain` - and json - `application/json` - endpoints
- simplifies URLs exchanging `&serialize=1` and `&json=1` for `&serialize` and `&json` respectively

Cheers!

### DIFF
--- a/releases/index.php
+++ b/releases/index.php
@@ -5,6 +5,7 @@ include_once $_SERVER['DOCUMENT_ROOT'] . '/include/prepend.inc';
 include_once $_SERVER["DOCUMENT_ROOT"] . "/include/branches.inc";
 
 if (isset($_GET["serialize"])) {
+	header('Content-type: text/plain');
 	$RELEASES[5][$PHP_5_4_VERSION]["date"] = $PHP_5_4_DATE;
 	$RELEASES[5][$PHP_5_3_VERSION]["date"] = $PHP_5_3_DATE;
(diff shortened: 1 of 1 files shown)


salathe:
At the time of writing, the only mirror that doesn't have ext/json is ru2.php.net. :joy_cat:

releases/index.php
```diff
@@ -57,6 +58,61 @@
...
+		$array = array();
+		foreach($RELEASES as $major => $release) {
+			list($version, $r) = each($release);
+			$r["version"] = $version;
+			$array[$major] = $r;
+		}
+		echo json_encode($array);
+	}
+	return;
+}
```
bjori: It seems very weird to me that the entire block is copied and only afaict 5 lines changed.
Can't this be done in a better way?
For example assign the array that is supposed to be encoded to a $output variable, and then if/else on the output format, issue the correct header and echo the correct serialization once.

bjori:
+1 on merging this (after simplifying that code a little) and adding a link to /releases/active.php from the sidebar too

bjori:
Any updates?

bjori:
Thanks!

PR #58, opened 2015-01-20
Fix potential file-include vulnerability

ircmaxell:
Fix potential file-include vulnerability by adding EXTR_SKIP to extract so it doesn't overwrite $params array.

WARNING: I have only proved this correct, I have not tested it


salathe:
The `print_view()` function isn't (currently) being used anywhere, but better safe than sorry. :+1:

PR #59, opened 2015-01-23
Update cached.php

0x90kh4n:
To show other files with the extension. Otherwise it will appear in the source code files .php and .inc.

### DIFF
--- a/cached.php
+++ b/cached.php
@@ -47,6 +47,8 @@ if (substr($abs, -3) == ".js" || substr($abs, -5) == ".json") {
     header("Content-Type: application/javascript");
 } elseif (substr($abs, -4) == ".css") {
     header("Content-Type: text/css");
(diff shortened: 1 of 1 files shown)


cached.php
```diff
@@ -47,6 +47,8 @@
     header("Content-Type: application/javascript");
 } elseif (substr($abs, -4) == ".css") {
     header("Content-Type: text/css");
+}else{
+    header("Location: index.php");
```
salathe: If you want to redirect to index.php, there's no need to fall through to the readfile() below. Might as well `exit;` after sending the header.

salathe:
Being able to get the source code of *.php, *.inc, etc. files is by design. See the comment at the very top of the file.

bjori:
As @salathe said, this is intentional

PR #61, opened 2015-02-25
Added some validation calls for the createNewsEntry

Stelian:
Script now checks if the category is valid on submission.
Script now does a warning based check for the image size to avoid high-res images being pushed.

### DIFF
--- a/bin/createNewsEntry
+++ b/bin/createNewsEntry
@@ -53,19 +53,45 @@ $categories = array(
 );
 $confs = array(2, 3);
 
+$imageRestriction = array(
+	'width' => 400,
+	'height' => 400
+);
+
 do {
+	$catVerified = false;
 	fwrite(STDOUT, "Categories:\n");
 	foreach($categories as $n => $category) {
 		fprintf(STDOUT, "\t%d: %s\t [%s]\n", $n, key($category), current($category));
 	}
 	fwrite(STDOUT, "Please select appropriate categories, seperated with space: ");
 
-	$cat = explode(" ", rtrim(fgets(STDIN)));
+	$catInput = (string) rtrim(fgets(STDIN));
 
-	if ($cat) {
+	if (0 !== strlen($catInput)) {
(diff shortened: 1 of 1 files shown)


bjori:
looks good to me.

One thing that as annoyed over the years is if the markup is not xhtml it'll automatically save it as html, rather then returning "this was html, not xhtml, sure you want to continue? Or would you rather repaste the markup?"

Stelian:
@bjori would treat that in a different pull since that is a bit more broken, in terms of it saves an empty file for invalid html, which is a pain to deal with.

On top of that, I have started a discussion in the emailing list on how to deal with cfp and conference text/date so it might be the case that some extra changes are required.

bjori:
(its not an empty file that gets saved, the <content> element becomes type="html" and tags escaped - we just strongly preferred it to be xhtml).

Feel free to merge this yourself

bjori:
@Stelian seems you never merged this?

Stelian:
@bjori I totally forgot about this PR :) Does it still make sense to merge? Amazingly enough it has 0 conflicts since April

salathe:
Bump. Let's get this merged in, if you want to do the honours?

Stelian:
Hi peter,

I am currently on a ski trip until Monday so that's the earliest I can. If you merge it yourself, would you please change the validation for image to something more sensitive like 250 pix as per previous discussions?

Regards,
Stelian

EliW:
Hey @Stelian - I was looking at potentially merging this myself, but got mentally stuck on the image size discussion.   I think there is a catch-22 here, as if you look at lots of the existing images that are used, a number of conference logos are horizontal in nature.  And restricting to 250x250 disadvantages those.   At the same time. a 400x400 is gigantic.

What if we have a more strict vertical, but allow more horizontal.  Such as 400x250 or 350x250.  I think that would be a good compromise there.

(And I assume that your ski trip is over, so perhaps you can make the change and merge it yourself *grin*)

PR #63, opened 2015-03-06
Better handling for deprecated functions

Sobak:
This pull request introduces better handling for functions deprecated in our manual. It is a direct implementation for the mockup proposed in https://bugs.php.net/bug.php?id=66861

**Complementary pull request for PhD is [here](https://github.com/php/phd/pull/12)**

eh1d9ew

It's not the prettiest implementation I can dream on, but I suppose that is easier to discuss on something almost ready and improve it, than write it from scratch.

In this case, PhD injects `$TOC_DEPRECATED` alongside with `$TOC`, when generating files for PHP.net. I decided to store deprecation state in `versions.xml`, because it seemed to be easier approach. Example of deprecated function entry:

```xml
<function name='session_name' from='PHP 4, PHP 5' deprecated='PHP 7.0'/>
<!-- just an example ;) -->
```

In fact, value of the `function[deprecated]` is currently unused. All about script cares is whether attribute exists or not, so we can use simple `function[deprecated="true"]`. However, I think that storing a proper version number there, makes implementing more advanced features easier.

I would be very glad if at least one person could **test it**. Just set deprecated attribute for some functions in manual sources, `configure.php` it, render with patched PhD and test if everything behaves well. I did testing, of course, but I cannot guarantee that I covered all scenarios.

Changes in `versions.xml` for manual are not ready. I don't have a time to find every sigle deprecated function, but I'll try to provide initial changes.

Ah, one implementation detail: I'm worried about chapters like `book.mysql`. If `$TOC` suddenly became empty (because all menu items will land in `$TOC_DEPRECATED`), then menu won't be displayed at all. What would you propose in that case? Simply taking code, rendering deprecated functions list, out of the conditional for `$TOC`?

Thanks,
Maciej.

PS: Sorry for my Engrish.

### DIFF
--- a/include/footer.inc
+++ b/include/footer.inc
@@ -7,10 +7,11 @@
     }
 
     ?>
-<?php if (!empty($config['related_menu'])): ?>
+<?php if ($config['related_menu'] || $config['related_menu_deprecated']): ?>
     <aside class='layout-menu'>
     
         <ul class='parent-menu-list'>
+            <?php if ($config['related_menu']): ?>
             <?php foreach($config['related_menu'] as $section): ?>
             <li>
                 <a href="<?php echo $section['link']; ?>"><?php echo $section['title']; ?></a>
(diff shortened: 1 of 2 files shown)


bjori:
The mysql deprecation chapter should likely be folded back into the normal structure and use this functionality.
The deprecation chapter was a workaround.

I think its a brilliant idea to store this info in version.xml

Sobak:
@bjori: Hey, I think that I made it unclear. I mean situation where all items of the contextual menu are marked as deprecated. Then `$TOC` will be an empty array and all elements will be stored in `$TOC_DEPRECATED`.

Currently, when `$TOC` is empty, page won't have related menu at all. And because loop generatiing deprecated menu is nested inside this condition, it means also no deprecated menu. My first thought is just to pull the loop out of the condition, but maybe there are better ways?

I hope it's more clear this time, but I have difficultiies understaing my gibber myself :)

include/footer.inc
```diff
@@ -32,6 +32,19 @@
     
             </li>
             <?php endforeach; ?>
+
+            <?php if (!empty($config['related_menu_deprecated'])): ?>
```
bjori: make the root condition here if ($config["related_menu"] || $config["related_menu_deprecated"]) {}
Then if ($config["related_menu"]) {} block just before the foreach like you are doing here

PR #65, opened 2015-03-25
Add a new /download/filename.ext method that is wget friendly

scottchiefbaker:
This patch addresses https://bugs.php.net/bug.php?id=69292 by creating a new routing method in **error.php** that handles file downloads by serving up files directly using `http://www.mirror.com/download/filename.ext`. This preserves all the original download methods, so existing links will **not** be affected.

This patch also remaps the file download page to give out the new `/download/` style links.

wget downloads will now output the correct filename.

### DIFF
--- a/error.php
+++ b/error.php
@@ -131,6 +131,26 @@ if (preg_match("!^get/([^/]+)$!", $URI, $what)) {
     }
 }
 
+// Method of downloading files that is wget friendly
+// This method serves requests for /download/filename.ext from THIS mirror
+if (preg_match("|^download/(.+?)$|", $URI, $match)) {
+    $filename = $match[1];
+
+    include $_SERVER['DOCUMENT_ROOT'] . "/include/do-download.inc";
+    $filename = get_actual_download_file($filename);
+
+    if ($filename) {
+        // Found the file, so we do a redirect to the file directly
+        status_header(200);
+        download_file($MYSITE, $filename);
+    } else {
+        // File missing on the server
+        status_header(404);
+        include $_SERVER['DOCUMENT_ROOT'] . "/include/get-download.inc";
+    }
+
+    exit;
+}
 
 // ============================================================================
 // Nice URLs for download files, so wget works completely well with download links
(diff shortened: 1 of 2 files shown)


scottchiefbaker:
Any comments on this?

bjori:
This feels like it could break some absurd things.
We might as well take the bullet and fix the url properly (and keep these /from/foo/bar/mirror for bc)....
Unsure what do here really.

scottchiefbaker:
I tried really hard with this page to NOT break backwards compatibility. This introduces a new download method, and maintains the old download method. I tried to go for the best of both worlds. What are you concerned it would break?

scottchiefbaker:
Any updates on this? It really should be 100% backwards compatible.

bjori:
this is not backwards compatible no.

from/this/mirror is no longer support at all -- which is a huge bc break -- why do you think its 100% backwards compatible?

scottchiefbaker:
`/from/this/mirror` is totally still in the code:

https://github.com/scottchiefbaker/web-php/blob/master/error.php#L157

I purposed did not touch any of that code. I simply added a NEW method to handle wget friendly links:

https://github.com/scottchiefbaker/web-php/blob/master/error.php#L136

It might make more sense to put the new method in the other download method but the code already had a ton of options/exceptions so I separated it.

scottchiefbaker:
Where do we stand on this? I just did a wget of the new PHP release and I got: 

```
2017-04-13 14:06:22 (42.0 MB/s) - ‘mirror’ saved [15011816/15011816]
```

a `mirror` file instead of the `php-7.1.3.tar.bz2` that would be more appropriate?.

scottchiefbaker:
@bjori I just did wget on a new version of PHP and I get a "mirror" instead of `php-7.1.6.tar.bz2` file again because we haven't fixed this bug yet. Can you comment on the status of this please?

petk:
I haven't tested this but this would be very useful new feature for php.net indeed...

scottchiefbaker:
@bjori where do stand on this? Issue has been open for three years now.

php-pulls:

Let's put this one to bed.

scottchiefbaker:
@php-pulls I'm not sure what your comment means. Is this issue getting addressed? Or is this a `WONTFIX`.

salathe:
Won't fix.

scottchiefbaker:
@salathe can you elaborate on why this isn't getting fixed?

PR #67, opened 2015-04-17
Add sha256 hashes to the download page.

natmchugh:
MD5 is broken and should not be used anymore. I have added SHA-256 hashes to the current release info and downloads page.

In the future it would be better to remove the MD5 hashes but for now both can exist on the downloads page.

### DIFF
--- a/downloads.php
+++ b/downloads.php
@@ -61,6 +61,7 @@ site_header("Downloads",
           <?php download_link($rel['filename'], $rel['filename']); ?>
           <span class="releasedate"><?php echo date('d M Y', strtotime($rel['date'])); ?></span>
           <span class="md5sum"><?php echo $rel['md5']; ?></span>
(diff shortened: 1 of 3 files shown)


bjori:
@smalyshev @Tyrael @jpauli @johannes - You guys would have to maintain this... What do you guys think?

jpauli:
I have no problem in switching, it is rather easy to get a sha256 on modern Unixes; simply don't forget to ping us (RMs) when this is done, so that we won't forget, and we will patch our RM process notes to reflect the change

Tyrael:
I'm fine with it.

bjori:
Done. Also updated the release procedures.

bjori:
Thanks @natmchugh

jpauli:
Thank you people :-)

PR #68, opened 2015-04-19
fix for "scroll to page anchor"

WesNetmo:
...and starting an attempt to clean this file.

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -10,40 +10,6 @@ String.prototype.toInt = function()
   return parseInt(this);
 }
 
-/* {{{ Hash Scroll */
-var jHtmlBody = $('html, body');
-function scroll(target, interval)
-{
-  if (typeof interval == 'undefined') {
-    interval = 400;
-  }
-
-  var hash = target;
-  target = $('[id="' + target.substring(1) + '"]');
-  
-  var offsetTop = 
-    target.offset().top 
-    - 52
-    - target.css('margin-top').toInt()
-  ;
-
-  jHtmlBody.animate(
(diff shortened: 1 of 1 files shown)


morrisonlevi:
Thanks for looking into this. I am not sure about some of the details; the `callbackCalled = true;` looks shady. I'm also not completely sure that this will prevent the browser's default scroll in some cases. I'll try and find some time to play with this to see how it works and if I can it to break.

WesNetmo:
yes, because i don't know why $('html, body') was used. maybe for compatibility with some old browser. that causes the callback to be executed twice which is not what we want so for now i just prevented that to happen. clearly it's a temporary solution till i can investigate more (it's not different from previous implementation anyway)

js/common.js
```diff
@@ -376,19 +410,7 @@ $(document).ready(function() {
         var $this = $(this);
         $this.append("<a class='genanchor' href='#" + $this.attr('id') + "'> ¶</a>");
     });
-
-    // Bind events for #[id]
-    $(window).on('hashchange', function(e) {
-      e.preventDefault();
```
WesNetmo: I've removed this because the hashchange event is not cancelable. It couldn't have ever worked

js/common.js
```diff
@@ -376,19 +410,7 @@ $(document).ready(function() {
...
-      return false;
-    });
-
-    $('a[href^=#]').click(function(e) {
```
WesNetmo: also this was pretty much pointless, since `a[href^=#]` matches `<a href="#foo">` but not `<a href="foo.php#foo">`

PR #69, opened 2015-05-13
fix blacklisted() to treat $mosquitoes as literals (not regexps)

cmb69:
Currently, the blacklisted email addresses are treated as regular expression, but it appears that they are not written as such. Especially dots (.) in values of the $mosquitoes array might give false positives.

An alternative to this fix would be to properly escape the $mosquitoes. Not sure what's best.


bjori:
https://github.com/php/web-php/commit/5cda0b8979ad97fe93cc13905243281b44215332

Thanks!

PR #71, opened 2015-05-28
Adding in latest conferences

EliW:
Finally got time to do this (was away running a conference).   Added in the two most recent conference requests from webmaster list, plus php[world] Call for Papers


bjori:
Thanks!

PR #76, opened 2015-08-31
added IPC 2015 conference announcement news item inclduing image

rmunsky:
Hi,

i forked the repository, called createNewsEntry, added all data and an image, added, commited and pushed.

Please have a look.

Would be great if we could get this online.

Thanks a lot in advance,

Robert


bjori:
It looks like the content itself wasn't valid xml (maybe and & rather then &amp;?).
Could you try again?

You can confirm it is working if the <content type=""> is xhtml, and the content within it is not html encoded.

rmunsky:
if i do xmllint on the files i added or changed i got no error. i added <div xmlns="http://www.w3.org/1999/xhtml">. 

i will commit and push and create another pull request

rmunsky:
pls have a look again.

bjori:
I understand you don't get any error using xmllint. That is because the entire content is xml encoded as you can see.

When this hits the website, it won't decode the entities so the actual announcements will be a one line blurb with &a; &lot; of &entities; &and; &no; &new; &lines; &or; &links;

rmunsky:
so... now its fine i guess. its plain html now, the xml inside the content is not encoded anymore. sorry for the confusion.

bjori:
merged, thanks

sandsmedia2016:
Hi, 

as my colleague Robert already asked, it would be great if could release our event entry for the upcoming International PHP conference on your site. We already created a pull request a few days ago.

Thanks!

PR #82, opened 2015-10-09
Add links to event and cfp

heiglandreas:
Currently when a news-item is created for a conference with a CfP there is no way of telling whether the given date and link refer to the conference or the CfP. To be able to use the information provided here as source for f.i. callingallpapers.com that isn't as nice as it could be.

I therefore altered the script to create a new news-entry slightly to query informations for Conference and CfP separately and write them to the XML-file in separate ways to be able to distinguish between the two.

This adds the following new XML-Elements:
- default:conferenceStart - holds the date the conference starts
- defaul:cfpEnd - holds the date the cfp ends
- default:cfpwebsite - holds the URL to the CFP-site

Along with that there can be now two XML-Tags `default:finalTeaserDate` containing the Conference-Start-Date and the CfP-End-Date respectively. 

If that is a problem ping me and I'll see to a fix.

### DIFF
--- a/bin/createNewsEntry
+++ b/bin/createNewsEntry
@@ -97,6 +97,19 @@ if ($conf) {
 		fwrite(STDERR, "I told you I would run it through strtotime()!\n");
 	} while(1);
 	$item->appendChild($dom->createElementNs("http://php.net/ns/news", "finalTeaserDate", date("Y-m-d", $t)));
+
+	do {
+		fwrite(STDOUT, "What's the URL of the conference/cfp-Website? ");
+		$i = trim(fgets(STDIN));
+		$u = filter_var($i, FILTER_VALIDATE_URL);
+		if (false !== $u) {
+			break;
+		}
+
+		fwrite(STDERR, "That doesn't seem to be a valid URL");
+	} while(1);
+	$item->appendChild($dom->createElementNs("http://php.net/ns/news", "website", $u));
+
 } else {
 	$href = $BASE . "/index.php";
 }
--- a/bin/createNewsEntry
+++ b/bin/createNewsEntry
@@ -89,6 +89,34 @@ foreach($cat as $keys) {
 }
 
 $href = $BASE . "/index.php";
+
+if ($cfp) {
+    /* /cfp news item */
+    do {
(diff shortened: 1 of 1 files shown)


Stelian:
Hi, @heiglandreas, not sure if the news parser can deal with this or not, maybe @bjori can confirm this?

heiglandreas:
@Stelian Where is that news-parser located at? I can have a look at it and probably either fix the parser or my code :wink:

bjori:
http://git.php.net/?p=web/master.git;a=blob;f=scripts/pregen_news;h=3369e812ddf5a65fce89fafe0d93d36a93912160;hb=HEAD

heiglandreas:
Thanks @bjori!

heiglandreas:
@bjori @Stelian I've just tested the news-parser with the following little snippet and it didn't break.

```php
#!/usr/local/bin/php -q
<?php /* vim: set noet ts=2 sw=2 ft=php: : */
if (!$argv[1]) die("usage: $argv[0] directory\n");
$root = $argv[1];
include "scripts/pregen_news";
pregen_atom("$root/archive/archive.xml", "$root/feed.atom", "$root/include/pregen-news.inc");
```

 The only drawback is that currently the ```finalTeaserDate``` will be taken from the CfP if both CfP **and** conference have been selected. I'll see to it that only **one** ```finalTeaserDate``` will exist in the events XML-file which will solve that. I'll take the later date (which will be the conference-date)

salathe:
-1 on this change. We should be publishing separate articles for calls for papers, and conference advertisements.

heiglandreas:
@salathe +1 on that. But from what I've seen so far that's not what happens in reality. The alternative would be to remove the possibility to create news-items with cfp **and** conference.

Which way to go?

heiglandreas:
I've just added a different way to target the issue of inaccurate CfP and Conference-informations in *one* news-entry in PR #83

heiglandreas:
Anyone around caring to tell me what I can do to either get this merged? Are there any informations missing from my side?

Stelian:
@heiglandreas I agree with @salathe on this one, so the other one makes more sense, would close this one

php-pulls:
**Comment on behalf of stelianm at php.net:**

Closed in favor of #83

PR #83, opened 2015-10-09
Removes possibility to add CfP AND conference

heiglandreas:
This PR eliminates the possibility to add a conference AND a CfP in one news. If you want to create both you will then have to create two separate news. This makes it easier to get informations regarding CfP
and Conferences separate from one another

It is a different way to target a better usability of CfP and Conference-Data from the php.net-Website than described in PR #82

### DIFF
--- a/bin/createNewsEntry
+++ b/bin/createNewsEntry
@@ -49,7 +49,7 @@ $categories = array(
 	array("frontpage"   => "PHP.net frontpage news"),
 	array("releases"    => "New PHP release"),
 	array("conferences" => "Conference announcement"),
-	array("cfp"         => "Call for Papers"),
+	array("cfp      "   => "Call for Papers"),
 );
 $confs = array(2, 3);
 
@@ -61,11 +61,18 @@ do {
 	fwrite(STDOUT, "Please select appropriate categories, seperated with space: ");
 
 	$cat = explode(" ", rtrim(fgets(STDIN)));
+	$cat = array_intersect(array_keys(array_values($categories)), $cat);
 
-	if ($cat) {
+	$intersect = array_intersect(array(2,3), $cat);
+	if ($cat && (count($intersect) < 2)) {
 		break;
 	}
-	fwrite(STDERR, "You have to pick at least one category\n");
+
+	if (count($intersect) == 2) {
+		fwrite(STDERR, "Pick either a CfP OR a conference\n");
+	} else {
+		fwrite(STDERR, "You have to pick at least one category\n");
+	}
 } while(1);
 
 


bin/createNewsEntry
```diff
@@ -49,7 +49,7 @@ $categories = array(
 	array("frontpage"   => "PHP.net frontpage news"),
 	array("releases"    => "New PHP release"),
 	array("conferences" => "Conference announcement"),
-	array("cfp"         => "Call for Papers"),
+	array("cfp      "   => "Call for Papers"),
```
bjori: That doesn't look right?

heiglandreas: That looks ugly but it causes the description of the CfP-Category to be alligned like the other categories in the script like this:

```bash
$ bin/createNewsEntry
Trying archive/entries/2015-10-16-1.xml
Please type in the title: my cool feature
Categories:
	0: frontpage	 [PHP.net frontpage news]
	1: releases 	 [New PHP release]
	2: conferences	 [Conference announcement]
	3: cfp       	 [Call for Papers]
Please select appropriate categories, seperated with space:
```

before it looked like this:

```bash
$ bin/createNewsEntry
Trying archive/entries/2015-10-16-1.xml
Please type in the title: My cool feature
Categories:
	0: frontpage	 [PHP.net frontpage news]
	1: releases 	 [New PHP release]
	2: conferences	 [Conference announcement]
	3: cfp  	 [Call for Papers]
Please select appropriate categories, seperated with space:
```

That's the only reason for this change and as the key is not used somewhere else I though it might be a way to align that nicely.

Shall I revert it?

heiglandreas:
Anyone around caring to tell me what I can do to either get this merged? Are there any informations missing from my side?

Stelian:
@heiglandreas this is a rather old one, does it still make sense to merge and does it actually work with the rendering?

heiglandreas:
This PR doesn't change the output, it only prohibits people from selecting two options at once. So rendering shouldn't be an issue here. In my eyes it makes sense to merge it. 

As it doesn't prevent people from manually editing the resulting XML-file it might not necessarily fix the issue of people submitting entries with CfP *and* Conf-entry

Stelian:
To be fair I don't think people would go that far, they just go with what the tool generates for them, no more no less, so might be helpful.

EliW:
Given that the code all looks good to me, and we are having issues now of people using the tool and, in fact, choosing both options.   I'm going to go ahead and commit this :)

EliW:
Well, nevermind :)  Guess I can't.  While I have commit access to the /archive/ directory, I guess I don't to /bin/ as it rejected my push because of not having karma :-/   Ah well.  Sorry @heiglandreas - Hey @Stelian can you merge this yourself?  (and/or give me karma to do so?)

salathe:
>  (and/or give me karma to do so?)

@Stelian cannot but I can, and [have done so][1].

[1]: http://svn.php.net/viewvc?view=revision&revision=342137

EliW:
Thanks!   And it's been merged!

PR #86, opened 2015-11-03
Remove the link to the EOL versions download page

ThibaudDauce:
Hi,

This PR is just a proposition to remove the link to the download page for EOL versions in the supported versions matrix. Tell me if you think it's a good idea, if not you can close this PR :-)

I personaly think it's important to show people it's not a good idea to download this version and they should consider an other option.

I'm new to this repo so I'm not sure if it's the good way to do it (this kind of condition in HTML).

Thanks,

### DIFF
--- a/supported-versions.php
+++ b/supported-versions.php
@@ -51,9 +51,13 @@ site_header('Supported Versions', array('css' => array('supported-versions.css')
 				?>
 				<tr class="<?php echo $state ?>">
 					<td>
-						<a href="/downloads.php#v<?php echo htmlspecialchars($release['version']) ?>">
+						<?php if ($state == 'eol'): ?>
 							<?php echo htmlspecialchars($branch) ?>
-						</a>
+						<?php else: ?>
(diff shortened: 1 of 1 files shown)


LawnGnome:
The patch itself looks good (thanks!), but I'm going to close this without merging.

While I agree with the principle of discouraging the use of obsolete versions, I think a user who's at the EOL page and wants to download an old version should already know they're shooting themselves in the foot, and I'd rather not remove the ability altogether, since it can be useful for people trying to track down issues.

ThibaudDauce:
I perfectly agree with that :-)

There's still the "Download" button at the top of the page, I just removed the quick link to EOL version. But I understand your point for tracking down issues.

I just checked the http://php.net/supported-versions.php page and the 5.4 line is not even there. Was it remove from `get_active_branches()`? It's even better than my PR :-)

LawnGnome:
@ThibaudDauce Yep, I removed 5.4 from the active releases array, since it's no longer active. (Usually happens a couple of months after the final release of a branch.)

php-pulls:

Adam never got around to closing this.

PR #93, opened 2016-02-11
blacklisted "Sofie Vanderplaetsen" due to repeated spamming

cmb69:


salathe:
I don't believe this code gets exercised when mailing a list directly; though I could be wrong!

cmb69:
You're probably right, Peter. I merely assumed that this is the right place, but on further rethinking the mailing list is likely blacklisted via DB.

PR #95, opened 2016-02-16
Midwest php 2016

jsundquist:
Fixing the midwest php logo that is failing to display on php.net


php-pulls:

Fixed in 7ce66174b1

PR #97, opened 2016-03-21
Add ability to to download latest PHP7 stuff using PHP7-LATEST

janaurka:
This was already supported for PHP4-LATEST and PHP5-LATEST as noted in [Bug 70946](https://bugs.php.net/bug.php?id=70946). The PR simply adds the functionality for PHP7-LATEST.

### DIFF
--- a/error.php
+++ b/error.php
@@ -137,7 +137,10 @@ if (preg_match("!^get/([^/]+)$!", $URI, $what)) {
 if (preg_match("!^get/([^/]+)/from/([^/]+)(/mirror)?$!", $URI, $dlinfo)) {
     
     $df = $dlinfo[1];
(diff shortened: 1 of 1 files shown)


error.php
```diff
@@ -135,29 +135,32 @@
 // ============================================================================
 // Nice URLs for download files, so wget works completely well with download links
 if (preg_match("!^get/([^/]+)/from/([^/]+)(/mirror)?$!", $URI, $dlinfo)) {
-
```
salathe: Can you remove these whitespaces-only changes from the patch?

janaurka: Sure, done. Check the new Commit https://github.com/php/web-php/pull/97/commits/86df840901706328f33a7cd46613f1dda5c21971

PR #100, opened 2016-03-31
Make release dates consistent.

schlessera:
When parsing the changelog to extract version numbers and release dates, I noticed that some release dates were "text-only" and did not include the `<time>` element.

This patch makes the release dates consistent across all releases, in the format `<b><?php release_date($date); ?></b>`.

### DIFF
--- a/ChangeLog-5.php
+++ b/ChangeLog-5.php
@@ -370,7 +370,7 @@ site_header("PHP 5 ChangeLog", array("current" => "docs", "css" => array("change
 
 <section class="version" id="5.6.13"><!-- {{{ 5.6.13 -->
 <h3>Version 5.6.13</h3>
-<?php release_date('03-Sep-2015'); ?>
+<b><?php release_date('03-Sep-2015'); ?></b>
 <ul><li>Core:
 <ul>
   <li><?php bugfix(69900); ?> (Too long timeout on pipes).</li>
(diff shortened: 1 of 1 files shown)


cmb69:
Thanks for the PR, Alain!

Of course, all release dates should be processed with `release_date()` instead of being written verbatim, but I'm not sure about the `<b>` – wouldn't it be better to make that a child of the `<time>` element (IOW: put it into `release_date()`)?

schlessera:
I had only continued the way it was done for most of the entries without questioning the approach.

But putting the responsibility of adjusting the font weight into the `release_date()` function would certainly make it easier to change the formatting later on.

I can think of two possible ways of making this consistent across the board:

1. Warp the `<time>` tag with `<b></b>` inside the `release_date()` function.
2. Add `.releasedate { font-weight: bold; }` to the stylesheet.

I would opt for the second one.

Do you want me to update and adapt the pull request to include that change (or number 1. if you prefer)?

cmb69:
I agree that the CSS approach is preferable; should probably go to [changelog.css](https://github.com/php/web-php/blob/master/styles/changelog.css#L12-L16).

It would be great if you can update and adapt the PR.

schlessera:
This is superseded by #129, because I stupidly forgot to create a branch for the pull request.

PR #105, opened 2016-06-02
Postgres is a Fraudian slip for MySQL or generic DB(?)

PallHaraldsson:


PallHaraldsson:
@weltling You put "Postgres" in. I'm familiar with it but less with MySQL or at least the naming convention.

Use Mysqlnd (I'm not clear on what this ending or other are for), or just Mysql or something else (but not Mysqli)?

PDO takes care of database abstraction; I'm alost 100% sure Mysql* functions have nothing to do with MySQL.

MySQL is the name, and similarly PostgreSQL (while Postgres used and should ave stayed that way..), and it seems your choice not to use the official capitalization[s].

weltling:
@PallHaraldsson the ticket causing the behavior change was #71820, which is about ext/pgsql. It affected both etx/mysqli and ext/pgsql. Thus, the reverting info is put under the same section.

Probably don't care much about the camel cases, it's used both ways in the NEWS :)

Thanks.

PallHaraldsson:
Yes, #71820 has to do with PostgreSQL. Search did not find "Mysql" there.. The relation to #72151 isn't obvious to me.

In ChangeLog-5.php I only reclassified #72151, correctly I think. In the other it seems not right to group together, while I guess should be handled the same in both changelogs?

I do not care about spelling too much. I do care about PostgreSQL, am not really a PHP (or MySQL) programmer.

Fell free to merge this change (or do some more changes) or close if I wasn't helping fixing what seemed like an obvious wrong to me.

weltling:
@PallHaraldsson #71820 affected *_fetch_object from both mysqli and pgsql. Change in #72151 is per fact the revert commit to #71820. I've added an explicit note that #71820 was reverted. Thanks for the alert.

Cheers.

PR #111, opened 2016-06-15
Adds a prepend link to the conf-list

heiglandreas:
That way it's easier to show people how to add a conference to the list.

For that I created a page in the [wiki](https://wiki.php.net/conferences) that the added link links to...

### DIFF
--- a/conferences/index.php
+++ b/conferences/index.php
@@ -26,7 +26,9 @@ foreach($NEWS_ENTRIES as $entry) {
         }
     }
 }
-$panels = "";
+$panels = '<p class="prepend"><a href="https://wiki.php.net/conferences">Want to see your conference appear here?</a></p>';
+
+
 foreach($frontpage as $entry) {
     $link = substr($entry["id"], 15); // Strip http://php.net/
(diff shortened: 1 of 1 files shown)


EliW:
@bjori — This looks like a great addition to me.  Any objections?

conferences/index.php
```diff
@@ -26,7 +26,9 @@
         }
     }
 }
-$panels = "";
+$panels = '<p class="prepend"><a href="https://wiki.php.net/conferences">Want to see your conference appear here?</a></p>';
```
salathe: I don't think the "prepend" class does anything, did you mean "panel"?

heiglandreas: No, I actually meant "prepend". Currently the "panel"-class only sets the bottom-margin. My idea was to have a different class to allow a different styling of the prepend link as it's not a conference we're linking to there. But currently there's no styling added.

salathe: That seems pointless, but it's a minor thing so... do what you want.

salathe:
Merged. Hopefully the wiki page will be useful.

bjori:
thanks!

PR #113, opened 2016-07-12
Move Korean translation to $INACTIVE_ONLINE_LANGUAGES

cmb69:
Apparently, the Korean translation of the PHP manual is very incomplete,
and even worse, most of what's there is out-dated,[1] so we mark the
translation as inactive for the time being.

[1] http://doc.php.net/revcheck.php?p=graph&lang=kr


Stelian:
@bjori any concern with this one?

Stelian:
As per discussions on the mailing list, this can be merged and reverted once the translation is in better shape. Unless we see concerns here, I will merge this in tomorrow.

PR #124, opened 2016-09-06
logo.php: Check if QUERY_STRING is set before using it

plstand:
Fixes a PHP Notice introduced in 0df71ff798c722f3.


cmb69:
Thanks!

PR #127, opened 2016-09-13
Fix Bug #71404 (Top page of Japanese document isn't up to date)

m-takagi:
Maybe this fixes outdated top page of Japanese document (http://php.net/manual/ja/)

see: https://bugs.php.net/bug.php?id=71014


cmb69:
Thanks!

PR #128, opened 2016-09-22
Announcing PHP UK 2017 Conference and CFP

johannac:
Adding 2 news entries to announce PHP UK Conference 2017 and the open call for papers.


EliW:
Thanks for the submission @johannac ... Though I'm unsure about the 'double submission'.  That is to say, you are putting in two news entries back-to-back, just one for the conference and one for the CfP.   Traditionally you'd just do one that mentions both (and marked as CfP submission).   And then optionally do another one once the schedule is set as a conference announcement.

As it stands, these two will just end up sitting right next to each other.

If you'd like to combine these into just one CfP+conference announcement I'd be happy to push it through.  If you'd still like them to be separate like this, I'd need to get feedback from others on the team like @bjori

salathe:
FWIW I'm happy that the CFP is split into a separate article than the conference announcement; that way they can each go into their appropriate section on the homepage.

johannac:
@EliW Could I maybe back date the conference announcement a couple months?

johannac:
@EliW Sorry just re-read your comment and I understand now - I've removed the general conference announcement and left just the CFP. Will submit the general conference announcement again in the future when the schedule is set.

cmb69:
> Though I'm unsure about the 'double submission'.

Note that there's the related PR #83?

EliW:
Merged in.  Sorry for the delay @johannac - I got swamped and then this totally slipped my mind.

EliW:
To @cmb69 — Yeah, that PR was specifically to stop someone from making 'one entry' that was listed as both CfP and Conference.   The point was to have people make two entries.  As I suggested to Johanna here.   But typically these entries are not just both blasted at the same time.  Therefore causing on the /conferences/ page the two entries to just appear right next to each other.   But 'tradition' states that you do a CfP announcement.  And then later do a Conference Announcement once you have the full schedule — Plus that works better for the conference organizers in terms of driving the appropriate traffic at the appropriate time to their website anyway. :)

cmb69:
@EliW Thanks for the explanation. Anyhow, we should either merge PR #83 or close it. :-)

BTW: is there a way to preview new archive entries?

EliW:
I agree, but IMO it's not really my place to do either of those.  I'm not involved in the day-to-day of running/supporting php.net's website.  I just volunteered to be 'Conference Boy' because I complained one time too many about it taking too long for conferences to get approved :)   So now I do it.

You'd really have to get @bjori or someone else more involved in the full website picture to get that other PR pulled.or.cancelled

PR #130, opened 2016-10-18
Add missing links to PHP 7 ChangeLogs.

schlessera:
Add missing links to PHP 7 change log to two sidebars that reference the 4 & 5 change logs:

- http://php.net/releases/
- http://php.net/sitemap.php


cmb69:
Good catch! Thanks for the PR.

PR #139, opened 2016-12-27
Added PHPKonf Istanbul PHP Conference 2017

emir:


emir:
Thank for merging @EliW. 

Our CfP also open until Jan, 31. Should I duplicate this for changing category term like this?

`+  <category term="cfp" label="Call for Papers"/>`

PR #141, opened 2017-01-17
ConFoo Montreal early bird (3 days left) and Vancouver 2017 announce

afilina:
Time-sensitive news: early bird expires this Friday. @php-pulls
Thanks.


afilina:
Bumping, as the news is about an early-bird that expires tomorrow. @php-pulls

EliW:
Hey there @afilina — Traditionally conferences only get a max of 2 submissions to the php.net/conferences page.  One announcement for their Call for Speakers, and one announcement when their schedule is published.  

They don't then make other announcements like Early Bird Sales ending.  (If we allowed those type of messages, then the conferences pages would be filled with tons of re-announcements for each conference, instead of just being a 'list of conferences'.)

It looks like ConFoo Montreal has already had a Call for Speakers announcement, and a Conference Announcement.

/cc @bjori

afilina:
I was not aware of this limitation, thanks. I will adjust our php.net communication schedule accordingly.

PR #144, opened 2017-02-21
Add CakeFest 2017

dakota:
Adding news item for CakeFest 2017


Stelian:
Merged in, should be up soon.

dakota:
Thanks 👍

Stelian:
@dakota I just saw that the image is rather big, would it be ok if we resize a bit? Covers most of the news article.

dakota:
I see so too. Sorry about that, you are welcome to resize it to a more manageable size, otherwise I could do it? Thank you.

Stelian:
@dakota, if you could submit a PR with a more balanced size that would be great, I do not want to ruin proportions or resolution or anything related.

PR #145, opened 2017-02-21
Make cakefest logo smaller

dakota:
Better logo for the CakeFest announcement


Stelian:
hi @dakota as I temporary resized the original image and literally just pushed it 3 minutes ago, could you merge master in? Apologies for the timing

dakota:
Not a problem. Just merged.

salathe:
The 250px wide version was a much better size. Anything under 300px is good, anything over *should* be made smaller; though from looking at the conferences page, reviewers must think differently.

Stelian:
@salathe afaik, there isn't an actual size recommendation, more of a common sense approach. I did add some basic validation to #61 but that aims for 400 not the 250px.

We can go with 250px as a rule from here on, and I can update the wiki page with a bit more details. What do you think?

PR #148, opened 2017-03-04
Add event Conferência PHPRS 2017

fernandodebrando:
Add event Conferência PHPRS 2017, From May 12 to 13, 2017, in Porto Alegre/RS-Brazil


Stelian:
@fernandodebrando as a general guideline we would like to keep the image sizes somewhere under 250px to not have the conferences page all over the place. Could you please resize it?

fernandodebrando:
@Stelian resized image. Thank you!

PR #154, opened 2017-04-25
Use exact word for parameter scroll #74493

JayPHP:
https://bugs.php.net/bug.php?id=74493

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -408,7 +408,7 @@ $(document).ready(function() {
     {
       var id = $(this).text().replace(/^[&$]{0,2}/g, '');
       var offsetTop = $('.parameters, .options').find('.parameter').filter(function() {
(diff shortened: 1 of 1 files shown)


morrisonlevi:
Will you please add a comment in the code that links to the corresponding bug in case the `contains` was intentional? That way when we re-discover why it uses `contains` we can properly fix both.

JayPHP:
@morrisonlevi sure, all done.

JayPHP:
Any chance of getting this merged?

peterdd:
On some pages this does not work: https://www.php.net/manual/en/mysqli.affected-rows.php

Maybe because 'mysql' === '\nmysql' is false..?

Related: https://bugs.php.net/bug.php?id=77036

parameterscroll

PR #166, opened 2017-07-21
Remove source.php from robots.txt

gxrneyme:
I've removed source.php from robots.txt as source.php no longer exists, and was still in the list from the previous version of PHP.net.


amin007:
+1

PR #168, opened 2017-08-13
Adding in the Midwest PHP 2018 call for papers announcement

jsundquist:
Announcing the Midwest PHP 2018 Conference Call for Papers


EliW:
Done, sorry for slight delay

PR #170, opened 2017-08-17
Alters email-verification

heiglandreas:
The current email-verification is completely based on regex that is notup-to-date any more in regard to newer TLDs. For example the current version checks whether the TLD is between 2 and 6 chars long which fails for ".digital" and some other longer TLDs

I've altered it so that the host-check is done by actually checking for an MX-record so host that currently validate against the regex will fail in future when no MX is available.

The Address-Part is still matched against a Regex.

### DIFF
--- a/include/email-validation.inc
+++ b/include/email-validation.inc
@@ -12,22 +12,30 @@ function clean_AntiSPAM($email)
 // Try to check that this email address is valid
 function is_emailable_address($email)
 {
-    // Exclude our mailing list hosting servers
-    $hosts_regex = "!(lists\.php\.net|chek[^\.*]\.com)!i";
-    $excluded_hosts = preg_match($hosts_regex, $email);
-
-    // Email addresses need to match this pattern
-    $email_regex = ":^([-!#\$%&'*+./0-9=?A-Z^_`a-z{|}~ ])+" .
-                   "@([-!#\$%&'*+/0-9=?A-Z^_`a-z{|}~ ]+\\.)+" .
-                   "[a-zA-Z]{2,6}\$:i";
-    
-    // If address is not under excluded hosts, and fits the regex,
-    // then we belive that it is a good email address
-    if (!$excluded_hosts && !empty($email)) {
-        return preg_match($email_regex, $email);
-    } else {
-        return FALSE;
+    // No email, no validation
+    if (empty($email)) {
+        return false;
     }
+
+    $parts = explode('@', $email);
+    // An email-address with more than one '@' can't be valid
+    if (count($parts) != 2) {
+        return false;
+    }
+
+    // addresses from our mailing-list servers
+    $host_part_regex = "!(lists\.php\.net|chek[^\.*]\.com)!i";
+    if (preg_match($host_part_regex, $email)) {
+        return false;
+    }
+
+    // When no MX-Entry can be found it's for sure not a valid email-address.
+    if (getmxrr($parts[1], $return_values) === false) {
+        return false;
+    }
+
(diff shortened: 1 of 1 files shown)


ishwon:
Thank you Andreas.
I've filed a bug report at https://bugs.php.net/bug.php?id=75080.

include/email-validation.inc
```diff
@@ -12,22 +12,30 @@ function clean_AntiSPAM($email)
...
-    // Exclude our mailing list hosting servers
-    $hosts_regex = "!(lists\.php\.net|chek[^\.*]\.com)!i";
-    $excluded_hosts = preg_match($hosts_regex, $email);
+    // No email, no validation
+    if (empty($email)) {
+        return false;
+    }
+
+    $parts = explode('@', $email);
+    // An email-address with more than one '@' can't be valid
```
cmb69: Nitpick: the local-part of an email address may contain `@`, but that may not really matter in practise.

cmb69:
At least in an ideal world, we should rely on FILTER_VALIDATE_EMAIL here.

include/email-validation.inc
heiglandreas: In an earlier version I was using ```subset($email(stripes($email, '@')+1)``` to get the domain part. Might be the better value there then…

heiglandreas:
That's right! I'll replace the last regex with a ```filter_var($email, FILTER_VALIDATE_EMAIL)```. Makes absolute sense… Perhaps even as the first check…

include/email-validation.inc
cmb69: I'm not sure if the rest of the code would be able to deal with quoted-strings, anyway. Most certainly `FILTER_VALIDATE_EMAIL` wouldn't. But we still could save the `getmxrr()` to a broken domain.

heiglandreas: It didn't work before, so I wouldn't worry about that too much. Though you are absolutely right, the question is whether we merge this soonish so more people can subscribe as soon as possible and then try to optimize the code from there…

cmb69:
> the question is whether we merge this soonish so more people can subscribe as soon as possible and then try to optimize the code from there…

IMO, this is the way to go, so I did merge. Please also have a look at the respective mailing lists/bug tracker, so that we can improve or revert if necessary.

heiglandreas:
Great! I have an eye on the bug tracker mailing-list anyway ;)

ishwon:
Hi Andreas. I tried subscribing now with the address `@lsl.digital` and the following message is displayed:

```
We were unable to subscribe you due to some technical problems.
Please try again later.
```

heiglandreas:
Sadly (or luckily?) I don't have access to the log-files of the Server to see what exactly is the issue… But from what I'Ve seen on the mailing list so far there seem to be some issues with the mailinglist - especially with the registration process. So even though the email-validation now passes the subscription to the mailing list doesn't work.

As the mailing list uses ezmlm in the background you should be able to register via email. Have a look at the last paragraph of http://php.net/mailing-lists.php or http://untroubled.org/ezmlm/ezman/ezman1.html for the details. You should be able to register by sending an email to ```php-internals-subscribe@lists.php.net``` f.e.

And perhaps @cmb69 can share some details on how to contact and get someone to fix the underlying issue…

ishwon:
Thank you. I was able to subscribe using the `ezmlm` command.

cmb69:
> And perhaps @cmb69 can share some details on how to contact and get someone to fix the underlying issue…

Hmm, mailing to internals@lists.php.net might help…

PR #173, opened 2017-09-26
Fix bug #69954

petk:
This patch fixes [bug 69954](https://bugs.php.net/bug.php?id=69954)

A minor documentation fix is also applied via edit.php.net (some duplicate `ini.ini.` prefixes removed).

### DIFF
--- a/include/errors.inc
+++ b/include/errors.inc
@@ -87,10 +87,10 @@ function status_header($num)
             @header("HTTP/1.1 $num $status");
             break;
     }
-    
+
     // BC code for PHP < 4.3.0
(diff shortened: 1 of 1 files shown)


petk:
Any chance to check this out by anyone or is there any plan to make those redirects differently?

petk:
I'm guessing this patch is too complex to be solved like this. I'll see if I can come with something simpler yet still possible to solve over this repository. In case anyone wonders the patch will probably remain available at
https://patch-diff.githubusercontent.com/raw/php/web-php/pull/173.patch

PR #177, opened 2017-09-27
cvs.php.net and cvsold.php.net are not used nor working anymore

petk:
There is a slight confusion with non working old CVS repository links mentioned in the [Tourist's Guide](http://php.net/sites.php). This patch refactors list of sites into probably more logical one. In case I've missed something, let me know. Thanks.


petk:
Reported via [bug #75399](https://bugs.php.net/bug.php?id=75399) instead and closing this pr to leave the repo maintainers come up with a better patch. Thanks.

PR #179, opened 2017-10-13
Remove non working lxr.php.net links

petk:
Hello, going through the links and text on php.net, there are mentioned links to lxr.php.net which seems to be down for quite a while now. I suggest to remove these from the [Getting involved guide](http://php.net/get-involved.php) and the [Tourist guide](http://php.net/sites.php) to not confuse reader point of view too much.

### DIFF
--- a/get-involved.php
+++ b/get-involved.php
@@ -61,8 +61,6 @@ $SIDEBAR_DATA ='
   <div class="headline">Useful links for developers:</div>
   <div class="body">
     <ul>
-     <li><a href="http://lxr.php.net">lxr</a> - an online interface to the
-       PHP source, providing search facilities useful for programmers and researchers</li>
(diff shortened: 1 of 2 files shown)


salathe:
Can you redo the diff without the whitespace changes?

petk:
Done with whitespaces added as in the current files and only the lxr contextual mentions removed.

petk:
As for the whitespaces and the basic coding style, I would also suggest to use [.editorconfig](http://editorconfig.org/) as noted in one of the previous pull request. Less things to worry about - the text showing up properly on the web and in future edits. Plus with such file we have common coding style in this repo.

petk:
Reported via [bug #75399](https://bugs.php.net/bug.php?id=75399) instead and closing this pr to leave the repo maintainers come up with a better patch. Thanks.

PR #180, opened 2017-10-13
Update PHP license 3.01 year range

petk:
This syncs the license in the php-src repository and the license folder here.


Sobak:
Thanks :)

PR #182, opened 2017-11-01
[Announcing] PHP Conference Brasil 2017

carusogabriel:
Since 2015 we announced [PHP Conference Brasil](http://phpconference.com.br/) in [PHP Conferences](http://www.php.net/conferences/). Let's make the same this year :smile:


salathe:
cc @EliW @adamculp

PR #183, opened 2017-11-01
[Refactored] createNewsEntry algorithm

carusogabriel:
Followed some [PSR-1](http://www.php-fig.org/psr/psr-1/) and [PSR-2](http://www.php-fig.org/psr/psr-2/) PHP code standard :+1:


Sobak:
Refactoring is always nice, but do not mix code style changes with the actual changes regarding logic-  such diff is almost impossible. It would be nice if you could send the PR with just code logic refactoring first

carusogabriel:
Closing this PR in favor of an upcoming with Coding Standard based on PSRs

PR #187, opened 2017-11-20
Update $.scrollTo to last version

lex111:
Fixes the bug [#75458](https://bugs.php.net/bug.php?id=75458)


cmb69:
If this has actually been tested, I'm +1.

However, php.net still uses jQuery 1.10.2 (released 2013-07-03), so the whole jQuery & plugin stuff should be updated – ideally to supported versions. jQuery 3 drops support for IE 8 and older, though, and since the site is not fully usable without JS (for instance, "Change language" does not work), that might be a problem.

lex111:
> However, php.net still uses jQuery 1.10.2 (released 2013-07-03), so the whole jQuery & plugin stuff should be updated – ideally to supported versions. jQuery 3 drops support for IE 8 and older, though, and since the site is not fully usable without JS (for instance, "Change language" does not work), that might be a problem.

This is not necessary, it is enough to update version of `scrollTo` plugin, it is compatible with the current old version of jQuery, so it will work.

lex111:
Any news?

morrisonlevi:
I now longer have a development server set up to test these changes; can anyone else verify this?

salathe:
@morrisonlevi It Works For Me™

morrisonlevi:
Thanks, @salathe.

lex111:
Hooray! :tada:

PR #190, opened 2017-11-27
Fix double dotted line in sidebar

hannesvdvreken:
See https://bugs.php.net/bug.php?id=75549

### DIFF
--- a/styles/theme-medium.css
+++ b/styles/theme-medium.css
@@ -414,9 +414,11 @@ aside.tips .panel > a:after,
 aside.tips .panel > span:after {
   content:" ";
   display:block;
-}
-aside.tips .panel > span:after {
   border-bottom:1px dotted #666;
+}
+aside.tips .panel > a {
+  display:block;
+  border-bottom: none;
 }
 aside.tips .panel > a:hover:after {
   border-color:#AE508D;
--- a/styles/theme-medium.css
+++ b/styles/theme-medium.css
@@ -414,6 +414,8 @@ aside.tips .panel > a:after,
 aside.tips .panel > span:after {
   content:" ";
   display:block;
+}
+aside.tips .panel > span:after {
   border-bottom:1px dotted #666;
 }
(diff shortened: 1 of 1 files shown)


lex111:
It's a bit not what I had in mind. I pointed out that need to remove short borders, i.e. long borders should only be displays.

| After apply patch | Expected | 
| - | - |
| image |  image

But maybe it's better, I don't know, just leaving only the short borders, it doesn't look very good:

image

hannesvdvreken:
so: the ones in a list should have short lines,
the ones under titles and links that aren't in a list: long lines.

right?

lex111:
@hannesvdvreken I think so.

salathe:
Something like the following looks better, to me.  The `border-bottom: none` line should be all that is really required to remove the extra border.

```diff
diff --git a/styles/theme-medium.css b/styles/theme-medium.css
index affadcef..8a0437b2 100755
--- a/styles/theme-medium.css
+++ b/styles/theme-medium.css
@@ -410,6 +410,9 @@ aside.tips a {
   color:#ccc;
   border-bottom:1px dotted #666;
 }
+aside.tips .panel > a {
+  border-bottom: none;
+}
 aside.tips .panel > a:after,
 aside.tips .panel > span:after {
   content:" ";
```

I'd also be temped to thrown in a `display: block` there as well, which is a helpful usability tweak to make the anchor fill the width and so be easier to click.

lex111:
@salathe agree, I tested these styles before when I did the screenshots above.

> I'd also be temped to thrown in a display: block there as well, which is a helpful usability tweak to make the anchor fill the width and so be easier to click.

It would be good :+1:

hannesvdvreken:
@salathe @lex111 like this?

lex111:
@hannesvdvreken yes exactly :+1:

salathe:
@hannesvdvreken yes, that looks good to me. 👍

PR #191, opened 2017-12-07
Array_map

Jul10l1r4:


Jul10l1r4:
.

philip:
This does not look like PHP to me; I suggest rejecting this PR.

That said, functions are defined in includes like `site.inc` rather than inline.

nikic:
Yeah, this change looks completely bogus...

PR #192, opened 2017-12-16
Clean elses

carusogabriel:
I've removed the `else`s when we have already returned something :put_litter_in_its_place:

### DIFF
--- a/include/layout.inc
+++ b/include/layout.inc
@@ -45,7 +45,8 @@ function highlight_php($code, $return = FALSE)
     ) . '</div>';
 
     if ($return) { return $highlighted; }
-    else { echo $highlighted; }
+    
+    echo $highlighted;
 }
 
 // Stats pages still need this
--- a/cal.php
+++ b/cal.php
@@ -205,11 +205,9 @@ function date_for_recur($recur, $day, $bom, $eom)
     }
     
     // ${recur}th to last $day of the month
-    else {
-        $eomd = date("w",$eom) + 1;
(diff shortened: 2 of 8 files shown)


include/layout.inc
```diff
@@ -45,7 +45,7 @@ function highlight_php($code, $return = FALSE)
     ) . '</div>';
 
     if ($return) { return $highlighted; }
-    else { echo $highlighted; }
+    $highlighted;
```
philip: Missing an echo?

philip:
Curious, did `else` harm you in another life? :)

include/layout.inc
carusogabriel: Yes, sorry, gonna fix it!

carusogabriel:
@philip No, I’m just clean them up to reduce some code and ways our code can get. It’s a kinda of refactoring ☺️

include/layout.inc
philip: In addition, ideally all patches should be void of whitespace changes as otherwise it's difficult to spot the actual changes and/or problems like the above.

carusogabriel: My IDE is doing this changes, will search how to stop it

carusogabriel:
Closing this PR in favor of an upcoming with Coding Standard based on PSRs

philip:
Ah @carusogabriel, sorry I forgot to merge this, I promise to take a quicker look at the next one instead of only taunting you about `else` ;D

PR #196, opened 2018-01-13
Use PSRs in web-php

carusogabriel:
With this PR, I want to propose the usage of [`PSR1`](www.php-fig.org/psr/psr-1/) and [`PSR2`](www.php-fig.org/psr/psr-2/) in our codebase of `web-php` :sweat_smile: 

I did one change per commit to facilitate review :+1: 

Also, passing `?w=1` to the `Files changed` tab will remove several files that only contains `Whitespace only changes` :put_litter_in_its_place:


philip:
The amount of work here is impressive but I don't think it's worth committing this huge change. It, for example, changes files last touched in the 90's for reasons like removing EOL whitespace and changing double quotes to single quotes. People will always commit double quotes for literal strings and imho that's okay.

At least, this is why I'm not merging the pull request.

carusogabriel:
@philip Sure, thanks for your feedback :smile:

Gemorroj:
Judging from the closed PR, you are extremely reluctant to accept changes. Even though the project code looks awfully outdated.
Please be more open, otherwise the project will continue to stagnate.

heiglandreas:
It's got nothing to do with reluctance to change but rather with a *lot* of changes for the sake of introducing a coding style to a "legacy" codebase.

That causes a hell of a lot of changes that - even though they might have been done automatically - need to be checked. And that takes time. Time that otherwise can't be spent on the project. 

And even though personally I absolutely favour coding standards they will not per se make the code better. They might make it more readable, but not better. 

And by enforcing a coding standard (and that's what would need to be done after a standard is applied to remove entropy) you'd reduce the number of contributors even more. So not having a coding standard actually makes the project more open, because *anyone* can contribute.

PR #213, opened 2018-03-29
Vertically center search box

jesobreira:
I don't know if it's my OCD but this was taking my attention every time I visited php.net:

captura de tela 2018-03-29 as 00 53 52

This very small PR moves the search field closer to the middle.

captura de tela 2018-03-29 as 00 54 12

### DIFF
--- a/styles/theme-base.css
+++ b/styles/theme-base.css
@@ -182,7 +182,7 @@ textarea {
 .navbar-search {
   position: relative;
   float: left;
-  margin-top: .625rem;
+  margin-top: .770rem;
   margin-bottom: 0;
   width:100%;
   -moz-box-sizing:border-box;


styles/theme-base.css
```diff
@@ -182,7 +182,7 @@ textarea {
 .navbar-search {
   position: relative;
   float: left;
-  margin-top: .625rem;
+  margin-top: .770rem;
```
Bilge: It would be better to use a vertical centring technique, instead of arbitrarily adjusting one margin, to ensure consistency across all devices and screen sizes.

alystair: @Bilge is there a place one can view browser stats for php.net so we know the most recent method we could use?

php-pulls:

Thank @jesobreira for your contribution. Applied via d26b9a7148a10ff7d75927ecf0908477bab60d0a

PR #230, opened 2018-06-30
Remove unused Git attributes ident

petk:
Hello,

The `$Id$` keywords were used in Subversion where they can be substituted with filename, last revision number change, last changed date, and last user who changed it.

In Git this functionality is different and can be done with Git attribute ident. These need to be defined manually for each file in the `.gitattributes` file and are afterwards replaced with 40-character
hexadecimal blob object name which is based only on the particular file contents.

This patch simplifies handling of `$Id$` keywords by removing them since they are not used anymore.

### DIFF
--- a/ChangeLog-4.php
+++ b/ChangeLog-4.php
@@ -1,5 +1,4 @@
 <?php
-// $Id$
 $_SERVER['BASE_PAGE'] = 'ChangeLog-4.php';
 include_once $_SERVER['DOCUMENT_ROOT'] . '/include/prepend.inc';
 include_once $_SERVER['DOCUMENT_ROOT'] . '/include/changelogs.inc';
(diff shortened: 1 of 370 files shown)


jpauli:
I'm not sure about it, so I did not merge it.
Let's wait for others to comment if we really don't use this keyword anymore in any kind of processing

Perhaps @Tyrael knows ?

Tyrael:
looks good to me, these are just leftovers from the subversion times.

petk:
So this is good to merge then? :)

petk:
Bumping...

salathe:
> I think it's better to merge this someone who is also familiar with the deployment process of php.net in this case.

Two people who fall into this category have looked and added their comments. 

> I don't have sufficient karma to push to this repository. 

[Now you do](http://svn.php.net/viewvc?view=revision&revision=345820) (give it an hour or two to work its way through the pipes).

Thanks for your continued contributions! 👍

petk:
Thank you for enabling me the push karma for web. I appreciate this. Then this has been applied via 0f95b8c7092395df67ee7c86fe25cb130d0db2e2

Thank you everyone for reviewing this.

PR #234, opened 2018-07-30
Fixed the vertical align of "supported versions" chart

javiereguiluz:
The chart in http://php.net/supported-versions.php has a minor issue in Firefox. The version labels are not vertically aligned.

## Before (Chrome, Firefox, Safari)
before

## After (Chrome, Firefox, Safari)
after


LawnGnome:
Merged; thanks!

PR #237, opened 2018-08-21
added conference announcement for Cascadia PHP 2018

lapubell:
please add our conference to the php site. woo hoo!


adamculp:
Merged. Had to fix a small conflict because a news item was merged prior to this one.

PR #244, opened 2018-10-03
Fix search bar under https://secure.php.net site

coladict:
Due to whatever configuration difficulty you haven't solved on the site for several years now, the search bar only works in the non-https version. Browsers block the scripts because of CORS protections. This should solve that problem.


php-pulls:

Pushed as cd346fb10bcedb30e8e99f4803c7b3858e1845b2

PR #250, opened 2018-10-21
Remove automake as a dependency to build PHP

petk:
The Automake and aclocal were part of the previous PHP build system
where Automake created Makefile from the Makefile.in templates and
aclocal was used to produce the aclocal.m4. Need for Automake as a
dependency to build PHP has been removed in PHP 4.3.0 release and
aclocal also in that period of 4.3 releases.

Refs:
- http://php.net/ChangeLog-4.php
- http://git.php.net/?p=php-src.git;h=e715fb00f8850990d51cee4e08e84a58f0eb0012
- http://git.php.net/?p=php-src.git;h=9d9d39a0de3bec962c343051011f5a2ed7d7b242

### DIFF
--- a/git.php
+++ b/git.php
@@ -50,7 +50,6 @@ site_header("Git Access", array("current" => "community"));
    <li><i>PHP 7.3</i>: 2.68+</li>
   </ul>
  </li>
- <li><i>automake</i>: 1.4+</li>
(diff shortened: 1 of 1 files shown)


petk:
Applied via 29b30614a6942d57765d61b71a3b2e6c6a4d4a0c

PR #251, opened 2018-10-21
Bump minimum PHP requirement for php.net sites to 5.4

petk:
Currently all mirrors already use at least 5.4 versions and also the code in the repository already uses 5.4 syntax in some places, such as `releases/feed.php` file with `__DIR__` constant.

This patch syncs this state a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,5 +24,5 @@ To install a full official mirror please see [the mirroring guidelines](http://p
 
 ## Code requirements
 
-Code must function on a vanilla PHP 5.3 installation as some php.net mirrors still run PHP 5.3.
+Code must function on a vanilla PHP 5.4 installation as some php.net mirrors still run PHP 5.4.
 Please keep this in mind before filing a pull request.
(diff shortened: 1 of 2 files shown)


petk:
P.S: [Supported versions timeline](http://php.net/supported-versions.php) ...

salathe:
> Currently all mirrors already use at least 5.4 versions

According to the [mirror administration](https://master.php.net/manage/mirrors.php) page (php.net account login required), a good chunk of the mirrors are still on 5.3. 


>and also the code in the repository already uses 5.4 syntax in some places, such as `releases/feed.php` file with `__DIR__` constant.

That particular constant came was added in 5.3.0.

petk:
@salathe thanks for the info. Uh, I've missed that terribly. Good list for an overview of all that.

petk:
Closing in favor of discussion on [mailing lists](http://news.php.net/php.webmaster/27399)...

PR #252, opened 2018-11-01
Bug #72707: Filter releases by minor or patch version

zrhoffman:
Before: https://secure.php.net/releases/index.php?json&version=7.1&max=10 returns all version 7 releases, minor version ignored.

After: https://secure.php.net/releases/index.php?json&version=7.1&max=10 only returns version 7.1 releases.

Also after: https://secure.php.net/releases/index.php?json&version=7.1.22 returns version 7.1.22 only.

Existing functionality (e.g., filtering by major versions like 4, 5, 7, etc.) should be unaffected.

### DIFF
--- a/include/branches.inc
+++ b/include/branches.inc
@@ -358,8 +358,6 @@ function compare_version($arrayA, $versionB)
 
 function version_array($version, $length = null)
 {
-	static $fill = array(0, 0);
-
 	$versionArray = array_map(
 		'intval',
 		explode('.', $version)
--- a/include/branches.inc
+++ b/include/branches.inc
@@ -359,24 +359,21 @@ function compare_version($arrayA, $versionB)
 function version_array($version, $length = null)
 {
 	static $fill = array(0, 0);
+
 	$versionArray = array_map(
 		'intval',
 		explode('.', $version)
 	);
-	if (is_int($length)) {
-		$versionArray = array_merge(
-			$versionArray,
-			$fill
-		);
-	} else {
-		$length = 3;
-	}
 
-	array_slice(
-		$versionArray,
-		0,
-		$length
-	);
+	if (is_int($length)) {
+		$versionArray = count($versionArray) < $length
+			? array_pad($versionArray, $length, 0)
+			: array_slice(
+				$versionArray,
+				0,
+				$length
+			);
+	}
 
 	return $versionArray;
 }
--- a/include/branches.inc
+++ b/include/branches.inc
@@ -330,3 +330,53 @@ function get_branch_support_state($branch) {
 
 	return null;
 }
+
+function compare_version($arrayA, $versionB)
+{
+	static $sortValues = array(
+		true => 1,
+		false => -1,
+	);
+
+	$length = count($arrayA);
+	$arrayB = version_array($versionB, $length);
+
+	if (!is_array($arrayA) || !is_array($arrayB)) {
+		return $sortValues[$arrayA > $arrayB];
+	}
+
+	foreach ($arrayA as $index => $componentA) {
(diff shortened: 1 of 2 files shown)


petk:
Thank you for adding this one @zrhoffman The new functionality looks good and working as it should...

include/branches.inc
```diff
@@ -330,3 +330,53 @@ function get_branch_support_state($branch) {
...
+
+function version_array($version, $length = null)
+{
+	static $fill = array(0, 0);
+	$versionArray = array_map(
+		'intval',
+		explode('.', $version)
+	);
+	if (is_int($length)) {
+		$versionArray = array_merge(
```
salathe: `array_merge(array(7, 2), array(0, 0))` gives `array(7, 2, 0, 0)`.  Is this really what you intend to happen here?

include/branches.inc
```diff
@@ -330,3 +330,53 @@ function get_branch_support_state($branch) {
...
+			$versionArray,
+			$fill
+		);
+	} else {
+		$length = 3;
+	}
+
+	array_slice(
```
salathe: `array_slice()` doesn't change `$versionArray`, it returns the slice (that you aren't using).

include/branches.inc
zrhoffman: Essentially yes before slicing it to `$length`, but `array_pad()` is more straightforward. Changed in 5fdaddee9c

include/branches.inc
zrhoffman: Oops, you're right, that should be an assignment. Fixed in 5fdaddee9c.

petk:
I've rechecked this one and after the changes the data looks ok compared to previous returns. I think we can move to the merge after @salathe rechecks... Thanks

petk:
If this one is fine, I'll merge it tomorrow or if someone with access does it sooner... There are some `each()` function usages still present in this file (not related to this change but related to #253

releases/index.php
```diff
@@ -10,36 +10,42 @@
...
 				}
-			} else {
-				$r["version"] = $version;
+			}
 
-				$machineReadable = $r;
+			if (!$maxSet) {
+				$version = key($machineReadable);
+				$machineReadable = current($machineReadable);
+				$machineReadable["version"] = $version;
```
salathe: This block will result in `$machineReadable = array("version" => null);` if the version in `$_GET["version"]` is not found (e.g. `7.7777777`).  That's probably not the best response to give.

releases/index.php
```diff
@@ -10,36 +10,42 @@
 	$machineReadable = array();
 
 	if (isset($_GET["version"])) {
-		$ver = (int)$_GET["version"];
+		$versionArray = version_array($_GET["version"]);
+		$ver = $versionArray[0];
 
 		if (isset($RELEASES[$ver])) {
-			list($version, $r) = each($RELEASES[$ver]);
+			$RELEASES = array_replace_recursive($RELEASES, $OLDRELEASES);
```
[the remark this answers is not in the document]
zrhoffman: Good catch! Fixed in ca1f5741ed

releases/index.php
zrhoffman: Agreed. We should print "Unknown version" in those cases. Fixed in dee30de722 and 2aacc05a87

petk:
Sorry for the noise... Just checking, is this one ready to merge?

zrhoffman:
@salathe's points have been addressed. It should be good to go.

php-pulls:

Applied via b7231b60cdd3d1bbf03621216b889d19995756a4

Thank you @zrhoffman.

PR #253, opened 2018-11-04
Refactor each() function to foreach()

petk:
The `each()` function has been deprecated since PHP 7.2 and shouldn't be used anymore:
- http://php.net/manual/en/function.each.php

### DIFF
--- a/bin/bumpRelease
+++ b/bin/bumpRelease
@@ -21,7 +21,8 @@ if (isset($_SERVER['argv'][2])) {
 	$info = $RELEASES[$major][$version];
 } else {
 	// Calling without a minor will just grab the most recent minor.
-	list($version, $info) = each($RELEASES[$major]);
+	$version = key($RELEASES[$major]);
+	$info = current($RELEASES[$major]);
 }
 
 $info["date"] = ${"PHP_{$major}_DATE"};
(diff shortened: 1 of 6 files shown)


bin/bumpRelease
```diff
@@ -21,7 +21,8 @@ if (isset($_SERVER['argv'][2])) {
 	$info = $RELEASES[$major][$version];
 } else {
 	// Calling without a minor will just grab the most recent minor.
-	list($version, $info) = each($RELEASES[$major]);
+	$version = $major;
```
salathe: This changes the contents of `$version` from a whole version number (e.g. `7.2.12`) to just a major version number (i.e. `7`).

bin/bumpRelease
```diff
@@ -21,7 +21,8 @@ if (isset($_SERVER['argv'][2])) {
...
+	$info = $RELEASES[$major];
```
salathe: This changes the contents of `$info` from the information about a single release (the most recent release for the major version, e.g. `$RELEASES[7]["7.2.12"]), to an array of information about the latest minor release for major version (e.g. `$RELEASES[7]`).

petk: Thank you for the review. Fixed in the next commit...

petk:
I'll wait for the #252 to get finalized first and then replace two remaining each() occurrences in the conflicting file here... Thank you.

PR #254, opened 2018-11-12
Correct 7.1.23 release date

MaxGiting:
I think 7.1.23 release date was wrong on [http://php.net/releases/](http://php.net/releases/)


php-pulls:
**Comment on behalf of cmb at php.net:**

Thanks! Applied as dfadd340848dc4f26f2216438c542e7b0b9ad5bf.

PR #257, opened 2018-12-02
Fix off by-one error. Today (Dec 2nd), it shows 'today', where it's really the last day.

derickr:


derickr:
I think it's currently ambiguous because it says "Security Support Until" in the header. That is wrong in this case, as it's "until tomorrow", not today. As that would mean that today there would not have been security support for this branch. "Until" is usually meant exclusively. Instead of my suggestion of "last day", it can also be "tomorrow" instead.

winks:
@cmb69 It can be read as correct if it stands on its own - but this was the original observation: 

![img][https://i.imgur.com/GXhVKlv.png]

cmb69:
Ah, I see. May we should use “midnight” then? However, it seems to me that this is somewhat moot anyway, since it is already tomorrow in some parts of the world. :) For instance, au1.php.net shows: “
3 Dec 2018 | today”.

winks:
Yeah, there's still the timezone issue. But that's a bit too granular I think. Being consistent is good enough.

cmb69:
Okay, LGTM.

cmb69:
Applied as 09f261e36459d5aed5b6a3ab7c8b27ee6e001bb0.

petk:
A side note: there is now a new branch `security-update-off-by-one` left in the web-php repository that probably should be removed :)

PR #266, opened 2019-03-03
Add "Bulgaria PHP Conference 2019" to conferences announcments

mkosturkov:
This is a request to add the 2019 edition of Bulgaria PHP Conference to the list of conferences on php.net.

This is the first time I do such a PR, so I apologize in advance if I've gotten something wrong.


petk:
Thank you @mkosturkov for the conference info. Merged.

PR #267, opened 2019-03-10
Announce Bulgaria PHP Conference 2019 CFP

mkosturkov:
Bulgaria PHP Conference 2019 opened their CFP, so it'd be great to have the conference listed under "Conferences calling for papers"

NOTE: There is an entry for the conference already under "Upcoming conferences". I am not sure how the process works, so I did not remove it. If I should, please, reject this PR and let me know that I should, so I can fix the situation and create a new one.

### DIFF
--- a/archive/archive.xml
+++ b/archive/archive.xml
@@ -9,6 +9,7 @@
     <uri>http://php.net/contact</uri>
     <email>php-webmaster@lists.php.net</email>
   </author>
+  <xi:include href="entries/2019-03-10-1.xml"/>
   <xi:include href="entries/2019-03-07-3.xml"/>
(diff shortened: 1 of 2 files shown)


petk:
I'll check this out soon today... Basically, yes once the conference has been announced duplicate entries should be thought about a bit. Usually conferences have one or two announcements depending on the context - CFP opened and the event announcement. We can adjust this most likely for Bulgaria conference...

petk:
Applied via f68f0093156c61ffb1f90581f129e57564a9fc1c Thanks @mkosturkov 👍

php-pulls:

Applied.

mkosturkov:
Thanks!

PR #268, opened 2019-03-11
Redirect from inactive languages to EN (doc bug #77637)

salathe:
This patch adds in redirects from manual page links for inactive languages to go to their English equivalent.

For example, since our Bulgarian translation is marked as inactive, visitors going to http://php.net/manual/bg/function.strpos.php will get redirected to http://php.net/manual/en/function.strpos.php

### DIFF
--- a/error.php
+++ b/error.php
@@ -635,6 +635,17 @@ if (preg_match("!^manual/(.+)/function\.(.+)-(.+).php$!", $URI, $array)) {
     }
 }
 
+// ============================================================================
+// For manual pages for inactive languages, if the EN page exists, redirect there
+if (preg_match("!^manual/([^/]+)/([^/]+).php$!", $URI, $match) &&
+    isset($INACTIVE_ONLINE_LANGUAGES[$match[1]])) {
+    $try = find_manual_page("en", $match[2]);
+    if ($try) {
+        status_header(301);
+        mirror_redirect($try);
+        exit;
+    }
+}
 
 // ============================================================================
 // 404 page for manual pages (eg. not built language)


error.php
```diff
@@ -635,6 +635,17 @@
     }
 }
 
+// ============================================================================
+// For manual pages for inactive languages, if the EN page exists, redirect there
+if (preg_match("!^manual/([^/]+)/([^/]+).php$!", $URI, $match) &&
+    isset($INACTIVE_ONLINE_LANGUAGES[$match[1]])) {
+    $try = find_manual_page("en", $match[2]);
+    if ($try) {
+        status_header(301);
```
salathe: 301 is meant to be permanent, does this mean we'll have a hard time reviving inactive translations?

error.php
```diff
@@ -635,6 +635,17 @@
     }
 }
 
+// ============================================================================
+// For manual pages for inactive languages, if the EN page exists, redirect there
+if (preg_match("!^manual/([^/]+)/([^/]+).php$!", $URI, $match) &&
+    isset($INACTIVE_ONLINE_LANGUAGES[$match[1]])) {
+    $try = find_manual_page("en", $match[2]);
```
salathe: We could use `$LANG` instead of `en` here, but I figured `en` is always the best place to direct people.

cmb69:
If someone is looking for a particular translation, telling them the translation would have been **moved** to EN, seems wrong to me. Instead, I'd prefer to stick with the 404 response (might be changed to 410), and to add a link to the EN version in the response body.

salathe:
@cmb69 from most of the complaints thus far, it seems to be Google directing folks to the translated pages, not necessarily a specific choice on the part of the user.

This should really be "fixed" by Google no longer sending visitors to those pages now that they just serve a 404 page.

salathe:
Superseded by friendlier error page in 4716c2ba20.

PR #274, opened 2019-03-27
Remove unused legacy reST redirections

petk:
The php.net/reST pages once served some source code files. It has been removed via 74bc9e6ad77fb860ae8e171d46b4914a3737d7a9 in 2013 so these redirections are most likely not relevant anymore for PHP.net audience.


petk:
Applied via b8332da255ab483c9b4b7cb9b9b8bcf41463c597 

One day maybe it would be also good to refactor HTTP status codes. Currently the site returns 302 instead of 404 and these will be popping up:
-> https://www.google.com/search?q=site%3Aphp.net+"reST"

PR #275, opened 2019-03-29
Remove references, redirects, and selection of mirrors

derickr:
This removes the use of mirrors across php.net, and moves everything to https.


petk:
@derickr is it also ok with you to remove the stalled branch `get-rid-of-the-mirrors` before we move too much into the commits future? Thanks.

derickr:
Oh sure. I had forgotten. I've deleted it now.

PR #276, opened 2019-04-02
Add historical maintainer for lv1.php.net (15+ years)

laacz:
There was another maintainer of php.net's Latvia mirror site since early 2000s (until August of 2018). That was me. If not much of trouble, I'd like to be mentioned.


salathe:
Thanks for the PR, and your many years supporting the mirror!

PR #281, opened 2019-05-10
Fix typo in URL.

alanr15:


php-pulls:

Applied via 97da026.

Thanks!

PR #284, opened 2019-06-01
Keep CVEs in changelog

cmb69:
When snipping the author attribution from each changelog entry, we have
to keep eventual CVEs.  We therefore only remove the last parenthesized
part and possibly surrounding whitespace.


cmb69:
Obsoleted by d8e2012418740e9e26f576e25b1f2463739abea5.

PR #288, opened 2019-06-25
Bump required bison version for PHP-7.4

petk:
Since PHP 7.4, parser and lexer files are no longer bundled with the Git repository PHP code. And the Bison 3.0.0 or later is needed to generate parser files.


cmb69:
Good catch! Please merge.

petk:
Merged. Thanks...

PR #289, opened 2019-07-06
Changing styles for navigation bar

SerhiiCho:
### Intro
Navigation bar is a little bit outdated for 2019, especially the mobile version. I think it's a good idea to make it look more **modern** for new people coming into the language. But at the same time don't make anything fancy and try to stick with the base structure. **Navigation bar** is a good place to start, but in future I'd like to continue with other parts as well. Just want to know your guys opinion about restyling [php.net](https://php.net)

### Search bar on mobile view
The biggest change is that search bar is no longer visible on mobile view, instead there is a button with search icon on the left of the **navbar**. After first click it will show the search and **auto focus** on search input, but after second click it will try to **submit** the search form.

### Quick preview
web-php

_**Feel free to edit if you are a maintainer.**_


SerhiiCho:
Quick preview [link](http://ip172-18-0-57-bkg76bdtcgkg00eqlvf0-7770.direct.labs.play-with-docker.com/docs.php) on play with docker, the link will be dead after 3 hours 50 minutes since this comment.
Only Home and Documentation pages work for the demo.

Sobak:
I'm not the right person to judge aesthetics - I'm backend all the way lol, and don't even have any strong opinion on that one. I do think, however, that huge diff caused by changing the identation is scarying people from reviewing the pull request.

Not only it will take more time for someone to review it but you're going to be asked to revert unrelated changes for the sake of diff readability and not screwing tools like git blame.

One functional problem I can see with that change is that it alters the colors which are also used on other PHP.net websites (e.g. doc.php.net, news.php.net, people.php.net etc). It means that if we want to have new look we would also need complementary change to the php/web-shared repo.

SerhiiCho:
The indentation part should have been the separate PR I guess. But it definitely needs to be fixed. I'm closing the pull request

SerhiiCho:
I've created an [extension](https://addons.mozilla.org/en-US/firefox/addon/php-revival/#&gid=1&pid=2) for Firefox for changing the styles on php.net, maybe it will be helpful to some people

SimonCastano:
I think this should not have been closed 
I love php but if we don't make a redesign of the user interface
php will keep losing contributors

and that's why people say that php is dead.
let's improve the documentation

PR #297, opened 2019-09-16
Move PHP 4 warning

nicoSWD:
Hi!

I feel like a _red giant warning about PHP 4 documentation_ is not the first thing people should see in 2019 when looking for documentation.

I think it's a little less confusing and attention grabbing below the table, like this:

Screenshot 2019-09-16 at 18 05 51

What do you think?

(Oh, and there's no such thing as _informations_)

### DIFF
--- a/docs.php
+++ b/docs.php
@@ -65,14 +65,14 @@ foreach ($ACTIVE_ONLINE_LANGUAGES as $langcode => $langname) {
  </tr>
 </table>
 
-    <div>
-        <p>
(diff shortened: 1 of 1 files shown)


nicoSWD:
Thanks, @cmb69!

What does it take to move this forward? I understand that this is very low priority, just wondering what the protocol is here.

cmb69:
Thanks for the reminder. Since nobody objected, I have applied this as 886585ad0b89d92a4e2af33733d2c377f2257172 now.

PR #298, opened 2019-09-17
Update announcement for Bulgaria PHP Conference 2019

mkosturkov:
Bulgaria PHP Conference 2019 is less than two months from now and this is an announcement for its final program, tickets and contacts.

Please, in case of any problems with this PR, let me know, so I can fix whatever is needed!

Thank you in advance for the cooperation!


cmb69:
Thanks! Applied as 7bfc3236f9bb9a98e847a97db8094bedaf1b672f.

mkosturkov:
Thank you!

PR #303, opened 2019-11-10
ScotlandPHP 2020

dannyockilson:
In an attempt not to forget to add the conference to the site until a week before it starts I thought I'd go fully to the other extreme.

We will have more information to add as we get things sorted but wanted to get the basic info (date/venue) in as soon as we knew it.

Let me know if there's any issues, thanks


cmb69:
Well, I think announcing ScotlandPHP a year in advance is *very* early, and I have doubts that this would be helpful for the conference. @salathe, thoughts?

dannyockilson:
I can keep the text on our branch and schedule it in later, think the ordering would be a problem with it getting lost in a sea of newer entries

PR #304, opened 2019-11-28
hash error

bashkarev:


derickr:
Thanks! I've fixed this in a slightly different way, to prevent this from happening next time.

PR #306, opened 2019-12-11
Removed dead information from Readme

markrandall:
Removed official mirrors from readme, there are no official mirrors anymore. 

Removed PHP 5.3 (!) requirement from readme, replaced with 7.3

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,12 +17,7 @@ operation of this website, such as
  - User contributed notes for manual pages
  - A "router" for the builtin PHP webserver
 
-
-## Setting up an official mirror
-
-To install a full official mirror please see [the mirroring guidelines](http://php.net/mirroring).
-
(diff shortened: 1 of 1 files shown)


cmb69 (commented):
Thanks! LGTM, except for the PHP version requirement. It seems the site is currently running on 7.2, so that should probably be used as minimum requirement.

marandall:
> It seems the site is currently running on 7.2, so that should probably be used as minimum requirement.

As 7.2 is in security-only should we stick a pin in it and wait for someone to upgrade to 7.4, or just reference 7.2?

cmb69:
I would just reference 7.2 for now.

cmb69:
Thanks! Applied as 8b832daa28b63b81097d3282e6c3ab046c5802e1.

PR #307, opened 2019-12-12
Adding PHPWales to the party

JustSteveKing:
Adding the upcoming PHPWales to the party - had issues with the command line script not accepting the HTML information (would not exit), so sorry if it isn't quite formatted 100%


archive/entries/2019-12-12-1.xml
```diff
@@ -0,0 +1,33 @@
...
+  <published>2019-10-31T09:04:51+01:00</published>
+  <updated>2019-10-31T09:04:51+01:00</updated>
+  <link href="https://www.php.net/conferences/index.php#id2019-12-12-1" rel="alternate" type="text/html"/>
+  <link href="https://www.phpwales.co.uk" rel="via" type="text/html"/>
+  <default:finalTeaserDate xmlns="http://php.net/ns/news">2020-06-03</default:finalTeaserDate>
+  <category term="conferences" label="Conference announcement"/>
+  <default:newsImage xmlns="http://php.net/ns/news" link="https://www.phpwales.co.uk" title="PHP Wales 2020">php-wales-2020.png</default:newsImage>
+  <content type="xhtml">
+    <div xmlns="http://www.w3.org/1999/xhtml">
+        <h3>
```
cmb69: I don't think a `<h3>` is good here, mostly because the title is already rendered as `<h3>` on https://www.php.net/conferences/ (not sure if it's rendered differently elsewhere). Why not use an introductory paragraph?

JustSteveKing: Thanks for the advise! I have changed this to a `p` tag now

cmb69:
Thanks! Applied as e8b2ce3e09a7a45cf0420fd51accb8319cd00a26.

PR #313, opened 2020-01-17
Added SunshinePHP 2020 conference announcement

adamculp:
Added SunshinePHP 2020 conference announcement


adamculp:
Thanks for the commit.

PR #315, opened 2020-01-23
Correct PHP Russia 2020 date

samdark:
Venue issues forced us to move conference date :(


cmb69:
Thanks! Applied as af7243a8cc2ee9e320dded5e551ae9a4844f4921.

PR #319, opened 2020-03-17
Fix typo: "Marc" is not the name of any month

SjonHortensius:


cmb69:
Good catch, thanks! Applied as 9b6a58916a18e19ba6292ec3db438779949ddf35.

PR #321, opened 2020-04-29
Announce The Online PHP Conference

sebastianbergmann:


cmb69:
Thanks! Applied as 6a4389fd31d540d4a7cca8e066b6c1a85fcc9718.

sebastianbergmann:
Thank you, @cmb69, for merging this so quickly.

However, the HTML rendered (my bad for not testing that locally) contains

```html
<a href=""><img src="/images/news/"></a>
````

We do not have a logo that we can display here. What do we need to do so that the above HTML is not rendered?

cmb69:
It seems to me [the linked image is output unconditionally](https://github.com/php/web-php/blob/64221f5e49b5e2c7fee02d7ce9956f188132ae47/conferences/index.php#L38), what looks like a bug.

PR #322, opened 2020-04-30
Only render HTML for image when there is an image

sebastianbergmann:
Fixes https://github.com/php/web-php/pull/321#issuecomment-621647915


cmb69:
Thanks! Applied as 640baa75da05828193acfd13196ae8dde6e0c837. (will take a while to get pushed to the server)

PR #324, opened 2020-05-04
Remove safe mode from main feature list

wimg:
Safe mode has been deprecated for a long time. There's no reason it should still be in the list of main features

### DIFF
--- a/manual/en/index.php
+++ b/manual/en/index.php
@@ -3,46 +3,45 @@ include_once __DIR__ . '/../../include/shared-manual.inc';
 $TOC = array();
 $PARENTS = array();
 
-$setup = array (
-  'home' =>
-  array (
-    0 => 'index.php',
-    1 => 'PHP Manual',
-  ),
-  'head' =>
-  array (
-    0 => 'UTF-8',
-    1 => 'en',
-  ),
-  'this' =>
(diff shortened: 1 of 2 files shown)


cmb69:
Hmm, is features.inc even used anywhere?

wimg:
I believe it is here : https://www.php.net/manual/en/

cmb69:
The TOC of the manual is build from the docs sources; e.g. https://www.php.net/manual/de/index.php shows German titles, while features.inc has English titles only. I assume features.inc is a relict from olden days.

wimg:
I fixed it in the right place. I also modified the html (and a small part of the PHP) to make it all more human readable, since it was quite awful, which explains why it was never maintained.
Only did the English version of it though...

cmb69:
I had a closer look, and the code your patching here is only for demo/testing purposes (actually anything in https://github.com/php/web-php/tree/master/manual). What you actually want to do is to modify (or even delete) https://github.com/php/doc-en/blob/master/features/safe-mode.xml (note the different repository), but since the manual is still supposed to document all versions starting from PHP 5.0, this might not be a good idea. Anyhow, feel free to provide a PR against https://github.com/php/doc-en. :)

PR #326, opened 2020-05-15
Fix bugfix title

demiurge-ash:


cmb69:
Ah, good catch! Thanks.

Applied as 5e1f78e602827fc1715486d8502f4019ff1a6e9a.

PR #328, opened 2020-05-21
breadcrumbs and next/prev links is now fixed as navbar

Resident234:
hi !

breadcrumbs and next/prev links is now fixed as navbar

this is necessary in order to make it easier to switch to the next page, since now, after reading the page, you don’t have to scroll to the very beginning of the page

image
image

### DIFF
--- a/include/header.inc
+++ b/include/header.inc
@@ -122,6 +122,37 @@ if (!isset($config["languages"])) {
       <input type="search" name="pattern" class="search-query" placeholder="Search" accesskey="s">
     </form>
   </div>
+    <?php if (!empty($config['breadcrumbs'])): ?>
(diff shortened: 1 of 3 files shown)


afilina:
Looks like the bar appears twice when at the top of the page:

image

Also, it takes up a bit too much of real estate on mobile devices in portrait mode. Example on an iPhoneX:

image

saundefined:
I agree with Anna, the bar appears twice

jimwins:
Here's an alternative implementation that moves the breadcrumbs within `<div id="layout">` and then `position: sticky` to keep it in place, on desktop only.

https://github.com/jimwins/web-php/commit/a3e819e9cc4253f9125360b7ef4b484cab4d9079

PR #287, opened 2019-06-19
 Replace http:// to https://

dusta:
Change of important links

PR #329, opened 2020-05-25
Use HTTPS links on Get Involved page

caugner:


cmb69:
Thanks for the PR, but it may overlap with PR #287.

caugner:
Thanks, I wasn't aware of that PR.

PR #332, opened 2020-06-10
Simplify design. Use bold colors.

Anteste:
I want to simplify the desigb by using bold colors

### DIFF
--- a/styles/home.css
+++ b/styles/home.css
@@ -1,12 +1,10 @@
 /* Home Page */
 
 #intro p {
-  color:#fff;
   text-rendering: optimizeLegibility;
 }
 #intro p,
 #intro ul {
-    text-shadow:0 1px 2px rgba(0,0,0,.5);
     font-size:1.125rem;
 }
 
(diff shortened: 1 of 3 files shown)


cmb69:
@morrisonlevi, since this is about an old commit of you, what do you think?

lhsazevedo:
Adding an screenshot here just so people now what it looks like.
It looks simpler and bolder indeed, but I think the [PHP 8 release page](https://www.php.net/releases/8.0/en.php) style is more in line with the PHP brand.

screenshot

samdark:
I think this one should be closed since, as @lhsazevedo mentioned, current style matches PHP brand way better.

@Anteste thanks anyway.

Pythone:
I think using the color of the year would be good

kamil-tekiela:
It has been 8 years and the design has changed a little bit. The PR is now grossly outdated. It did contain some interesting improvements, but they should be made in a fresh commit. This PR as it is now cannot be merged and reviewing it now doesn't make much sense. If you are still interesting in pursuing these changes, please create a new PR.

PR #335, opened 2020-06-29
Update search.js

Kelta-King:
ES6 syntax for modernizing the site...

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -11,12 +11,13 @@
      * @constructor
      * @param {String} label The label to show the user.
      */
-    var Backend = function (label) {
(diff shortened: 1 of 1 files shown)


js/search.js
```diff
@@ -11,12 +11,13 @@
...
      * @param {String} label The label to show the user.
      */
-    var Backend = function (label) {
+    let Backend = function (label) {
         this.label = label;
         this.elements = {};
     };
 
     /**
+
```
morrisonlevi, line 20: I assume this specific space was un-intentional?

Kelta-King: Yes

PR #338, opened 2020-07-10
Fix typo for CVE in 09 July release announcements

JordanMussi:
Link and text different by 1 digit. CVE from text is completely unrelated so changed to match link from curl website. :)


carusogabriel:
cc @sgolemon

cmb69:
Thank you! Applied as c79d12f17e1e03514e98582252f157cb9e15695e.

PR #344, opened 2020-09-25
Add the current PHP logo as the favicon

RiversideRocks:
Just added the current PHP logo as the favicon for consistency.


morrisonlevi:
As far as I can tell (graphics are not my forte), this is a 32x17 image; that's definitely not a standard size for favicons.  Also, it looks like it has the oval and a white outline, which do not scale down to this size well.

Do you want to try making updated 32x32 and 96x96 versions?

RiversideRocks:
I ran the official php svg through ffmpeg and this is the result:


I looks a bit crushed, so I am going to reduce the height.

morrisonlevi:
I think the "alternative" logo will be better starting point for smaller sizes: https://www.php.net/download-logos.php.

RiversideRocks:
https://www.php.net/images/logos/php-icon.png

This looks quite good in the favicon. Only issue is that it is quite small.

image

Would this work?

wh0:
Do you have a page with that new icon set? I want to test at 1x DPI.

RiversideRocks:
Give this a shot:
https://php-demo-favicon.glitch.me/

RiversideRocks:
This page uses the 32x32 logo from php.net.

wh0:
Thanks. Here's how it looks, for anyone else wondering:

image

RiversideRocks:
Does this work @morrisonlevi?

lhsazevedo:
I know it's been some time since this was proposed, but here is a suggestion using the logo created by @morrisonlevi.
The background blue is #4F5B93, from the PHP 8 release page.

image
image

It looks closer to the current icon, and I find it more readable as well.
If you think it would be good to go ahead with this, let me know and I'll open a PR.

lhsazevedo:
I guess this can be closed now

PR #346, opened 2020-10-02
Added missing reference to bug 47021

mbeccati:
Becuase I put it in the wrong section the same day the previous releases were branched.


cmb69:
Thanks! Applied as 4deab21db7486e6b36e33e14213db7a389863bff. Note that we omit the author information in the changelog (not in NEWS, though).

mbeccati:
@cmb69 Thanks, I didn't even notice... Makes sense that my name was there in the PDO entry because I also had forgotten the `).`

PR #349, opened 2020-10-31
Fix parameter scrolling for proper variadic parameters

cmb69:
This would be required if we'll have [proper variadics](https://github.com/php/doc-en/pull/172) with [proper rendering](https://github.com/php/phd/pull/29).


nikic (approved):

cmb69:
Applied as c18e2b3db75363b43875bf1b9151b8b882dbfdb1.

PR #356, opened 2020-11-26
Add preg_last_error_msg() function to PHP 8 announcement page

nicoSWD:


Girgias:
I'm not sure this makes sense, this page is not meant to be the migration guide, and the only functions which it covers are core functions.

nicoSWD:
Ah okay, fair enough. I figured this was simply an oversight.

Girgias:
It might make sense to add it, but there are also other functions (LDAP/OpenSSL, etc) which got added but one could argue PCRE being always available should make this list.

(Disclaimer I haven't worked on this page just did the French translation so take my opinion with a grain of salt)

cmb69:
Anyway, most important is that the function is in the [migration guide](https://www.php.net/manual/en/migration80.new-features.php#migration80.new-features.pcre) and also has [its manual page](https://www.php.net/manual/en/function.preg-last-error-msg.php).

PR #357, opened 2020-11-26
Fixed ridiculous example

xepozz:
Now class `User` processes the **routing** for **posts** action :grinning: 

image

I've fixed ridiculoues example of routing classes :+1: 

image

### DIFF
--- a/releases/8.0/de.php
+++ b/releases/8.0/de.php
@@ -98,11 +98,13 @@ META
         <div class="php8-compare__label">PHP 7</div>
         <div class="php8-code phpcode">
             <?php highlight_php_trimmed(
-                '/**
-* @Route("/api/posts/{id}", methods={"GET", "HEAD"})
-*/
-class User
-{'
+                'class PostsController
+{
+    /**
+     * @Route("/api/posts/{id}", methods={"GET"})
+     */
+    public function get($id) { /* ... */ }
+}'
             );?>
         </div>
       </div>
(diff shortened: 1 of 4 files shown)


releases/8.0/de.php
```diff
@@ -98,11 +98,13 @@
         <div class="php8-compare__label">PHP 7</div>
         <div class="php8-code phpcode">
             <?php highlight_php_trimmed(
-                '/**
-* @Route("/api/posts/{id}", methods={"GET", "HEAD"})
```
azjezz, line 102: a simpler fix would have been changing `api/posts/{id}` to `api/users` 🤔

xepozz: And changing `User` -> `UsersController`?)

xepozz: All the same we need to present method's attribute when we using `Route` annotation.
Controller's action is most popular example instead of separated action's action.

azjezz: actually, it can be applied to both at the same time, and this is not a real application, it's only an example to show the new syntax, and is not referring to a specific library/framework. 

https://github.com/azjezz/symfony-boilerplate/blob/f4661afc5790d3a7b42925fba1de23bfa9bb3d6b/src/Controller/User/DeleteController.php#L29-L56

xepozz: image
@azjezz WDYT about this ^?

xepozz: I made in separate PR #358

cmb69:
Thank you!

PR #364, opened 2020-11-27
Proposed improvements to PHP 8 page

BenMorel:
As [proposed on @internals](https://externals.io/message/112307#112308), a couple suggestions to improve the PHP 8 page:

- what happens when `match()` doesn't match;
- clarification of the nullsafe operator: IMO, "fails" could imply that an unknown property, or a property that evaluates to a non-object is acceptable.

### DIFF
--- a/releases/8.0/en.php
+++ b/releases/8.0/en.php
@@ -248,6 +248,7 @@ echo $result;
         <li>Match is an expression, meaning its result can be stored in a variable or returned.</li>
         <li>Match branches only support single-line expressions and do not need a break; statement.</li>
         <li>Match does strict comparisons.</li>
+        <li>Match throws an <code>UnhandledMatchError</code> if no match is found and no <code>default</code> is provided.</li>
       </ul>
     </div>
   </div>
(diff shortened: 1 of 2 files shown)


saundefined (approved):

saundefined:
@BenMorel, @cmb69,  maybe also highlight the `code` block? 

e.g.
```css
/* styles/php8.css */
.php8-compare__content code {
  color: rgba(39, 40, 44, 0.7);
  background-color: rgba(39, 40, 44, 0.05);
  padding: .2em .4em;
  font-size: 85%;
  border-radius: 6px;
}
```

BenMorel:
@saundefined Done!

cmb69:
I guess this is fine, but I also presume that most translations will not be updated. It might be a good idea to leave this page alone, and to focus on the manual instead. :)

saundefined:
Hey! As I can see, this case is indicated in the documentation.

After the release of the page about PHP 8.1, I think this PR can be closed, since, in fact, the translations will not react to this change.

I hope that by the next releases we will be able to slightly improve the functionality for translations so that it does not block the update of the original :)

Anyway thanks for the PR! 🙌

PR #365, opened 2020-11-27
Translate php8.0 announcement to spanish

ange1a:


php-pulls:

Comitted as 69bca726

PR #367, opened 2020-11-28
added .conferences class to images, updated css to limit size

gunnard:
https://bugs.php.net/bug.php?id=78060


cmb69:
Thanks for the PR, but I don't think this is an appropriate solution, because it may distort the images (consider a 400px * 400px image). And even if we avoid distorted images, why send large images to the browser, when we only want to show a certain size? Okay, would make sense for displays with high density, but if we wanted to cater to that, we probaly should use `srcset` or such. OTOH, I don't see a real issue with the way it's currently handled: just don't commit overly large images.

gunnard:
Your explanation makes sense. Thank you!

Sent from ProtonMail mobile

-------- Original Message --------

PR #368, opened 2020-11-28
Translate the switcher

tvlooy:
It might be a good idea to translate the language switcher so people don't have to know English to find their language in it.

### DIFF
--- a/releases/8.0/common.php
+++ b/releases/8.0/common.php
@@ -58,12 +58,12 @@ META
 function language_chooser(string $currentLang): void {
     $LANGUAGES = [
         'en'    => 'English',
-        'pt_BR' => 'Brazilian Portuguese',
-        'de'    => 'German',
-        'es'    => 'Spanish',
-        'fr'    => 'French',
-        'ru'    => 'Russian',
-        'zh'    => 'Chinese Simplified',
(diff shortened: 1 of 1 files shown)


sgolemon:
+1 on doing this from me.  It didn't quite sit right that these should be anglified.

releases/8.0/common.php
```diff
@@ -58,12 +58,12 @@ function common_header(string $description): void {
...
-        'pt_BR' => 'Brazilian Portuguese',
-        'de'    => 'German',
-        'es'    => 'Spanish',
-        'fr'    => 'French',
-        'ru'    => 'Russian',
-        'zh'    => 'Chinese Simplified',
+        'de'    => 'Deutsch',
+        'es'    => 'Español',
+        'fr'    => 'Français',
+        'pt_BR' => 'Português do Brasil',
```
sgolemon, line 64: Not sure why you changed the order as well, but I'm not going to worry about that for now.  IMO we should probably asort() this whole thing.

php-pulls:

Merged as 290632d2

releases/8.0/common.php
tvlooy: ah yes, asort() would be better :)

PR #370, opened 2020-11-28
Italian translation for the PHP 8 release page

Jean85:


cmb69:
Thank you!

Jean85:
It doesn't seem to work :( did I do any mistake in the code at the top?

cmb69:
That is because the Italian language is marked as inactive (the IT docs shouldn't be shown, because they're out-dated). I'm looking for a way to resolve that.

cmb69:
I'm afraid there is no (non super hacky) way to make that work, without removing IT from the [list of inactive languages](https://github.com/php/web-php/blob/57b20d2798492ab77c6f8f3ab773e4d8338cf258/include/languages.inc#L59-L85). However, doing so may make the Italian docs available. Don't know what to do. :(

Girgias:
The Italian translation looks like it's not that much translated according to http://doc.php.net/revcheck.php?p=filesummary&lang=it, I wonder how "easy" it would be to go trough the translation and fix the build.
The main issue is the files which don't have a revision number sadly as it can basically mean they need to be retranslated fully at this point :/

cmb69:
A broken build is not necessarily an issue, but rather that likely the majority of translated files are outdated, what is bad, since we don't indicate that to the reader. One option would be to `svn delete` all possibly out-dated files, but that would make it harder to grab up the Italian translation later.

cmb69:
@Jean85, should work now: https://www.php.net/releases/8.0/it.php. Thanks again for the translation!

PR #373, opened 2020-11-29
Bengali translation of the PHP 8 Announcement Page

SMAtaurRahman:


releases/8.0/bn.php
```diff
@@ -0,0 +1,502 @@
...
+include_once __DIR__ . '/common.php';
+
+releases\php80\language_redirect('bn');
+
+releases\php80\common_header(
+    'PHP 8.0 is a major update of the PHP language. ' .
+    'It contains many new features and optimizations including ' .
+    'named arguments, union types, attributes, constructor property promotion, ' .
+    'match expression, nullsafe operator, JIT, and ' .
+    'improvements in the type system, error handling, and consistency.');
```
cmb69, line 12: This paragraph should be translated as well.

releases/8.0/bn.php
```diff
@@ -0,0 +1,502 @@
...
+      <div class="php8-compare__arrow"></div>
+      <div class="php8-compare__block example-contents">
+        <div class="php8-compare__label php8-compare__label_new">PHP 8</div>
+        <div class="php8-code phpcode">
+            <?php highlight_php_trimmed(
+                'echo match (8.0) {
+  \'8.0\' => "Oh no!",
+  8.0 => "This is what I expected",
+};
+//> This is what I expected'
```
cmb69, line 241: Consider to translate the strings ("Oh no!" and "This is what I expected") in this example.

include/languages.inc
```diff
@@ -13,6 +13,7 @@ $LANGUAGES = array(
     'en'    => 'English',
     'ar'    => 'Arabic',
     'bg'    => 'Bulgarian',
+    'bn'    => 'Bengali',
```
cmb69, line 16: Please revert this change. It is no longer needed; the addition in releases/8.0/common.php is sufficient.

cmb69 (commented):
Thank you for the translation! I can't comment on the translated parts, because I don't speak Bengali, but left a few notes.

releases/8.0/bn.php
SMAtaurRahman: Thank you for pointing it out. I have updated the translation.

releases/8.0/bn.php
SMAtaurRahman: Usually, we don't write bengali in code comments. Because some editor may not have unicode support. IMO, I think we should left it as it is. Some of the other translation left it too.

include/languages.inc
SMAtaurRahman: Thanks. Should I rebase on master?

releases/8.0/bn.php
cmb69: Fair enough. :)

include/languages.inc
cmb69: Unless there would be merge conflict, rebase is not necessary.

cmb69:
Could someone please review?

PR #375, opened 2020-11-30
Use own language selection for PHP 8 release page

cmb69:
The general language selection capability of php.net is nice, but it
does not allow to have translations of the PHP 8 release page which do
not have active doc translations.  To circumvent this limitation, we
use an own super simple language selection capability for the PHP 8
release page.  We do not even bother to support non JS enabled clients,
since that was not supported before, although it could be done.


kallesommernielsen (approved):

sgolemon (approved):
Nice and practical solution.

cmb69:
Thanks!

PR #379, opened 2020-12-05
Fix typo

kamil-tekiela:
Credit goes to https://chat.stackoverflow.com/transcript/message/51088407

### DIFF
--- a/releases/8.0/de.php
+++ b/releases/8.0/de.php
@@ -124,7 +124,7 @@ releases\php80\common_header(
   public function __construct(
     float $x = 0.0,
     float $y = 0.0,
-    float $z = 0.0,
+    float $z = 0.0
   ) {
     $this->x = $x;
     $this->y = $y;
--- a/releases/8.0/en.php
+++ b/releases/8.0/en.php
@@ -126,7 +126,7 @@ releases\php80\common_header(
   public function __construct(
     float $x = 0.0,
     float $y = 0.0,
-    float $z = 0.0,
+    float $z = 0.0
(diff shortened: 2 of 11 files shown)


cmb69:
Thanks! It would be a good idea to fix this for all available translations.

kamil-tekiela:
> Thanks! It would be a good idea to fix this for all available translations.

Done

cmb69:
Thank you!

carusogabriel:
Wait, wait, wait: why is this a typo? The linked message makes no sense, this is PHP >= 8.0, testing this in a PHP 7 version will break for sure, but not because of the comms, but the missing promotion support. And this is promoting the RFC that allows a trailing comma in a list of parameters

cmb69:
Prior to PHP 8.0.0, trailing commas in parameter definition lists were not allowed, but the PHP 7 example used one.

kamil-tekiela:
> Wait, wait, wait: why is this a typo? The linked message makes no sense, this is PHP >= 8.0, testing this in a PHP 7 version will break for sure, but not because of the comms, but the missing promotion support. And this is promoting the RFC that allows a trailing comma in a list of parameters

On the left-hand side, there are examples of the old way of doing things in PHP 7.4. The trailing comma is only allowed since PHP 8.0 so that example would not work on PHP 7.4. Someone must have made this silly mistake when writing the code and didn't spot the trailing comma which makes this a typo. The code would compile fine under PHP 8.0, but the idea is to show how it worked in PHP 7.4.

PR #380, opened 2020-12-05
Fix the chant "Go update!

kamil-tekiela:
It really bugs me that there is no comma there. I suggest to replace it with the word upgrade.


cmb69:
Interestingly, most of the translations already translate your suggestion. However, French is something like "Upgrade to PHP 8.0 right away!", and German "Upgrade to PHP 8.0 now!" Maybe that is a nice alternative to the "go"?

saundefined (approved):

kelunik (approved):

cmb69 (approved):
Thank you!

PR #385, opened 2020-12-09
Legacy redirect

afilina:
I was discussing this on phpdoc@lists.php.net and the team was open to the idea of my maintaining the retired manuals separately. Legacy manuals are created from published archives and now live as an [opens-source project][repo]. Zend has agreed to help with hosting and maintenance.

- Check whether this manual page exists in the legacy manual on [php-legacy-docs.zend.com][legacy]
- Display a custom 404 that suggests trying the legacy manuals, with disclaimer about accountability.
- Use jQuery to append the hash part of the link, which isn't visible to the backend.
- Cache results of the legacy lookup for faster 404s next time.
- If there is no legacy manual counterpart, then display the usual 404 page.

Screenshot:
image

[repo]: https://github.com/zendtech/php-legacy-docs
[legacy]: https://php-legacy-docs.zend.com


salathe:
While it is a great that you, and Zend, are hosting versions of the legacy docs on your website; I don't think it is anything that PHP.net should be involved with in any way.   What's next, linking to the [Internet Archive](https://archive.org) since they also have a copy of the page?

weierophinney:
@salathe Why do you feel it's not a good idea? There are still plenty of people using PHP 5 (for better or for worse), and having the historical documentation readily available can be useful to understand why code may have been written in a particular way. The ask here is only to redirect those requests to a different server since the docs team has already said they don't want to serve them off of php.net...

afilina:
Plus, the archive is open source, so if there are any inaccuracies, they will be fixed. It's not static line a zip archive. This will ensure that anyone who wants to upgrade their code to the latest PHP version can actually understand the old behavior. I myself refer to this doc almist daily.

The docs team seemed excited about the idea on the mailing list and suggested that I propose a patch here. They also confirmed that the main reason why the legacy docs got removed in September was because of the lack of bandwidth from the team to maintain all these versions.

harikt:
@salathe this is a custom 404 page which shows the links and not redirecting without giving you any insight. So this looks good to me. 

@afilina regarding the links, will it redirect to appropriate page or just land on home page ?

Thanks all for your good work.

salathe:
I've been contacted offline by some folks, re. my previous comment, which it appears came across as a rejection of the whole notion of having legacy versions of the manual being made available.  

- I'm _not_ against making the legacy docs available (or, more available than they already are). 
- I'm _not_ against Anna / Zend (or anyone else who wants to!) spending their time and effort maintaining these legacy docs. 
- I _am_ against pointing visitors to zend.com for the legacy docs from the php.net error page.  This is my _opinion_ only, which really counts for nought, and is not a rejection on any technical grounds.

Now, on to what I think _should_ be done:

My opinion is that hosting and making available the legacy docs should be something that the PHP.net project undertakes: indeed it _does_ right now, just in a not-so-accessible format (the archive tarballs on doc.php.net).  I would like to see this changed, such that these easily-browsable, searchable, etc. archives are hosted under php.net (e.g. alongside, or in replacement of the tarballs at [doc.php.net/archives](http://doc.php.net/archives/)); in a Git repo,  available for updates/contributions to those (i.e. Anna / Zend) who want to maintain the legacy docs.  We have the hosting, tooling, etc. to make this happen right now.

cmb69:
I'm not so optimistic about our capabilities to host the legacy docs and the repos on php.net. I mean, we're not really good at maintaing our infrastructure, so any additional burden may cause more issues elsewhere. Also, wouldn't we need to review PRs and cater to documentation bugs? We're not doing a particularly great job with this alreay.

I don't have an issue with this PR (except maybe that the checks for the legacy documentation, albeit cached, may cause bad performance). For now, we may just add general links to https://www.php.net/manual/php4.php and https://www.php.net/manual/php5.php.

afilina:
@cmb69 Just to better understand your suggestion regarding performance. Do you mean that only the current php4 and php5 manuals should point to the legacy manuals, not the 404 pages?

I would ideally also like to see the 404s offer a path forward, in case someone lands there from StackOverflow or a forum, like I did. I could always pre-generate the cache file and add it to the PR, instead of creating it on the fly.

cmb69:
@afilina, frankly, I have no idea whether performance would be an issue in practice; I just noticed that when trying out the PR, but of course, the cache wasn't filled then. Pregenerating the cache file would definitely solve that, but would be more manual work.

afilina:
@cmb69 I can automate that. Since there's very little chance of creating new pages in the legacy docs, this would most likely be a one-off. If generating the cache file ever becomes an issue, I can add a script to the `php-legacy-docs` project and a CI task to enforce it. On the `web-php` side, if it's not in the cache, the site will assume the legacy page doesn't exist. I'll just wait for a confirmation that there is an intent to merge before I invest further time.

afilina:
@salathe Hi, can you please approve this pull request or reply to @cmb69's counter-points? It seems like your objection is the only thing blocking this PR right now. Thanks.

salathe:
> @salathe Hi, can you please approve this pull request or reply to @cmb69's counter-points?

Re. additional burden on our server(s): these are just static HTML pages that any server can handle, if load is an issue (I doubt it would be) then front them with Myra.  While manpower (specifically lack of it) for maintaining infra. is an ongoing concern more generally throughout the project, I don't think hosting these pages adds anything to that.

Re. having to review PRs and cater to bugs, why would "we"? (While anyone wanting to, obviously could...) Anna / Zend have volunteered to take on that burden already, whether it's in their own place or under the PHP.net umbrella.

Re. the code and determining whether a legacy page exists; I like the direction the discussion has been moving (a static, ideally automatically curated, list of the legacy pages).

> It seems like your objection is the only thing blocking this PR right now. Thanks.

I don't like the proposal to link visitors to _\<third-party site outside of the PHP.net umbrella\>_, that's a personal opinion not a technical objection; however, I don't see that as enough to consider the PR blocked.  

I _would_ like to hear from folks in the "PHP.net web team" specifically, and others more generally, with their opinions though: the lack of discussion (even just thumbs up) is disconcerting since we can't tell whether there is general approval, disapproval, or indifference.  I guess committing the changes and seeing what happens is one way forward.

heiglandreas:
> While manpower (specifically lack of it) for maintaining infra. is an ongoing concern more generally throughout the project, I don't think hosting these pages adds anything to that.

Meeting proposals from new people willing to help out maintaining infra with silence might not really help in that regards. Just saying ;-) (And perhaps hijacking this for other Docs-Topics :grin:)

afilina:
@cmb69 I addressed the performance issue. 
- Added open-source [script][script] to legacy docs to index the pages. The JSON is now lighter to ease up on memory usage.
- Removed HTTP requests and caching, and instead using the pre-generated index, which will be committed to this repo.

[script]: https://github.com/zendtech/php-legacy-docs/tree/main/bin

cmb69 (commented):
Thanks again! Since there has been no more feedback, I'm going to commit.

PR #386, opened 2020-12-10
Allow to access symlinked docs

cmb69:
Since `realpath()` resolves links, symlinked docs as suggested by the
"Setting up Documentation environment" page[1] are incompatible with
the router script.  Of course, the Webserver can be run without the
router script, but in that case shortcuts (e.g. `/json_decode`) won't
work.

Thus, we remove the `realpath()` resolution.

[1] <http://doc.php.net/tutorial/local-setup.php>

### DIFF
--- a/.router.php
+++ b/.router.php
@@ -2,14 +2,11 @@
 $_SERVER["SERVER_ADDR"] = $_SERVER["HTTP_HOST"];
 
 $filename = isset($_SERVER["PATH_INFO"]) ? $_SERVER["PATH_INFO"] : $_SERVER["SCRIPT_NAME"];
-$afilename = $_SERVER["DOCUMENT_ROOT"] . $filename;
-$len = strlen($_SERVER["DOCUMENT_ROOT"]);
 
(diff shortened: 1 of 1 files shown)


.router.php
```diff
@@ -3,7 +3,6 @@
 
 $filename = isset($_SERVER["PATH_INFO"]) ? $_SERVER["PATH_INFO"] : $_SERVER["SCRIPT_NAME"];
 $afilename = $_SERVER["DOCUMENT_ROOT"] . $filename;
-$afilename = realpath($afilename);
 $len = strlen($_SERVER["DOCUMENT_ROOT"]);
 
 if (strncmp($_SERVER["DOCUMENT_ROOT"], $afilename, $len) == 0) {
```
salathe, line 8: The `strncmp()` check is redundant now, since `$afilename` will *always* start with `$_SERVER["DOCUMENT_ROOT"]`.

cmb69: Oh, indeed! Thanks.

salathe (approved):

PR #389, opened 2020-12-29
Replace "Edit" with "Submit a Pull Request" link

cmb69:
Since the documentation has moved from SVN to Git, we replace the
"Edit" links in the manual with links to the respective GH mirrors.

### DIFF
--- a/include/shared-manual.inc
+++ b/include/shared-manual.inc
@@ -447,7 +447,7 @@ function manual_setup($setup) {
       {$language_chooser($config['lang'], $config['thispage'])}
     </div>
     <div class="edit-bug">
-      <a href="https://edit.php.net/?project=PHP&amp;perm={$config['lang']}/{$config['thispage']}">Edit</a>
+      <a href="https://github.com/php/doc-{$config['lang']}/pulls">Submit a Pull Request</a>
       <a href="https://bugs.php.net/report.php?bug_type=Documentation+problem&amp;manpage=$id">Report a Bug</a>
     </div>
   </div>
--- a/include/shared-manual.inc
+++ b/include/shared-manual.inc
@@ -441,13 +441,14 @@ function manual_setup($setup) {
 
     $id = substr($setup['this'][0], 0, -4);
     $language_chooser = 'manual_language_chooser';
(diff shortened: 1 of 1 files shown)


Girgias:
Might want to wait until the GH mirrors are propagated? As currently the translations are all empty repos.

include/shared-manual.inc
```diff
@@ -447,7 +447,7 @@ function manual_setup($setup) {
       {$language_chooser($config['lang'], $config['thispage'])}
     </div>
     <div class="edit-bug">
-      <a href="https://edit.php.net/?project=PHP&amp;perm={$config['lang']}/{$config['thispage']}">Edit</a>
+      <a href="https://github.com/php/doc-{$config['lang']}/pulls">Submit a Pull Request</a>
```
nikic, line 450: At least for `doc-en`, can we link directly to an edit page like https://github.com/php/doc-en/edit/master/reference/array/functions/array-change-key-case.xml ? I guess for translations this may not be possible, as we don't know if it's in the translation repo or not.

cmb69: Oh, I wasn't aware of these edit pages – nice! Anyhow, the only relevant info regarding the page is `$config['thispage']` which contains the xml:id. There is, however, no readily available mapping to the file which contains the respective element. [PhdOE builds this mapping](https://github.com/php/web-doc-editor/blob/master/php/RepositoryManager.php#L1808) and stores it in the DB. We would have to do something similar (maybe added to revcheck), and to deploy the resulting database.

nikic: I see. In that case I'd suggest to link to the root of the repo (without the `/pulls`) here.

nikic (approved):

include/shared-manual.inc
cmb69: Fine, done. I also catered to pt_BR -> pt-br etc, but did not actually test that.

nikic: I renamed the `doc-pt-br` repo to `doc-pt_br` just now to fix the mirroring, so the str_replace is no longer needed.

PR #396, opened 2021-04-01
Replace references to git.php.net

saundefined:


cmb69:
Ah, good catch! I wonder whether we should redirect to the raw views, e.g. https://github.com/php/php-src/raw/PHP-7.3/NEWS, instead.

Also, would you mind to add redirects for `php74news` and `php80news`? Obviously, these have been forgotten so far.

saundefined:
@cmb69, thanks! Fixed.

cmb69 (approved):
Thank you!

Do you have karma to commit yourself?

saundefined:
@cmb69 thank you for the review. 
Committed it 🎉

PR #406, opened 2021-05-19
Enhancement: Expose happy path

localheinz:
This pull request

* [x] exposes the happy path

### DIFF
--- a/.router.php
+++ b/.router.php
@@ -3,10 +3,12 @@ $_SERVER["SERVER_ADDR"] = $_SERVER["HTTP_HOST"];
 
 $filename = $_SERVER["PATH_INFO"] ?? $_SERVER["SCRIPT_NAME"];
 
-if (file_exists($_SERVER["DOCUMENT_ROOT"] . $filename)) {
-    /* This could be an image or whatever, so don't try to compress it */
-    ini_set("zlib.output_compression", 0);
-    return false;
+if (!file_exists($_SERVER["DOCUMENT_ROOT"] . $filename)) {
+    include_once "error.php";
+
+    return;
 }
 
-include_once "error.php";
+/* This could be an image or whatever, so don't try to compress it */
+ini_set("zlib.output_compression", 0);
+return false;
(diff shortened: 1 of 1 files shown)


.router.php
```diff
@@ -3,10 +3,10 @@
 
 $filename = isset($_SERVER["PATH_INFO"]) ? $_SERVER["PATH_INFO"] : $_SERVER["SCRIPT_NAME"];
 
-if (file_exists($_SERVER["DOCUMENT_ROOT"] . $filename)) {
-    /* This could be an image or whatever, so don't try to compress it */
-    ini_set("zlib.output_compression", 0);
-    return false;
+if (!file_exists($_SERVER["DOCUMENT_ROOT"] . $filename)) {
+    include_once "error.php";
```
cmb69, line 7: Thanks, the patch looks good to me, but maybe add a comment that this never returns?

.router.php
```diff
@@ -3,10 +3,12 @@
...
+
+    return;
```
localheinz, line 9: @cmb69 

How about this?

cmb69 (approved):
Looks good to me. Maybe others disagree? This is a hard call, since the good parts needs to return `false`.

.router.php
sgolemon: I would add: Never assume the CWD or the include path (or error handling):

Use:   `require_once(__DIR__ . '/error.php');`

((Yes, I know the existing version did this wrong, but we can do better!))

localheinz: @sgolemon 

Adjusted!

sgolemon:
Can someone explain to me why the return values in this file are `false` and `null`?  Feels like a wonky API.

cmb69:
> If a PHP file is given on the command line when the web server is started it is treated as a "router" script. The script is run at the start of each HTTP request. If this script returns false, then the requested resource is returned as-is. Otherwise the script's output is returned to the browser.

In the other case ("otherwise"), the return value doesn't matter, but we need to return.

cmb69:
@salathe reacted with thumbs-down emoji. Could you please explain your reasoning?

.router.php
```diff
@@ -3,10 +3,12 @@
...
+if (!file_exists($_SERVER["DOCUMENT_ROOT"] . $filename)) {
+    require_once __DIR__ . '/error.php';
+
+    return;
 }
 
-include_once "error.php";
+/* This could be an image or whatever, so don't try to compress it */
+ini_set("zlib.output_compression", 0);
+return false;
```
Girgias, line 14: I would prefer:
```suggestion
return null;
```
Here and above.

localheinz:
Thank you, @cmb69, @Girgias, and @sgolemon - and please take a look at #878, if you can!

PR #407, opened 2021-05-25
Point "Report a Bug" link to github

nikic:
This creates an empty issue with the following body:

 ```
From manual page: https://php.net/$id

---
```

### DIFF
--- a/include/shared-manual.inc
+++ b/include/shared-manual.inc
@@ -449,7 +449,7 @@ function manual_setup($setup) {
     </div>
     <div class="edit-bug">
       <a href="https://github.com/php/doc-{$repo}">Submit a Pull Request</a>
-      <a href="https://bugs.php.net/report.php?bug_type=Documentation+problem&amp;manpage=$id">Report a Bug</a>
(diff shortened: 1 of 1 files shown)


saundefined:
@nikic, `doc-ru` and `doc-tr` have enabled issues too.
Maybe use the lang-depended "report a bug" link and English as default?

<details>
  <summary>e.g.</summary> 

```php
$issuesEnabled = ['ru', 'tr'];
$bugsUrl = 'https://github.com/php/doc-' . (in_array($repo, $issuesEnabled, true) ? $repo :  'en') . '/issues/new?body=' . urlencode("From manual page: https://php.net/$id\n---\n");
  ```

</details>

Girgias:
In this case we should enable issues for all doc repos, the French one doesn't have them enabled as of yet.

nikic:
@saundefined I think the key question is, when somebody clicks "Report a Bug" on a translated page, how often do they want to report an error in the content (that they just happen to see in a translated version) vs how often do they want to report a translation mistake.

I think it may make more sense to point everything at doc-en and then move issues to translations if they really are about translations only.

saundefined (approved):

PR #408, opened 2021-05-25
Enhancement: Document platform requirement in `composer.json`

localheinz:
This pull request

* [x] documents the platform requirement in `composer.json`

Follows https://github.com/php/web-php/pull/606#discussion_r912514942.

💁‍♂️ This project currently does not use `composer` to manage its dependencies, but it does not need to stay that way. Apart from using `composer` to manage dependencies, `composer.json` also serves documentation purposes. For a start, I propose to add `composer.json` with platform requirements - we can take it from there, if that make sense for you.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,19 @@
+{
+    "name": "php/web-php",
+    "description": "The www.php.net site.",
+    "license": "proprietary",
+    "type": "project",
+    "homepage": "https://www.php.net/",
+    "support": {
+        "source": "https://github.com/php/web-php"
+    },
+    "require": {
+        "php": "^7.3"
+    },
+    "config": {
+        "platform": {
+            "php": "7.3.19"
+        },
+        "sort-packages": true
+    }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+/vendor/
+
 backend/mirror.gif
 backend/mirror.png
 backend/mirror.jpg
--- a/composer.lock
+++ b/composer.lock
@@ -1,0 +1,23 @@
+{
+    "_readme": [
+        "This file locks the dependencies of your project to a known state",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
+        "This file is @generated automatically"
+    ],
+    "content-hash": "c0f8a9b8c5c6efaa1eed8cf44ed8695e",
+    "packages": [],
+    "packages-dev": [],
+    "aliases": [],
+    "minimum-stability": "stable",
+    "stability-flags": [],
+    "prefer-stable": false,
+    "prefer-lowest": false,
+    "platform": {
+        "php": "^7.3"
+    },
+    "platform-dev": [],
+    "platform-overrides": {
+        "php": "7.3.19"
+    },
+    "plugin-api-version": "2.3.0"
+}


composer.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "php/web-php",
+    "type": "project",
+    "description": "The www.php.net site.",
+    "homepage": "https://www.php.net/",
```
localheinz, line 5: Taken from the description of this repository.

composer.json
```diff
@@ -0,0 +1,16 @@
...
+    "license": "proprietary",
+    "require": {
+        "php": "^8.0"
```
localheinz, line 8: I am just guessing here - can you help me out?

carusogabriel: 7.2, check the readme: https://github.com/php/web-php#code-requirements

localheinz: @carusogabriel 

Ha, didn't notice.

But why PHP 7.2? Why not PHP 5.6, or, let's say, a version of PHP that is actually supported?

Girgias: Because it hasn't been tested with a more recent version?

Also no need to be aggressive in your comment, the code base is maintained by volunteers who don't have unlimited time.

cmb69: Not sure which version the server is running. :/

localheinz: @cmb69 

Can we find out? Would be nice to move forward!

Things that come to my mind:

- once the version requirement is properly document (in `composer.json` instead of `README.md`), we could start using GitHub Actions
- we could then add jobs to 
   -  enforce coding standards
   - run a static code analysis
   - run a dependency analysis
   - run tests
 - sooner or later we could start using a supported version of PHP

What do you think?

localheinz: @Girgias 

Apologies for my comment, but I had hoped that the website would run on a more recent PHP version!

cmb69: IMHO no need to apologize. :)

Generally I'm in favor of your suggestions, but it may make sense to discuss this on the Webmaster mailing list, since there has been some push-back regarding using third-party components, to avoid endorsement of any.

localheinz: @cmb69 

That makes a lot of sense!

I‘ll try to join the mailing list!

localheinz: 🙀

9182792B-1107-4AC9-A10D-72D1F4CDDFAF

cmb69: I'm afraid that form might be broken for years; there are detailed instructions at the bottom of https://www.php.net/mailing-lists.php about another option to subscribe.

kamil-tekiela:
Why is composer.lock committed?

localheinz:
@kamil-tekiela 

See https://naderman.de/slippy/slides/2018-01-26-composer-lock-demystified.pdf:
CleanShot 2022-06-17 at 15 52 59@2x

composer.json
```diff
@@ -0,0 +1,19 @@
...
+    "homepage": "https://www.php.net/",
+    "support": {
+        "source": "https://github.com/php/web-php"
+    },
+    "require": {
+        "php": "^7.2"
+    },
+    "config": {
+        "platform": {
+            "php": "7.2"
```
localheinz, line 15: See https://github.com/php/web-php/blob/32f5a0283c3b24818cabf184f1dcfa085889c79b/README.md#code-requirements.

composer.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "php/web-php",
+    "description": "The www.php.net site.",
+    "license": "proprietary",
```
cmb69, line 4: Ugh, I'm afraid you're right. Not nice.

composer.json
```diff
@@ -0,0 +1,19 @@
...
+    "type": "project",
+    "homepage": "https://www.php.net/",
+    "support": {
+        "source": "https://github.com/php/web-php"
+    },
+    "require": {
+        "php": "^7.2"
```
cmb69, line 11: I assume the code is compatible with PHP 8.0 and maybe 8.1.

composer.json
cmb69: TIL! However, the [docs state](https://getcomposer.org/doc/06-config.md#platform):

> Therefore if you use this it is recommended, and safer, to also run the [check-platform-reqs](https://getcomposer.org/doc/03-cli.md#check-platform-reqs) command as part of your deployment strategy.

cmb69 (commented):
I'm generally in favor of adding a composer.json, and if we declare config.platform.php, committing composer.lock appears to be a good idea as well.

composer.json
```diff
@@ -0,0 +1,19 @@
...
+    "name": "php/web-php",
+    "description": "The www.php.net site.",
+    "license": "proprietary",
+    "type": "project",
+    "homepage": "https://www.php.net/",
+    "support": {
+        "source": "https://github.com/php/web-php"
+    },
+    "require": {
+        "php": "^7.3"
```
cmb69, line 11: Shouldn't we claim compatibility with PHP 8, e.g.
```suggestion
        "php": "^7.3 || ^8"
```

localheinz: @cmb69 

I would agree if this were a library - but since this is an application, it will only run in a specific production environment, will it not? That specific production environment currently uses PHP 7.3.19, as you reported.

cmb69: I was afraid it would not be possible to use PHP 8 locally, but apparently that is not really an issue due to `platform: php: 7.3.19`. So we can leave this as is.

cmb69 (approved):
Thank you! This is fine for me.

cmb69:
Thank you!

localheinz:
Thank you, @carusogabriel, @cmb69, @Girgias, and @kamil-tekiela!

PR #409, opened 2021-05-26
Remove reference to Freenode and replace with Libera.Chat

jezemery:
I think we should stop suggesting freenode as the primary irc network

### DIFF
--- a/support.php
+++ b/support.php
@@ -87,7 +87,7 @@ site_header(
  Otherwise known as IRC. Here you can usually find experienced PHP people
  sitting around doing nothing on various channels with php in their names.
  Note that there is no official IRC channel. Check
- <a href="http://www.freenode.net/">freenode</a> or any other major network
+ <a href="https://libera.chat/">Libera.Chat</a> or any other major network
  (<a href="http://www.efnet.org/">EFNet</a>,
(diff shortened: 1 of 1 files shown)


adrian-enspired:
fair enough, but note we don't _quite_ yet control the libera `#php` channels. process is taking a while.

once we do, freenode/##php will be set to redirect to freenode/##moved_to_libera

derickr:
On 26 May 2021 15:04:06 BST, Adrian ***@***.***> wrote:
>fair enough, but note we don't _quite_ yet control the libera `#php`
>channels. process is taking a while.
>
>once we do, freenode/##php will be set to redirect to
>freenode/##moved_to_libera

Let me know if you would like me, as somebody with a php.net account to do the paperwork.

jezemery:
@derickr I saw this on the IRC channel.

> <ss23> 
> Ah, Derick / Xatenev : The paperwork has been filed for a few days now, I presume they're just going through it slowly

adrian-enspired:
@derickr thanks, will let you know once things move

libera admins have a huge backlog of requests right now

adrian-enspired:
@derickr someone from libera chat may or may not contact you in the next few days.

adrian-enspired:
registration is done, this can go forward

PR #413, opened 2021-07-01
Update checksums for 8.0.8 to match published archives

tianon:
See https://github.com/php/web-php/commit/64de340efe5d43092a182d9366db3af2ff1ed917#r52925273 :sweat_smile:

I calculated these by `wget -qO- 'https://www.php.net/distributions/php-8.0.8.tar.xz' | sha256sum` (and `.bz2` and `.gz`).

Alternatively, perhaps the published archives should be updated to the versions that matched these checksums?
(Not sure which fix is correct, but figured I'd get a PR up in case updating the checksums is the right answer.)

### DIFF
--- a/include/version.inc
+++ b/include/version.inc
@@ -24,9 +24,9 @@ $RELEASES = (function() {
         'date'    => '01 Jul 2021',
         'tags'    => [], // Set to ['security'] for security releases.
         'sha256' => [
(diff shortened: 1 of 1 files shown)


tianon:
Fixed via https://github.com/php/web-php/commit/6c63c0a4168739a0572d4ed80d062f56626b40e4 :tada: :heart:

carusogabriel:
Thank you so much, I've copied the wrong ones :(

I should have looked at https://gist.github.com/carusogabriel/b6833d26731a1f7c46daa35a492fd97a

PR #416, opened 2021-07-20
Changed <a name="..."> to <a id="...">  

mikeschinkel:
When opening this repo in PhpStorm it flagged @name as a deprecated HTML attribute, so it seemed appropriate to change then to @id. In cases where there were both a @name and @id attribute and their values were the same, this commit removes the @name attribute _(there were no cases where both @name and @id attributes existed but having different values.)_


cmb69:
Thank you! It might be more important, though, to also fix that in the scripts which are used to generate these entries, e.g. https://github.com/php/web-php/blob/f16e2cb6a3b0ea30a2c6a585180c20f12f941e82/bin/news2html#L98

mikeschinkel:
@cmb69 — lol!  I didn't know that, but know aI do! :-)

I will fix there and updated.

BTW, I also have a few other PRs are will add, each of which is very narrow so it will be easier to review.

mikeschinkel:
@cmb69 — Where are these files?  I can't find them in my cloned repo?

cmb69:
One is bin/news2html. I think there is at least one other to generate the archive entries, but I'm not sure where that is (might be included/used from bin/createNewsEntry).

mikeschinkel:
@cmb69 — Sorry, I just had a brain fart.  
  
  I was looking for the `/f16e2cb6a3b0ea30a2c6a585180c20f12f941e82/` directory.  **Obviously** it wasn't there, as that was a commit hash.  Doh!

mikeschinkel:
@cmb69 — PR updated to include `news2html.php`.

One additional question. 

PhpStorm flagged [line 72](https://github.com/php/web-php/blob/f16e2cb6a3b0ea30a2c6a585180c20f12f941e82/bin/news2html#L73) as an error because of the `\1`:

```
array('<?php bugfix(\1); ?>', '<?php peclbugfix(\1); ?>', 
      '<?php implemented(\1); ?>', '<?php githubissuel(\'php/php-src\',\1); ?>'),
```

Is that valid in some earlier version of PHP that you might still be using?  Shouldn't it be `\$1` instead of just `\1`?:

```
array('<?php bugfix(\$1); ?>', '<?php peclbugfix(\$1); ?>', 
      '<?php implemented(\$1); ?>', '<?php githubissuel(\'php/php-src\',\$1); ?>'),
```

If yes I can either add that fix to this PR or create a new PR; whichever your preference.

cmb69:
> PhpStorm flagged line 72 as an error because of the \1:

Oh, right, this should more correctly be `\\1` or even better `$1`. I'd prefer a separate PR.

mikeschinkel:
@cmb69 — I believe this PR has addressed the issues you mention and should be ready to merge, 

I created two other PRs to deal with the regex issue (#423) and a fix to allow `news2html` to process the only example NEWS files I could find in the repo (#417.)

archive/2007.php
```diff
@@ -14,7 +14,7 @@
 <hr>
 
 <div class="newsItem">
-    <a name="2007-11-08-1" id="2007-11-08-1"><h1>PHP 5.2.5 Released</h1></a>
+    <a  id="2007-11-08-1"><h1>PHP 5.2.5 Released</h1></a>
```
salathe, line 17: Nit: double space character in `<a  id=` throughout this file.

mikeschinkel: @salathe — Must be a search/replace error. I will fix and update.

mikeschinkel: @salathe — Fixed.

mailing-lists.php
```diff
@@ -306,14 +306,14 @@
 // Print out a table for a given list array
 function output_lists_table($mailing_lists)
 {
-    echo '<table cellpadding="5" border="0" class="standard mailing-lists">', "\n";
+    echo '<table class="standard mailing-lists">', "\n";
```
mikeschinkel, line 309: These should not be in this PR.  I am removing them.

include/layout.inc
```diff
@@ -321,8 +321,8 @@ function display_event($event, $include_date = 1)
         $eday = (isset($event['end']) && !empty($event['end'])) ? strtotime($event['end']) : 0;
     }
 ?>
-<table border="0" cellspacing="0" cellpadding="3" width="100%" class="vevent">
- <tr bgcolor="#dddddd"><td>
+<table class="vevent">
+ <tr><td>
```
cmb69, line 325: This change would have a visible effect (see e.g. https://www.php.net/cal.php?id=272&cm=07&cy=2021). I think we would need to adjust the CSV if we're going to remove the deprecated attributes. Maybe just leave them for now?

include/layout.inc
```diff
@@ -367,7 +367,7 @@ function display_event($event, $include_date = 1)
     echo ' (<span class="location">' , $COUNTRIES[$event['country']] , '</span>)';
 ?>
  </td></tr>
- <tr bgcolor="#eeeeee" class="description"><td>
+ <tr class="description"><td>
```
cmb69, line 370: Same as above.

cmb69 (commented):
Thank you for the PR! Except for removing the table attributes this looks very good to me.

include/layout.inc
mikeschinkel: @cmb69 — Did you mean `CSS` instead of `CSV`?

mikeschinkel: @cmb69 — Also, this snuck into this PR but I did not intend for it to be here. It should be in a different PR. I will remove here and add to that one if it is not already there.

cmb69: Yes, CSS, of course. :)

include/layout.inc
mikeschinkel: @cmb69 — I have removed the accidental changes in `layout.inc` from this PR.

cmb69: Thank you!

PR #417, opened 2021-07-21
Fix to allow news2html to handle multi-line 'module'

mikeschinkel:
I am not sure this PR is valid because I only had one set of NEWS files to process.  But, maybe it is:

I needed a NEWS file to test bin/news2html and the only ones in the repo were /resources/NEWS_5_4_0_*.txt. Unfortunately, those had a few 'entries' that appeared to have no 'modules' and that caused the script to crash. So I added logic to recognize those and assign them to the 'Core' module and that allowed the script to work. As part of this, I unravelled the if..elseif..else logic to be a series of if..continue since the if statements were all in a loop and no code was run after the if..elseif..else statement.

### DIFF
--- a/bin/news2html
+++ b/bin/news2html
@@ -42,13 +42,23 @@ while(($ln = fgets($fp)) !== false) {
 	if ($ln[0] == '-') {
 		// module
 		$module = trim(substr($ln, 1), " \t\n:");
-	} elseif (preg_match('/^\s+\.\s/',$ln)) {
-		$entries[$module][] = trim(preg_replace('/^\s+\.\s+/', '', $ln));
-	} else {
-		// continued line
-		$c = count($entries[$module])-1;
-		$entries[$module][$c] = trim($entries[$module][$c] )." ".trim($ln);
+		continue;
 	}
+
+	if (preg_match('/^\s+\.\s/',$ln)) {
+		$entries[$module][] = trim(preg_replace('/^\s+\.\s+/', '', $ln));
+		continue;
+    }
+
+	if ( !isset($entries[$module])) {
+	    // Looks like we have a multi-line entry where a module should be
+	    $entries['Core'][] = $module;
+        $module = 'Core';
+    }
+
+	// continued line
+	$c = count($entries[$module])-1;
+	$entries[$module][$c] = trim($entries[$module][$c] )." ".trim($ln);
 }
 
 if ($changelog) { ob_start(); }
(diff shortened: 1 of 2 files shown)


license/index.php
```diff
@@ -27,7 +27,7 @@
 
 <ul>
  <li>
-  PHP 4, PHP 5 and PHP 7 are distributed under the
+  Starting with PHP 4, versions of the PHP software are distributed under the
```
mikeschinkel, line 30: This is already in master but evidently I didn't rebase before submitting the PR.

cmb69:
You can use the NEWS files in php-src for testing, e.g. https://github.com/php/php-src/blob/master/NEWS and https://github.com/php/php-src/blob/PHP-7.3/NEWS. Older NEWS files are likely irrelevant now,

mikeschinkel:
@cmb69 — Thanks.

salathe:
> Are NEWS files hand-coded, or generated from some other source?

The NEWS files are hand-written.

mikeschinkel:
@salathe — Thanks for answering.  
That is a high bar for getting the form exactly right every time... _(just thinking out loud.)_

bin/news2html
```diff
@@ -70,7 +80,7 @@ foreach($entries as $module => $items) {
 		// convert bug numbers
 		$item = preg_replace(
 			array('/Fixed bug #([0-9]+)/', '/Fixed PECL bug #([0-9]+)/', '/Implemented FR #([0-9]+)/', '/GitHub PR #([0-9]+)/'),
-			array('<?php bugfix(\1); ?>', '<?php peclbugfix(\1); ?>', '<?php implemented(\1); ?>', '<?php githubissuel(\'php/php-src\',\1); ?>'),
+			array('<?php bugfix("$1"); ?>', '<?php peclbugfix("$1"); ?>', '<?php implemented("$1"); ?>', '<?php githubissuel(\'php/php-src\',"$1"); ?>'),
```
sgolemon, line 83: Why this change?

1/ Adding quotes violates the type of the agument `"123" == 123` but `"123" !== 123`.
2/ \1 and $1 are effectively the same here, this seems to be change for the sake of change.

bin/news2html
```diff
@@ -42,13 +42,23 @@ while(($ln = fgets($fp)) !== false) {
...
-		$entries[$module][$c] = trim($entries[$module][$c] )." ".trim($ln);
+		continue;
 	}
+
+	if (preg_match('/^\s+\.\s/',$ln)) {
+		$entries[$module][] = trim(preg_replace('/^\s+\.\s+/', '', $ln));
+		continue;
+    }
+
+	if ( !isset($entries[$module])) {
```
sgolemon, line 53: This feels like a very specific fix for 4.x NEWS files.
Would you mind if this were opt-in behavior?
e.g. `bin/news2html --handle-multiline-entry-looks-like-module ...`

Modern NEWS files don't have this issue (you can find them in the php-src repo, every branch has its own NEWS file at the root of the repo), and nominally this script is only run on new releases.

bin/news2html
```diff
@@ -42,13 +42,23 @@ while(($ln = fgets($fp)) !== false) {
...
+	    // Looks like we have a multi-line entry where a module should be
+	    $entries['Core'][] = $module;
```
sgolemon, line 55: What's with this `Core` module biasing?  My guess is that the bug is only being seen in fixes for the core module, but it seems like it COULD occur in any module.

sgolemon (changes requested):

PR #419, opened 2021-07-21
Removes use of get_magic_quotes_gpc()

mikeschinkel:
As `get_magic_quotes_gpc()` has been removed in PHP 8.0, this PR removes it from the `clean()` function in `layout.inc.`

Hopefully its removal is valid, or at least how it was removed it valid.

### DIFF
--- a/include/layout.inc
+++ b/include/layout.inc
@@ -248,7 +248,7 @@ function sect_to_file($string) {
 }
 
 function clean($var) {
-  return htmlspecialchars( $var, ENT_QUOTES);
+  return htmlspecialchars($var, ENT_QUOTES);
 }
 
 // Clean out the content of one user note for printing to HTML
--- a/include/layout.inc
+++ b/include/layout.inc
@@ -248,7 +248,7 @@ function sect_to_file($string) {
 }
 
 function clean($var) {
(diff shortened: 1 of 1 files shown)


mikeschinkel:
@tiffany-taylor — This PR should fix the mailing list submit form problem, assuming it is acceptable security wise? 

_(I haven't had to deal with much magic quotes/strip slashes code in PHP so I am not 100% sure if what I did is ok.)_

Also, here is the email I sent to the `php.webmaster` list referencing this PR:

https://news-web.php.net/php.webmaster/29448

tiffany-taylor:
I will try to review this over the weekend. My computer is unplugged at present, but I'll be hooking it back up if not tomorrow, then by Saturday.

mikeschinkel:
Thanks @tiffany-taylor.  

Actually thanks to @ramsey's email I looked again and don't think this was it, although this is needed.

Instead I think it was using `HTTP` for the URL instead of `HTTPS`.  [Here's the PR](https://github.com/php/web-php/pull/426) for that.

cmb69:
This change is still needed for PHP 8 compatibility, and causes no issues if the server is running PHP 5.4+ (what I guess it does). The space in front of $var should be removed though.

mikeschinkel:
@cmb69 — Space removed.

cmb69:
Thank you!

PR #421, opened 2021-07-21
Replaced deprecated table attributes with css

mikeschinkel:
HTML deprecated table element attributes @cellpadding, @cellspacing, @border, @align, etc. so this PR replaces them with CSS classes, selectors and style rules.  It also removed a CSS style rule that was being overwritten by the next line, it removes an invalid font rule (See https://stackoverflow.com/a/20641079/102699), it removes redundant `px` from `0px` and it sneaks in a <label> element to wrap an <input> element per PhpStorm's code inspection instructions.

### DIFF
--- a/license/index.php
+++ b/license/index.php
@@ -27,7 +27,7 @@ site_header("License Information", array("current" => "help"));
 
 <ul>
  <li>
-  PHP 4, PHP 5 and PHP 7 are distributed under the
+  Starting with PHP 4, versions of the PHP software are distributed under the
   <a href="http://www.php.net/license/3_01.txt">PHP License v3.01</a>, copyright (c) the <a href="/credits.php">PHP Group</a>.
   <ul>
    <li>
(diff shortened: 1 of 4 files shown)


mailing-lists.php
```diff
@@ -345,8 +345,9 @@ function output_lists_table($mailing_lists)
 <?php output_lists_table($internals_mailing_lists); ?>
 
 <p class="center">
- <strong>Email:</strong>
- <input type="text" name="email" size="40" value="user@example.com">
+ <label for="email"><strong>Email:</strong>
```
kelunik, line 348: There's no need for the `for` attribute if the label wraps the input.

kelunik (approved):

license/index.php
```diff
@@ -27,7 +27,7 @@
 
 <ul>
  <li>
-  PHP 4, PHP 5 and PHP 7 are distributed under the
+  Starting with PHP 4, versions of the PHP software are distributed under the
```
mikeschinkel, line 30: This is already in master but I evidently didn't rebase to master before this PR.

mailing-lists.php
mikeschinkel: @kelunik -— `for` removed.

derickr:
This conflicts. Can you rebase on master?

PR #422, opened 2021-07-21
Add returns to end of functions

mikeschinkel:
PhpStorm flags functions that don't have returns at the end so this PR rearranges code so that the returns are always at the end, or in one case adds one that was missing.

### DIFF
--- a/include/layout.inc
+++ b/include/layout.inc
@@ -42,8 +42,10 @@ function highlight_php($code, $return = FALSE)
         $highlighted
     ) . '</div>';
 
-    if ($return) { return $highlighted; }
-    else { echo $highlighted; }
+    if (! $return) {
+        echo $highlighted;
+    }
+    return $highlighted;
 }
 
 // Same as highlight_php() but does not require '<?php' in $code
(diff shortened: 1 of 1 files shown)


include/layout.inc
```diff
@@ -42,8 +42,10 @@ function highlight_php($code, $return = FALSE)
         $highlighted
     ) . '</div>';
 
-    if ($return) { return $highlighted; }
-    else { echo $highlighted; }
+    if (! $return) {
+        echo $highlighted;
+    }
+    return $highlighted;
```
kelunik, line 48: This is a behavioral change, previously it returned implicitly null, so should it explicitly return null now?

mikeschinkel: @kelunik — Good point.  Let me revise.

mikeschinkel (commented):
Updated to return null per request.

include/layout.inc
```diff
@@ -44,6 +44,7 @@ function highlight_php($code, $return = FALSE)
 
     if ($return) { return $highlighted; }
     else { echo $highlighted; }
```
kelunik, line 46: The `else` could even be skipped here like below, so the if is only used for early return.

mikeschinkel: @kelunik — I was trying to keep to minimal changes.  I was hoping it would be okay to include changes that would make PhpStorm's code inspections not flag acceptable errors, but after this round of PRs is considered I'd like to discuss what the team's goals are for the site and if I can possible help modernize it.

mikeschinkel: @kelunik — Anyway. done.

kelunik:
Thanks!

PR #424, opened 2021-07-21
Fix mismatched <p> tags

mikeschinkel:
When inspecting the code with PhpStorm is reported that the usage of the `<p>` tags that were modified in this commit were invalid. Mostly the `<p>` tags were enclosing `<ul>..</ul>` tags which is not valid HTML. This commit fixes the HTML to be valid and includes

### DIFF
--- a/archive/2012.php
+++ b/archive/2012.php
@@ -459,7 +459,6 @@ site_header("News Archive - 2012", array("cache" => true));
      <p>
      This year promises to be the best ZendCon ever! We are planning a fun new exhibit hall, some great parties including a hackathon and, of course, oodles of excellent PHP content!
      </p>
-     <p>
      <ul>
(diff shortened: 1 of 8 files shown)


kelunik (approved):

archive/2012.php
```diff
@@ -468,21 +467,13 @@
...
-     </li>
-     <li>Tickets: Early reg discount until Sept 8
-     </li>
+     <li>What: PHP 2012 (ZendCon)</li>
+     <li>When: Mon Oct 22- Thurs Oct 25</li>
+     <li>Where: Santa Clara, CA</li>
+     <li>Tickets: Early reg discount until Sept 8</li>
      </ul>
-     </p>
-    </div>
```
kelunik, line 484: Isn't this `</div>` supposed to stay?

PR #425, opened 2021-07-21
Fix mismatched <p> tags

mikeschinkel:
When inspecting the code with PhpStorm is reported that the usage of the <p> tags that were modified in this commit were invalid. Mostly the <p> tags were enclosing <ul>..</ul> tags which is not valid HTML. This commit fixes the HTML to be valid and includes


mikeschinkel:
Accidental duplicate.

PR #426, opened 2021-07-22
Change mailing list submit URL to use HTTPS

mikeschinkel:
The mailing list URL "http://main.php.net/entry/subscribe.php" was not using "https://main.php.net/entry/subscribe.php" and thus was failing.

### DIFF
--- a/mailing-lists.php
+++ b/mailing-lists.php
@@ -72,7 +72,7 @@ if (isset($_POST['action'])) {
 
         // Get in contact with main server to [un]subscribe the user
         $result = posttohost(
-            "http://main.php.net/entry/subscribe.php",
+            "https://main.php.net/entry/subscribe.php",
             array(
                 "request"  => $request,
                 "email"    => $_POST['email'],


cmb69:
Thank you!

ramsey:
This is great! Thanks, @mikeschinkel !

faizanakram99:
@mikeschinkel 

Thanks it works, I was able to sign up now. Earlier it threw some error.

I would reply in mailing list where you asked for feedback (read in externals.io) but don't know how to do it.

mikeschinkel:
@faizanakram99 — Excellent. Thank you for the follow up.

> _"I would reply in mailing list where you asked for feedback (read in externals.io) but don't know how to do it."_

 You could just sign up for the _"PHP php.net internal infrastructure discussion"_ list on the mailing list [page you tested](https://www.php.net/mailing-lists.php), and you will get an email asking for confirmation. After you confirm then you would just email the list at php-webmaster@lists.php.net.

**_However_**, no need to email the list about your success now since by commenting here you've already let us know.

PR #439, opened 2021-08-31
usability: avoid gap between a wrapped link in aside bar

peterdd:
The upcoming conference links are wrapped and clicking between should activate the link, not require the user to decide to click the upper or bottom line of the link.

wrappedlinkgap

### DIFF
--- a/styles/theme-medium.css
+++ b/styles/theme-medium.css
@@ -407,11 +407,13 @@ aside.tips div.border {
 aside.tips h3 {
   color:#E6E6E6;
 }
+aside.tips li {
+  line-height: 1.2rem; /** avoid gaps in wrapped links **/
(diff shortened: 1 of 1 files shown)


peterdd:
Ah, another idea which keeps the hover underlining 'normal' and still no click-gap on wrapped link:

```css
aside.tips.li {margin-bottom:0.5rem;}
aside.tips a {padding-top:0.5rem;}
```

It also keeps the li items a bit separated (for better or 'worse'?)
php_linkgaps

saundefined:
> Ah, another idea which keeps the hover underlining 'normal' and still no click-gap on wrapped link:

Looks good,
can you update PR, and we'll try it on preview environment?

peterdd:
Instead of the padding-top 'hack' I now adjusted the line-height of the li-tag...

saundefined (approved):
Thanks!

Issue #440, opened 2021-09-03
Proposal to add PageGen time to each page


Willtech:
For various reasons including search indexing and server performance include PageGen time in the footer of each generated page.

Willtech:
UPdate footer creation as follows:

```html
<footer>
<div class="container footer-content">
<div class="row-fluid">
<ul class="footmenu">
<li><a href="/copyright.php">Copyright &copy; 2001-2021 The PHP Group</a></li>
<li><a href="/my.php">My PHP.net</a></li>
<li><a href="/contact.php">Contact</a></li>
<li><a href="/sites.php">Other PHP.net sites</a></li>
<li><a href="/privacy.php">Privacy policy</a></li>
<li><a href="https://github.com/php/web-php/blob/master/manual%2Fen%2Ffunction.random-inc.php">View Source</a></li>
<li>(PageGen 0.736984 Seconds)</li>
</ul>
</div>
</div>
</footer>
```

Willtech:
In every file (unless it can be batched with included)
At the beginning of the page generation
```php
//PageGen
$tt = microtime(true);
```

And, at the conclusion of the page generation, as practical as possible,
```php
<li>(PageGen <?php $tt = number_format(rtrim(sprintf("%.20f", (microtime(true) - $tt)), "0"),6,'.',',');
echo $tt; ?> Seconds)</li>
```

Willtech:
Ideally, this would have six decimal places if less than one second, and two decimals if more than one second.

Willtech:
Or, usiung the new method place this code as near as practical to the conclusion of page generation.

```php
<li>(PageGen <?php $tt = number_format(rtrim(sprintf("%.20f", (microtime(true) - $_SERVER['REQUEST_TIME_FLOAT'])), "0"),6,'.',',');
                     if (strpos(($tt."0"), "0") != 0) {
                         $tt = number_format($tt,2,'.',',');
                     }
                     echo $tt; ?> Seconds)</li>
```

cmb69:
Why? Are there any other popular sites which still do this?

salathe:
No, thanks.

Willtech:
> Why? Are there any other popular sites which still do this?

Sure, but if there are more front page results. This site had #1 position sometimes and front page always.
[...]

PR #441, opened 2021-09-03
PageGen: Include time generating page in each page footer

Willtech:
#440 Proposal to add PageGen time to each page.

Working solution tested on my hosting.


salathe:
No, thanks.

PR #444, opened 2021-09-13
fix #62125 no wrap and scroll for source code examples

peterdd:


peterdd:
https://bugs.php.net/bug.php?id=62125
https://www.php.net/manual/en/install.windows.manual.php

saundefined (approved):
Looks good, thanks!

PR #446, opened 2021-09-24
fix #67323 H2 on frontpage do not display in Opera

peterdd:
just do not use ::after css hack for the bottom border and keep it simple.


kelunik:
There's likely a difference with headlines that wrap in multiple lines, no?

peterdd:
What you probably fear is text-decoration: underline looks bad?

I think there is no problem with using border-bottom properties.

PR #447, opened 2021-09-25
qa.php.net, svn.php.net, people.php.net reachable with https

peterdd:
so it should be the default for that links


kelunik:
Thanks!

Issue #453, opened 2021-11-01
Intro adjustments proposal


lhsazevedo:
Hey there!

After some local experiments, I would like to propose some tweaks to the intro element of the home page, also known as hero. There are still some adjustments to be done for mobile devices, which I will do if this proposal is well accepted by the community, followed by a PR.

The goal was to use size, color, white space and call to actions to guide new visitors who might be a little confused by visual clutter. The change also empowers the PHP branding by adding a large logo above the lead, which now has a larger font and an emphasis on the "popular general-purpose scripting language" snippet.

![php2](https://user-images.githubusercontent.com/7695608/139642018-89bc193e-fd5c-4f22-b879-39d3e866b9a6.png)

Animated comparison:
![Comparisson](https://user-images.githubusercontent.com/7695608/139641955-537dfcf2-52ce-46d6-9a2e-cec4e29f446b.gif)

adielcristo:
Thanks for the proposal! I suggest changing the "Changes" link to "Changelog", as it is a more common term (my opinion, I don't have figures/sources for that).

PR #459, opened 2021-11-18
Adjust homepage hero

lhsazevedo:
Some days ago, I [proposed](https://github.com/php/web-php/issues/453) some adjustments to the design if the homepage hero section:

> The goal was to use size, color, white space and call to actions to guide new visitors who might be a little confused by visual clutter. The change also empowers the PHP branding by adding a large logo above the lead, which now has a larger font and an emphasis on the "popular general-purpose scripting language" snippet.

Encouraged by the reactions on the proposal, I went ahead and finished the mobile-first responsive styles and tweaked the CTA buttons style to follow the [PHP 8 Release page](https://www.php.net/releases/8.0/en.php) button design.

The "Get Started" button links to [/manual/en/getting-started.php](https://www.php.net/manual/en/getting-started.php)
The "Download" button link is [/downloads.php](https://www.php.net/downloads.php)

Desktop animated preview:
Desktop animated preview

Mobile animated preview:
Mobile animated preview


samdark (approved):

lhsazevedo:
Setup a temporary mirror. Preview this proposal here: https://php.lhsazevedo.dev/ 

The forced push was just a rebase with master.

styles/home.css
```diff
@@ -1,75 +1,142 @@
...
+  max-width: 240px;
+  margin-bottom: 24px;
 }
 
-#intro .row .blurb p:first-child {
-  margin-top:1.5rem;
+.hero-text {
+  margin-top: 0;
+  margin-bottom: 28px;
+  color:#E6E6E6;
```
kamil-tekiela, line 28: I feel like it looks better without this color

kamil-tekiela:
I like this and I am in favour of it. I would even put the aside permanently at the bottom. I would also change the "upgrading" text to "migration guide"

styles/home.css
lhsazevedo: Done, thanks for reviewing

lhsazevedo:
> I would also change the "upgrading" text to "migration guide"

Initially I intended to change this text as well, but later decided to left it out to reduce the number of changes in this patch. Should we create an issue for this?

Thanks for reviewing.

ramsey (approved):

cmb69:
Linking to the latest release page is also okayish for me, although that is not quite what I'd expect on "Get Started".

lhsazevedo:
Oh sure, the text would also be updated to reflect the link.

The button is now "What's new in 8.1" and links to the release page. I updated the [live preview](https://php.lhsazevedo.dev) (may need a force refresh), and there is a screenshot in my previous comment as well.

Looks great to me!

PR #462, opened 2021-11-19
improve local font support

hakre:
previously local font support directs absolute paths while it is unknown to
me if paths of any kind are accepted for resolution of fonts with the CSS
`local()` function.

therefore adding the known parts that `local()` function resolves to font-
names on the local system (with preference over previous).

for me this is a great improvement as the fonts don't need to be loaded
via HTTP any longer but just work when they are locally installed using
a desktop computer.

as the Fira Sans and Fira Mono typefaces are in use on the PHP website and
my local system normally ships with these, the website loads much faster,
especially visually (fonts are used to display the text of the website).

this also reduces requests to the web-server hosting the PHP homepage
incl. the manual when the system already has the fonts.

NOTE: the circumstance that local() with absolute path-names has no effect
      is - if - not addressed in this change in so far that the original
      directives remain unchanged. this is done b/c I don't understand them
      well and might construe a flaw.

### DIFF
--- a/fonts/Fira/fira.css
+++ b/fonts/Fira/fira.css
@@ -1,7 +1,8 @@
 @font-face{
     font-family: 'Fira Sans';
     src: url('/fonts/Fira/eot/FiraSans-Book.eot');
-    src: local('/fonts/Fira/Fira Sans Book'),
+    src: local('Fira Sans Book'),
+         local('/fonts/Fira/Fira Sans Book'),
          url('/fonts/Fira/eot/FiraSans-Book.eot') format('embedded-opentype'),
          url('/fonts/Fira/woff/FiraSans-Book.woff') format('woff'),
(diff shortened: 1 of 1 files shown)


hakre:
Me thinks this can benefit from a review on the point that if the previous `local()` directives are just superfluous, they can be removed. I for myself would need to dig through the specifications plus additional resources first. All me can say is that with this addition I can see great improvements.

Tested with the PHP built-in webserver (which has some issues and gets stuck with many requests to `cached.php` which also affects font loading _unless_ local fonts work, which is how we tested this).

samdark (approved):

saundefined:
Thank you!

hakre:
@samdark /cc @saundefined - reviewed myself; fix + more portability in #478

PR #123, opened 2016-09-06
Announcing PHPBenelux Conference 2017 Call for Papers is open!

DragonBe:
PHPBenelux Conference 2017 is an annual PHP oriented conference in Antwerp, Belgium. We offer two days of stellar tutorials and talks, epic social events and a lineup of the best local and international businesses involved with PHP.

Come and join the fun at https://cfp.phpbenelux.eu/ or visit our conference website https://conference.phpbenelux.eu/2017/ for more information about this event.

Happy PHP-ing,

The PHPBenelux Crew

PR #469, opened 2021-11-26
Add redirect for php.net/GH-123

nikic:
Redirect `php.net/GH-123` to the corresponding issue or pull request
on php/php-src. This is part of https://wiki.php.net/rfc/github_issues,
but is also useful independently for pull request references.


salathe:
It's a shame this is only for php-src, unlike the old bug tracker version of this.  It'll still be handy to have, regardless.

samdark (approved):

lhsazevedo (approved):

PR #471, opened 2021-11-28
German translation of PHP 8.1 release page

DaDeather:
A german translation of the release page for PHP 8.1


saundefined (approved):
Thank you!

PR #472, opened 2021-11-28
Fix wrong version definition in release page

DaDeather:
The release of PHP 8.1 is a minor release and not a major release as being pointed out on the current release page.

### DIFF
--- a/releases/8.1/languages/en.php
+++ b/releases/8.1/languages/en.php
@@ -1,9 +1,9 @@
 <?php
 
 return [
-    'common_header' => 'PHP 8.1 is a major update of the PHP language. Enums, readonly properties, first-class callable syntax, fibers, intersection types, performance improvements and more.',
+    'common_header' => 'PHP 8.1 is a minor update of the PHP language. Enums, readonly properties, first-class callable syntax, fibers, intersection types, performance improvements and more.',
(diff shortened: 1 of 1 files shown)


cmb69:
Well, it's still a major update regarding the features, I think.

DaDeather:
To be honest I find it a little bit irritating due to the usage of the words in regards to a release.

Since its being called a minor release on the [news page](https://www.php.net/archive/2021.php#2021-11-25-1).

cmb69:
`8.1.0` => `8` is the major version, `1` is the minor version, `0` is the revision number.

DaDeather:
I know the semantic versioning definition thanks :-). What I meant was it seems to be a little bit inconsistent in regards of the words being used.

Saying "latest minor release" on the news page and "major update of the PHP language" on the release page.

Maybe it makes more sense to call it both the same than by changing the release pages text to "latest minor release" as well?

If I'm the only one who's finding it odd than theres of course no sense in this PR anyway.

cmb69:
@patrickallaert, what do you think?

saundefined:
FWIW, the "major" in this case means "big update" 🤔

DaDeather:
Fine with me if its clear to everyone :-). Sorry for the trouble about definitions!

PR #481, opened 2021-12-19
Update spread operator pt_br translation

lhsazevedo:
Update spread operator translation after https://github.com/php/doc-pt_br/issues/49


saundefined (approved):
Thanks!

PR #482, opened 2021-12-21
Add DPC22 Call For Papers

tonivdv:
This is the CFP announcement for the Dutch PHP Conference 2022 :)


cmb69:
Thank you!

cmb69:
By the way, it may make sense to provide a smaller logo image; as is, [the announcement](https://www.php.net/conferences/index.php#id2021-12-21-1) doesn't look good on smaller screens.

PR #485, opened 2021-12-28
Updated Doc link about first class callable syntax.

mumumu:


saundefined (approved):
Looks good, thank you!

PR #487, opened 2022-01-08
Use en dash for the copyright years

infrastation:


cmb69:
Thanks for the PR! However, whether to use a hyphen (`-`) or an en dash (`–`) [for ranges is apparently controversial](https://en.wikipedia.org/wiki/Dash#Ranges_of_values). To avoid the many many changes we would have to do across the php repositories for little gain, I suggest to stick with the hyphen.

infrastation:
My experience is that in source code 7-bit ASCII punctuation is best left as it is, but on the web proper punctuation (`&hellip;` as another example) tends to look better. Of course there's more than one way to look at it and I came to make the point, not the decision.

infrastation:
Alright, never mind.

PR #490, opened 2022-01-15
Fix ext/oci redirections

kocsismate:
Currently, the redirection only works if the URL contains the `.php` suffix, so this PR fixes this behavior.


cmb69 (approved):
Ah, right. Thank you!

kocsismate:
Can you please merge this one as well?

PR #495, opened 2022-02-18
tes

monsad:


cmb69:
That would require to register for a SonarQube account, right?

Issue #500, opened 2022-02-24
Layout: pages without sidebars are not centered


SteveDesmond-ca:
Steps to reproduce:
- browse to https://www.php.net/eol.php with a viewport width between 768px and 979px

Observed result:
- content is squished to the side
![Screen Shot 2022-02-24 at 13 22 45](https://user-images.githubusercontent.com/195896/155584509-65be95cf-b4ba-4dd7-ab62-847bf30f4144.png)

Expected result:
- content is spread out across the viewport, similar to at other resolutions
![Screen Shot 2022-02-24 at 13 23 01](https://user-images.githubusercontent.com/195896/155584543-d55d9407-05d8-43ba-b379-df9def61f997.png)

kelunik:
I'm not sure whether we should prefer a consistent content width or use the remaining space. Having it empty looks weird, though.

PR #501, opened 2022-02-24
Fix content layout on pages without sidebars for 768-979px viewport widths

SteveDesmond-ca:
Per #500, pages without sidebar content (in an `aside` tag) are still pushed to the side at 70% width. I believe this is due to an over-eager selector (the one line changed in this PR).

The combination of `#layout-content` and `:only-child` is used in multiple places elsewhere in the site's CSS files, so I figured that was the safest option to minimize the amount of code being changed.

### DIFF
--- a/styles/theme-base.css
+++ b/styles/theme-base.css
@@ -1574,7 +1574,7 @@ div.soft-deprecation-notice blockquote.sidebar {
         width: 30% !important;
     }
 
-    #layout-content {
+    #layout-content:not(:only-child) {
         width: 70% !important;
     }
 }


samdark:
👍

cmb69 (commented):
Ugh, this nice fix has apparently been overlooked. @SteveDesmond-ca, could you please fix the merge conflict?

SteveDesmond-ca:
@cmb69 No worries, I luckily still had my PR branches around!

cmb69:
Thank you!

PR #504, opened 2022-03-15
Fix CSS style to also apply to destructorsynopsis

Girgias:
See:
https://www.php.net/manual/en/streamwrapper.destruct.php
and
https://www.php.net/manual/en/streamwrapper.construct.php

For what this is trying to fix.


cmb69 (approved):
Thank you!

PR #514, opened 2022-05-13
Add supported versions in json payload

renanbr:


releases/index.php
```diff
@@ -8,6 +8,11 @@
 
 	$machineReadable = [];
 
+    $supportedVersions = [];
```
ramsey, line 11: Can you change the space indents to tab indents to match the rest of the file? This will make it easier to review the diff on the PR.

Thanks!

ramsey (changes requested):

ramsey (approved):

PR #519, opened 2022-05-27
Change escaping function to build URL for searching

### LINKED ISSUE #518: Non-ASCII text searching from search bar is not working rightly.
## Description

There is case that does not get right result with search using non-ascii keyword.

## Steps to reproduce

On browser.

1. Access [PHP website](https://www.php.net/) 
1. Input Japanese text on search-input on header(ex: `コンストラクタ` = `Constructor`)
1. Click `Search php.net for コンストラクタ` under search-input. (see image)
1. Wait to display `/results.php`

## Actually

In `/results.php`

- Put `%u30B3%u30F3%u30B9%u30C8%u30E9%u30AF%u30BF` into textarea under page title
- Search-result is not found (display no results)

https://user-images.githubusercontent.com/668834/170723548-35a1df66-0150-466d-a162-1d2b4da595b2.mp4

## Expected

In `/results.php`

- Put `コンストラクタ` into textarea under page title
- Display many search-results.

## Notes

I will create PR by simple fixes.

attakei:
## Overview 

Fix #518

## Note

Search form submits raw-text to `/search.php`, but `search.js` build url with escaped text.
I think that it need not escape on JS too.

Ref: https://github.com/php/web-php/blob/master/include/header.inc#L127

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -360,7 +360,7 @@
 
                         dropdown.append(searchTemplate.render({
                             pattern: pattern,
-                            url: "/search.php?pattern=" + escape(pattern)
+                            url: "/search.php?pattern=" + pattern
                         }));
 
                         /* If the dropdown is hidden (because there are no
--- a/js/search.js
+++ b/js/search.js
@@ -360,7 +360,7 @@
 
                         dropdown.append(searchTemplate.render({
                             pattern: pattern,
-                            url: "/search.php?pattern=" + pattern
+                            url: "/search.php?pattern=" + encodeURIComponent(pattern)
                         }));
 
                         /* If the dropdown is hidden (because there are no


js/search.js
```diff
@@ -360,7 +360,7 @@
 
                         dropdown.append(searchTemplate.render({
                             pattern: pattern,
-                            url: "/search.php?pattern=" + escape(pattern)
+                            url: "/search.php?pattern=" + pattern
```
cmb69, line 363: I think we need to escape, but we should not use the very old `escape()` function, but rather `encodeURIComponent()` instead.

attakei: Thanks for feedback.
I seem surely your feedback is right.

I will commit for using `encodedURIComponent`.
Can I commit as ammend and force push.

cmb69: Yes, you can amend and force-push, but you can also add a new commit and push. We usually squash before merging anyway.

PR #520, opened 2022-06-06
Change the function not found mark up

georanma:
Update the function not found mark up to call out further that the search results is for functions, and that a "full site search" can be reached for their query at the link provided. The list of closest matches is still presented below.

### DIFF
--- a/quickref.php
+++ b/quickref.php
@@ -112,7 +112,7 @@ site_header("Manual Quick Reference", $head_options+array("current" => "help"));
 
 // Note: $notfound is defined (with htmlspecialchars) inside manual-lookup.php
 $notfound_enc = urlencode($notfound);
-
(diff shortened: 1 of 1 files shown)


georanma:
This does not necessarily resolve all UI issues. See example screenshot

image
Sort is a function, and it shows in the drop down results, but 'enter' gets the user to a "not found" page.

thinkverse:
Love this. It'll make the full website search more prominent. As it stands now, eyes are drawn to the list of closest matches when a search result isn't found and the link to the full website search is tucked away hidden in the sidebar.

> Sort is a function, and it shows in the drop down results, but 'enter' gets the user to a "not found" page.

Cannot recreate that issue on php.net, it redirects to the manual page as intended. Could it be a local issue only?

PR #521, opened 2022-06-09
Remove names of contributors from changelog

kamil-tekiela:
It appears to be a problem with a tool since the names don't appear anywhere else


cmb69 (approved):
Yeah, the messages need a period after the closing parenthesis so that the authors are removed.

Thanks for the fix!

PR #525, opened 2022-06-17
Static lambda

MathiasReker:
Lambdas not (indirect) referencing $this must be declared static.


localheinz:
@cmb69 

Yes, these lambdas can then not be bound to an object.

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/function_notation/static_lambda.rst.

sgolemon:
> Yes, these lambdas can then not be bound to an object.

These are all literally temporary lambdas though, what object are we worried about them being bound to?

I mean, it's fine, the `static` is harmless here, it's just a little bit of syntactic noise after all, but it doesn't actually *do* anything effective.

PR #526, opened 2022-06-17
Declare strict types

MathiasReker:
Force strict types declaration in all files.


cmb69:
I wonder *how much* of the functionality that would break.

MathiasReker:
> I wonder _how much_ of the functionality that would break.

No reason to risk any breaking.

cmb69:
I mean, I'm not opposed to declare strict_types, but we would need to make sure that nothing breaks. And unfortunately, we have no tests, as far as I know, nor any static analysis.

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/strict/declare_strict_types.rst.

PR #527, opened 2022-06-17
Modernize strpos

MathiasReker:
Replace strpos() calls with str_starts_with() or str_contains() if possible.


cmb69:
What is the minimum PHP version we are supporting? `str_start_with()` is only available as of PHP 8.0.0.

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/alias/modernize_strpos.rst.

PR #529, opened 2022-06-17
List syntax

MathiasReker:
List (array destructuring) assignment should be declared using the configured syntax.

### DIFF
--- a/cal.php
+++ b/cal.php
@@ -354,13 +354,13 @@ function read_event($fp)
     if (count($linearr) < 13) { return FALSE; }
 
     // Get components
-    list(
+    [
         $day, $month, $year, $country,
         $sdesc, $id, $ldesc, $url, $recur, $tipo, $sdato, $edato, $category
-    ) = $linearr;
+    ] = $linearr;
 
     // Get info on recurring event
-    @list($recur, $recur_day) = explode(":", $recur, 2);
+    @[$recur, $recur_day] = explode(":", $recur, 2);
 
     // Return with SQL-resultset like array
     return array(
--- a/index.php
+++ b/index.php
@@ -109,7 +109,7 @@ foreach ($active_branches as $major => $releases) {
     krsort($releases);
     foreach ((array)$releases as $release) {
         $version = $release['version'];
-        list($major, $minor, $_) = explode('.', $version);
+        [$major, $minor, $_] = explode('.', $version);
         $intro .= "
             <li class='hero-version'><a class='hero-version-link' href='/downloads.php#v$version'>$version</a> &middot; <a class='notes' href='/ChangeLog-$major.php#$version'>Changelog</a> &middot; <a class='notes' href='/migration$major$minor'>Upgrading</a></li>\n";
     }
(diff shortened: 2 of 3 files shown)


cal.php
```diff
@@ -354,13 +354,13 @@ function read_event($fp)
...
-    list(
+    [
         $day, $month, $year, $country,
         $sdesc, $id, $ldesc, $url, $recur, $tipo, $sdato, $edato, $category
-    ) = $linearr;
+    ] = $linearr;
 
     // Get info on recurring event
-    @list($recur, $recur_day) = explode(":", $recur, 2);
+    @[$recur, $recur_day] = explode(":", $recur, 2);
```
kamil-tekiela, line 363: Why is there an error suppression operator? What could throw a warning here?

index.php
```diff
@@ -109,7 +109,7 @@
     krsort($releases);
     foreach ((array)$releases as $release) {
         $version = $release['version'];
-        list($major, $minor, $_) = explode('.', $version);
+        [$major, $minor, $_] = explode('.', $version);
```
kamil-tekiela, line 112: Correct me if I am wrong, but `$_` isn't used anywhere. Is this a Perl force of habit?

localheinz: @kamil-tekiela 

I believe this is out of the scope of the change proposed by @MathiasReker.

localheinz (commented):
👍

index.php
localheinz: @kamil-tekiela 

I would not consider that good practice - it would be best to apply atomic commits.

The code in this repository is a can of worms, where would we stop?

cmb69: > The code in this repository is a can of worms, where would we stop?

Maybe before opening the can? ;)

cal.php
cmb69: I'm not sure whether we shouldn't drop cal.php altogether. Besides that it is obviously broken (e.g. there are records where addslashes() has been applied, or is still being applied), it feels strange to me to list user group meetings which may no longer be active. See, for instance, https://www.php.net/cal.php?id=6873&cm=06&cy=2022, which links to Website which is no longer available. And apparently nobody cares about those expirations. And https://www.php.net/cal.php?id=3560&cm=06&cy=2022, for instance, looks like spam to me.

cmb69: I'm not sure whether we shouldn't drop cal.php altogether. Besides that it is obviously broken (e.g. there are records where addslashes() has been applied, or is still being applied), it feels strange to me to list user group meetings which may no longer be active. See, for instance, https://www.php.net/cal.php?id=6873&cm=06&cy=2022, which links to Website which is no longer available. And apparently nobody cares about those expirations. And https://www.php.net/cal.php?id=3560&cm=06&cy=2022, for instance, looks like spam to me.

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/list_notation/list_syntax.rst.

index.php
morrisonlevi: I agree that atomic commits are better for projects like this. It is a never-ending can of worms if you don't do it this way.

PR #530, opened 2022-06-17
No alias functions

MathiasReker:
Master functions shall be used instead of aliases.

### DIFF
--- a/manual/spam_challenge.php
+++ b/manual/spam_challenge.php
@@ -39,7 +39,7 @@ $challenges = array(
 // generate a challenge
 function gen_challenge() {
 	global $challenges, $nums;
-	$c = $challenges[rand(0, sizeof($challenges)-1)];
+	$c = $challenges[rand(0, count($challenges)-1)];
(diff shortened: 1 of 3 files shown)


localheinz (commented):
👍

cmb69:
I think using the canonical function names makes sense. Any particular reason for not doing this @kamil-tekiela?

kamil-tekiela (approved):
No reason for not doing it. It won't change anything.

cmb69:
> It won't change anything.

It might make the code a little bit easier to grasp for those not used to the aliases. :)

tontonsb:
These (join and sizeof) aliases sometimes improve readability. Similar to how in html `article`, `section` and `main` are functionally [nearly] the same but carry semantics.

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/alias/no_alias_functions.rst.

PR #531, opened 2022-06-17
Short scalar cast

MathiasReker:
Cast (boolean) and (integer) should be written as (bool) and (int), (double) and (real) as (float), (binary) as (string).

### DIFF
--- a/submit-event.php
+++ b/submit-event.php
@@ -7,7 +7,7 @@ site_header("Submit an Event", array("current" => "community"));
 
 // No errors, processing depends on POST data
 $errors = array();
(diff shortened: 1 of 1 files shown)


submit-event.php
```diff
@@ -7,7 +7,7 @@
 
 // No errors, processing depends on POST data
 $errors = array();
-$process = (boolean) count($_POST);
+$process = (bool) count($_POST);
```
kamil-tekiela, line 10:
```suggestion
$process = [] !== $_POST;
```

kamil-tekiela (approved):
Thanks, LGTM

PR #533, opened 2022-06-17
Random api migration

MathiasReker:
Replaces rand, srand, getrandmax functions calls with their mt_* analogs or random_int.


kamil-tekiela:
It's still the same. Why change it if it has no impact on anything?

MathiasReker:
rand is a alias function of mt_rand since PHP 7.1. I prefer not to use alias functions.

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/alias/random_api_migration.rst.

PR #534, opened 2022-06-17
Elseif

MathiasReker:
The keyword elseif should be used instead of else if so that all control keywords look like single words.


cmb69:
Thank you!

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/control_structure/elseif.rst.

PR #536, opened 2022-06-17
Use arrow functions

MathiasReker:
Anonymous functions with one-liner return statement must use arrow functions.


kamil-tekiela:
It needs to support PHP 7.2

kelunik:
@kamil-tekiela is that still the case? In that case we should look into upgrading.

PR #537, opened 2022-06-17
Constant case

MathiasReker:
The PHP constants true, false, and null MUST be written using the correct casing.


cmb69:
> The PHP constants true, false, and null MUST be written using the correct casing.

RIght. However, the correct casing depends on the coding style.

localheinz (commented):
👍

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/casing/constant_case.rst.

cmb69:
Closing in favor of #660 which just has been merge. Thanks for the PR nonetheless!

PR #538, opened 2022-06-17
Line ending

MathiasReker:
All PHP files must use same line ending.


cmb69:
Yeah, makes sense to always use LF only; could you please resolve the merge conflicts?

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/whitespace/line_ending.rst.

PR #539, opened 2022-06-17
No closing tag

MathiasReker:
The closing ?> tag MUST be omitted from files containing only PHP.

### DIFF
--- a/credits.php
+++ b/credits.php
@@ -26,5 +26,3 @@ if ($credits) {
     echo $credits;
     site_footer();
 }
-
-?>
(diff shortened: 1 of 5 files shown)


kamil-tekiela:
Makes sense, but I would keep a newline at the end of the file.

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/php_tag/no_closing_tag.rst.

kamil-tekiela:
Superseded by #603

PR #540, opened 2022-06-17
New with braces

MathiasReker:
All instances created with new keyword must (not) be followed by braces.

### DIFF
--- a/images/supported-versions.php
+++ b/images/supported-versions.php
@@ -154,7 +154,7 @@ $height = $header_height + $footer_height + (count($branches) * $branch_height);
 	<!-- Today -->
 	<g class="today">
 		<?php
-		$now = new DateTime;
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
Unless our coding standard mandates this, I wouldn't change it. It improves nothing.

MathiasReker:
This is the only place that differs.

localheinz (commented):
👍

cmb69:
No, we don't have agreed on a CS. That should be discussed on the [PHP php.net internal infrastructure discussion ML](https://www.php.net/mailing-lists.php#internals).

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/operator/new_with_braces.rst.

PR #541, opened 2022-06-17
Single blank line at eof

MathiasReker:
A PHP file without end tag must always end with a single empty line feed.


kamil-tekiela (approved):
Looks good but one of your other PRs did the exact opposite. Could you not make a single PR with all the changes and also commit the CS-fixer file?

localheinz:
@kamil-tekiela @MathiasReker 

How do you feel about adding an [`.editorconfig`](https://editorconfig.org)?

```ini
root = true

[*]
insert_final_newline = true

localheinz (commented):
👍

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/whitespace/single_blank_line_at_eof.rst.

PR #542, opened 2022-06-17
Combine consecutive issets

MathiasReker:
Using isset($var) && multiple times should be done in one call.

### DIFF
--- a/manual/vote-note.php
+++ b/manual/vote-note.php
@@ -42,7 +42,7 @@ if ($_SERVER['REQUEST_METHOD'] == 'POST') {
         if (json_last_error() == JSON_ERROR_NONE && isset($r->status) && $r->status && isset($r->votes)) {
           $response["success"] = true;
           $response["update"] = (int)$r->votes;
-        } elseif (json_last_error() == JSON_ERROR_NONE && isset($r->status) && isset($r->message) && $r->status == false) {
+        } elseif (json_last_error() == JSON_ERROR_NONE && isset($r->status, $r->message)   && $r->status == false) {
           $response["success"] = false;
           $response["msg"] = $r->message;
         } else {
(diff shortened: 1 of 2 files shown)


manual/vote-note.php
```diff
@@ -42,7 +42,7 @@
         if (json_last_error() == JSON_ERROR_NONE && isset($r->status) && $r->status && isset($r->votes)) {
           $response["success"] = true;
           $response["update"] = (int)$r->votes;
-        } elseif (json_last_error() == JSON_ERROR_NONE && isset($r->status) && isset($r->message) && $r->status == false) {
```
kamil-tekiela, line 45: Please trim whitespace in both places.

my.php
```diff
@@ -10,7 +10,7 @@
 $options = array();
 
 // We have post data, and it is an available language
-if (isset($_POST['my_lang']) && isset($langs[$_POST['my_lang']])) {
+if (isset($_POST['my_lang'], $langs[$_POST['my_lang']])  ) {
```
localheinz, line 13:
```suggestion
if (isset($_POST['my_lang'], $langs[$_POST['my_lang']])) {
```

manual/vote-note.php
```diff
@@ -42,7 +42,7 @@
...
+        } elseif (json_last_error() == JSON_ERROR_NONE && isset($r->status, $r->message)   && $r->status == false) {
```
localheinz, line 45:
```suggestion
        } elseif (json_last_error() == JSON_ERROR_NONE && isset($r->status, $r->message) && $r->status == false) {
```

manual/vote-note.php
MathiasReker: Thanks @localheinz

cmb69:
Thank you!

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/language_construct/combine_consecutive_issets.rst.

PR #544, opened 2022-06-17
Fopen flags

MathiasReker:
The flags in fopen calls must omit t, and b must be omitted or included consistently.


kamil-tekiela:
Thanks, but this doesn't look very useful. The security section isn't used and the code should be removed. The `fopen` in cal.php don't require `b`  because as far as I know, it's the default.

PR #545, opened 2022-06-17
Increment style

MathiasReker:
Pre- or post-increment and decrement operators should be used if possible.


cmb69:
I think this is mostly a style preference, so I've labeled with `CS`.

localheinz (commented):
👍

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/operator/increment_style.rst.

PR #546, opened 2022-06-17
Include

MathiasReker:
Include/Require and file path should be divided with a single space. File path should not be placed under brackets.

### DIFF
--- a/manual/add-note.php
+++ b/manual/add-note.php
@@ -5,7 +5,7 @@ $_SERVER['BASE_PAGE'] = 'manual/add-note.php';
 include_once __DIR__ . '/../include/prepend.inc';
 include_once __DIR__ . '/../include/posttohost.inc';
 include_once __DIR__ . '/../include/shared-manual.inc';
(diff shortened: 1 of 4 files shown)


manual/add-note.php
```diff
@@ -5,7 +5,7 @@
 include_once __DIR__ . '/../include/prepend.inc';
 include_once __DIR__ . '/../include/posttohost.inc';
 include_once __DIR__ . '/../include/shared-manual.inc';
-include      __DIR__ . '/spam_challenge.php';
```
kamil-tekiela, line 8: Please keep it as it was here. It was more readable.

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/control_structure/include.rst.

manual/add-note.php
MathiasReker: fixed

kamil-tekiela (approved):

PR #547, opened 2022-06-17
No alternative syntax

MathiasReker:
Replace control structure alternative syntax to use braces.

### DIFF
--- a/downloads.php
+++ b/downloads.php
@@ -39,11 +39,11 @@ site_header("Downloads",
     )
 );
 ?>
-<?php $i = 0; foreach ($RELEASES as $MAJOR => $major_releases): /* major releases loop start */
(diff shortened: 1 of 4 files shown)


kamil-tekiela:
Why are you forcing this syntax here? Has this been discussed somewhere?

MathiasReker:
It is better practice. I aim to improve the overall code quality

kamil-tekiela:
> It is better practice. I aim to improve the overall code quality

In what way is it better practice? Please explain. This change doesn't improve code quality IMHO.

MathiasReker:
I get your point of not fixing this in inline HTML.

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/control_structure/no_alternative_syntax.rst.

PR #548, opened 2022-06-17
No useless else

MathiasReker:
There should not be useless else cases.

### DIFF
--- a/cal.php
+++ b/cal.php
@@ -222,11 +222,11 @@ function date_for_recur($recur, $day, $bom, $eom)
     }
 
     // ${recur}th to last $day of the month
-    else {
+     
         $eomd = date("w",$eom) + 1;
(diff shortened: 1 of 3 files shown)


cal.php
```diff
@@ -222,11 +222,11 @@ function date_for_recur($recur, $day, $bom, $eom)
     }
 
     // ${recur}th to last $day of the month
-    else {
+     
         $eomd = date("w",$eom) + 1;
```
kamil-tekiela, line 226: Please indent the code properly.

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/control_structure/no_useless_else.rst.

kamil-tekiela:
Superceded by #603

PR #549, opened 2022-06-17
No superfluous elseif

MathiasReker:
Replaces superfluous elseif with if.


kamil-tekiela (approved):

cmb69:
Thank you!

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/control_structure/no_superfluous_elseif.rst.

PR #550, opened 2022-06-17
No useless return

MathiasReker:
There should not be an empty return statement at the end of a function.

### DIFF
--- a/releases/8.0/common.php
+++ b/releases/8.0/common.php
@@ -6,7 +6,7 @@ include_once __DIR__ . '/../../include/prepend.inc';
 function language_redirect(string $currentLang): void {
     // We don't use the general language selection of php.net,
     // so soldier on with this one.
-    return;
+    
 }
 
 function common_header(string $description): void {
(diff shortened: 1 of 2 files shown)


releases/8.0/common.php
```diff
@@ -6,7 +6,7 @@
 function language_redirect(string $currentLang): void {
     // We don't use the general language selection of php.net,
     // so soldier on with this one.
-    return;
+
```
kamil-tekiela, line 9: Please remove blank lines

releases/8.1/common.php
```diff
@@ -7,7 +7,7 @@
 function language_redirect(string $currentLang): void {
     // We don't use the general language selection of php.net,
     // so soldier on with this one.
-    return;
+
```
MathiasReker, line 10:
```suggestion
```

releases/8.0/common.php
MathiasReker: Fixed

cmb69:
Shouldn't we better remove these functions and their calls altogether?

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/return_notation/no_useless_return.rst.

kamil-tekiela:
Obsolete due to https://github.com/php/web-php/pull/562

PR #551, opened 2022-06-17
No spaces inside parenthesis

MathiasReker:
There MUST NOT be a space after the opening parenthesis. There MUST NOT be a space before the closing parenthesis.


MathiasReker:
If there is no code style, then at least let's be consistent. This is the only place that differs. :-)

cmb69:
> If there is no code style, then at least let's be consistent. This is the only place that differs. :-)

Okay, consistency is a good argument. I will update the commit message accordingly.

Anyhow, thanks for the PR!

localheinz:
For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/doc/rules/whitespace/no_spaces_inside_parenthesis.rst.

PR #499, opened 2022-02-24
Combine EOL date and period columns

SteveDesmond-ca:
This PR proposes a fix for #498 by combining the EOL date and period columns, placing the period on a new line below the date and enclosed in parentheses.

The amended layout reduces the page's minimum non-breaking viewport width from 408px to 352px, which should account for the vast majority of modern smartphones (manufactured 2016 or later).

PR #557, opened 2022-06-27
Fix: Do not use `vertical-align: middle` for table cells

localheinz:
This pull

- [x] stops overriding the default `vertical-align: top` with `vertical-align :middle` for table cells

Somewhat related to #499.

💁‍♂️ For reference, see https://github.com/php/web-php/blob/144acf6e9cf903bf694e531e84dd9201d520dee0/styles/theme-base.css#L683-L685

### Before

CleanShot 2022-06-27 at 15 27 48@2x

### After

CleanShot 2022-06-27 at 15 28 04@2x

❗ This affects other tables using `class="standard"` as well.

### DIFF
--- a/styles/theme-base.css
+++ b/styles/theme-base.css
@@ -706,7 +706,6 @@ table.standard th {
 
 table.standard td {
     padding: 5px 10px;
-    vertical-align: middle;
 }
 
 table.standard tr:nth-child(even) td.subr,


cmb69:
Hmm, looks strange on https://www.php.net/docs.php (although that shouldn't be a table, in my opinion).

cmb69:
> Perhaps it would make sense to get rid of the `standard` class, too. There are not that many tables using the `standard` class. Default seems standard enough!

Maybe. But regarding this table, see PR #560.

morrisonlevi:
Hmm. For what it's worth, on the EOL page at least, I definitely prefer the `vertical-align: middle`. It seems easier to read for whatever reason.

localheinz:
Closing in favour of #499.

PR #558, opened 2022-06-27
Fix: Remove superfluous note

localheinz:
This pull request

- [x] removes a superfluous note

Slightly related to #499.

💁‍♂️ If we were consistent, we would have two options:

- add a similar note to all other major versions that are end of life
- remove this note


cmb69:
Makes perfect sense. Thank you!

PR #175, opened 2017-09-27
.editorconfig

petk:
Hello,

this patch adds [.editorconfig file](http://editorconfig.org) for consistency and code beautification in editors that use it and to sync the coding style used in this particular repository.

Coding styles in this commit are set to the following defaults:

All mentioned file types are set to:
* UTF-8 encoding
* Unix-style newlines with a newline ending
* trailing whitespace is trimmed

MarkDown files
* trimming of trailing whitespace is disabled since there might appear some to indicate a new line

CSS, JavaScript, HTML, XML files:
* indentation 2 spaces

PHP files (*.php and *.inc and few bin/ PHP CLI scripts)
* indentation 4 spaces

In case this is inconvenient for people maintaining this repository, let me know and I can adjust it accordingly. The picked settings are somehow most used in other open source code (PHP files somehow mostly use 4 spaces in open source libraries and frameworks), JS and CSS (from my point of view, use 2 spaces).

PR #559, opened 2022-06-27
Enhancement: Run `friendsofphp/php-cs-fixer` on GitHub Actions

localheinz:
This pull request

- [x] runs `friendsofphp/php-cs-fixer` on GitHub Actions

💁‍♂️ To get the ball rolling, here is a proposal for running `friendsofphp/php-cs-fixer` to enforce a coding standard on GitHub Actions.

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -9,6 +9,49 @@ on:
       - "master"
 
 jobs:
+  coding-standards:
+    name: "Coding Standards"
+
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.3"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v3"
+
+      - name: "Set up PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          extensions: "none, json, mbstring, tokenizer"
+          php-version: "${{ matrix.php-version }}"
+
+      - name: "Set up problem matchers for PHP"
+        run: "echo \"::add-matcher::${{ runner.tool_cache }}/php.json\""
+
+      - name: "Validate composer.json and composer.lock"
+        run: "composer validate --ansi --strict"
+
+      - name: "Determine composer cache directory"
+        run: "echo \"COMPOSER_CACHE_DIR=$(composer config cache-dir)\" >> $GITHUB_ENV"
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v3"
+        with:
+          path: "${{ env.COMPOSER_CACHE_DIR }}"
+          key: "php-${{ matrix.php-version }}-composer-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --ansi --no-interaction --no-progress"
+
+      - name: "Run friendsofphp/php-cs-fixer"
+        run: "vendor/bin/php-cs-fixer fix --ansi --config=.php-cs-fixer.php --diff --dry-run --verbose"
+
   tests:
     name: "Tests"
 
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -1,0 +1,14 @@
+<?php
+
+$finder = PhpCsFixer\Finder::create()->in(__DIR__);
+
+$config = new PhpCsFixer\Config();
+
+$finder = $config->getFinder()
+    ->ignoreDotFiles(false)
+    ->in(__DIR__)
+    ->name(__FILE__);
+
+$config->setRules([]);
+
+return $config;
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -67,6 +67,13 @@ about what you're working on, you can contact us via the
   ```
 
   to check your change doesn't break other features.
+- Run
+
+  ```
+  make coding-standards
+  ```
+
+  to automatically fix coding standard issues.
 - Review the change once more just before submitting it.
 
 ## What happens after submitting contribution?
@@ -119,6 +126,12 @@ Having said that, here are the organizational rules:
 
    ```
    make tests
+   ```
+
+5. Fix coding standard issues before committing code. To do so use
+
+   ```
+   make coding-standards
    ```
 
 6. Use reasonable commit messages.
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ backend/mirror.gif
 backend/mirror.png
 backend/mirror.jpg
 backend/GeoIP.dat
+vendor/
 .idea
 .DS_Store
 .DS_Store?
+.php-cs-fixer.cache
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,14 @@
 help: ## Displays this list of targets with descriptions
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[32m%-30s\033[0m %s\n", $$1, $$2}'
 
+.PHONY: coding-standards
+coding-standards: vendor ## Fixes code style issues with friendsofphp/php-cs-fixer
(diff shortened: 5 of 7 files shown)


.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,37 @@
...
+jobs:
+  coding-standards:
+    name: "Coding Standards"
+
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.4"
```
localheinz, line 20: I am aware of [`README.md`](https://github.com/php/web-php/tree/c40cac09f2b7d50e191ab6d474cf015032068e0a#readme) suggesting to use PHP 7.2, but 7.4 is the lowest version supported by [`friendsofphp/php-cs-fixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/composer.json#L17).

Yes, we could use an outdated version, but why bother?

.phive/phars.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
localheinz, line 1: We could also use

- `composer.json` in the root directory (see #408)
- a tool-specific `composer.json` in a separate directory
- check in `php-cs-fixer.phar` manually
- use the [tool support of `shivammathur/setup-php`](https://github.com/shivammathur/setup-php#wrench-tools-support)

.php-cs-fixer.php
```diff
@@ -0,0 +1,14 @@
...
+$finder = PhpCsFixer\Finder::create()->in(__DIR__);
+
+$config = new PhpCsFixer\Config();
+
+$finder = $config->getFinder()
+    ->ignoreDotFiles(false)
+    ->in(__DIR__)
+    ->name('.php-cs-fixer.php');
+
+$config->setRules([]);
```
localheinz, line 12: Using `[]` here will at least lint all PHP files.

That's a good starting point.

From there on, rule sets or fixers could be enabled.

As pointed out in the pull requests opened by @MathiasReker, a lot of fixers exist. 

As pointed out in my email to the mailing list, all available tools have advantages and disadvantages. Not using a tool, though, has the largest disadvantage.

.php-cs-fixer.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+$finder = PhpCsFixer\Finder::create()->in(__DIR__);
+
+$config = new PhpCsFixer\Config();
+
+$finder = $config->getFinder()
+    ->ignoreDotFiles(false)
+    ->in(__DIR__)
+    ->name('.php-cs-fixer.php');
```
localheinz, line 10: `ignoreDotFiles(false)` and `name('.php-cs-fixer.php')` will also lint and fix this file here.

Dot files and certain directories are excluded by default:

- https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.8.0/src/Finder.php#L32
- https://github.com/symfony/finder/blob/v5.4.8/Finder.php#L68

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,37 @@
+# https://docs.github.com/en/actions
+
+name: "Integrate"
+
+on:
+  pull_request: null
```
localheinz, line 6: This will trigger a build for pull requests.

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,37 @@
...
+  push:
+    branches:
+      - "master"
```
localheinz, line 9: This will trigger a build for pushes to the `master` branch.

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  coding-standards:
```
localheinz, line 12: Using 2 spaces for indentation - indentation of YAML files with 4 spaces tends to be rather awkward.

I would document it in `.editorconfig` - also see #175.

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.4"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v3"
```
localheinz, line 24: See https://github.com/actions/checkout.

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: "Set up PHP"
+        uses: "shivammathur/setup-php@v2"
```
localheinz, line 27: See https://github.com/shivammathur/setup-php.

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          coverage: "none"
+          extensions: "none, json, tokenizer"
+          php-version: "${{ matrix.php-version }}"
```
localheinz, line 31: We do not need to use a matrix here, since we only have a single job anyway, but it is rather convenient.

Want to change the PHP version?

Modify the matrix:

```diff
diff --git a/.github/workflows/integrate.yaml b/.github/workflows/integrate.yaml
index dcb77b91..dbe7778e 100644
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.4"
+          - "8.1"

     steps:
       - name: "Checkout"
```

It is easier to search and replace `"7.4"` instead of `7.4`.

.github/workflows/integrate.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: "Set up problem matchers for PHP"
+        run: "echo \"::add-matcher::${{ runner.tool_cache }}/php.json\""
```
localheinz, line 34: See https://github.com/shivammathur/setup-php#php.

.github/workflows/integrate.yaml
```diff
@@ -9,6 +9,33 @@ on:
       - "master"
 
 jobs:
+  coding-standards:
```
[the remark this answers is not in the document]
localheinz: @cmb69 

The advantage of a separate job is that

- you can immediately see what aspect of the build failed
- other jobs (i.e., tests) might still pass

.github/workflows/integrate.yaml
```diff
@@ -9,6 +9,49 @@ on:
...
+        with:
+          coverage: "none"
+          extensions: "none, json, mbstring, tokenizer"
+          php-version: "${{ matrix.php-version }}"
+
+      - name: "Set up problem matchers for PHP"
+        run: "echo \"::add-matcher::${{ runner.tool_cache }}/php.json\""
+
+      - name: "Validate composer.json and composer.lock"
+        run: "composer validate --ansi --strict"
```
localheinz, line 37: Validating `composer.json` and `composer.lock` here to ensure - among others - we are not working with `composer.json` out of sync with `composer.lock`.

.github/workflows/integrate.yaml
```diff
@@ -9,6 +9,49 @@ on:
...
+
+      - name: "Determine composer cache directory"
+        run: "echo \"COMPOSER_CACHE_DIR=$(composer config cache-dir)\" >> $GITHUB_ENV"
```
localheinz, line 40: See https://getcomposer.org/doc/03-cli.md#composer-cache-dir.

.github/workflows/integrate.yaml
```diff
@@ -9,6 +9,49 @@ on:
...
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v3"
```
localheinz, line 43: See https://github.com/actions/cache.

.php-cs-fixer.php
staabm, line 10:
```suggestion
    ->name(__FILE__);
```

.github/workflows/integrate.yaml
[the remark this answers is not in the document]
cmb69: Fair enough!

cmb69 (approved):
I think this is good to be merged. Afterwards we can add some actual rules (or rule sets) in separate PRs.

Thank you for the PR!

localheinz:
@cmb69 

In my experience, it's best to enable one fixer at a time - enabling a bunch of them (or a rule set) can become difficult to review.

localheinz:
Thank you, @cmb69!

PR #560, opened 2022-06-27
Drop useless documentation table

cmb69:


localheinz:
### Before

CleanShot 2022-06-27 at 19 59 39@2x

### After

CleanShot 2022-06-27 at 19 59 30@2x

kamil-tekiela (approved):

cmb69:
> There isn't actually any documentation that can be downloaded.
> 
> See http://localhost:8080/download-docs.php:

For local clones, you'd need to put respective files under distributions/manual/.

PR #562, opened 2022-06-27
Fix: Remove useless function

localheinz:
This pull request 

- [x] removes the useless function `language_redirect()`


kamil-tekiela (approved):
I have no idea why we even had that there. It's really useless.

cmb69:
Thank you!

PR #563, opened 2022-06-27
Fix: Remove unused references to a global variable

localheinz:
This pull request

- [x] removes unused references to the global variable `$MYSITE`

### DIFF
--- a/include/do-download.inc
+++ b/include/do-download.inc
@@ -25,8 +25,6 @@ function get_actual_download_file($file)
 // Download a file from a mirror site
 function download_file($mirror, $file)
 {
-    global $MYSITE;
-
     // Redirect to the particular file
     if (!headers_sent()) {
         status_header(302);
--- a/include/layout.inc
+++ b/include/layout.inc
@@ -524,7 +524,6 @@ function site_header($title = '', $config = array())
 }
 function site_footer($config = array())
 {
-    global $MYSITE;
     require __DIR__ . "/footer.inc";
 }
 


kamil-tekiela:
Have you checked for side-effects? The keyword `global` has side-effects. https://3v4l.org/b9kH9 I sincerely doubt that this code relied on side-effects, but it could be wise to verify.

include/do-download.inc
```diff
@@ -25,8 +25,6 @@ function get_actual_download_file($file)
 // Download a file from a mirror site
 function download_file($mirror, $file)
 {
-    global $MYSITE;
```
localheinz, line 28: Variable is not referenced in scope:

https://github.com/php/web-php/blob/8fbbe3bb896b3e93a745f4938dee97fc337a4ecd/include/do-download.inc#L26-L42

include/layout.inc
```diff
@@ -484,8 +484,6 @@ EOT;
 
 function site_header($title = '', $config = array())
 {
-    global $MYSITE;
```
localheinz, line 487: Variable is not referenced in scope:

https://github.com/php/web-php/blob/8fbbe3bb896b3e93a745f4938dee97fc337a4ecd/include/layout.inc#L485-L524

include/layout.inc
```diff
@@ -524,7 +522,6 @@ function site_header($title = '', $config = array())
 }
 function site_footer($config = array())
 {
-    global $MYSITE;
```
localheinz, line 527: Variable is not referenced in scope:

https://github.com/php/web-php/blob/8fbbe3bb896b3e93a745f4938dee97fc337a4ecd/include/layout.inc#L525-L529

localheinz (commented):
Unless there is some magic related to globals involved that I am unaware of, the global variable `@MYSITE` is not referenced here.

kamil-tekiela:
I know it's not referenced in scope, but it's still a keyword that has a global effect. If the variable is not defined in the global scope it sets its value to `null`. That's why I am saying it has a side-effect. See my example above. If you remove this declaration we could possibly get "undefined variable" errors somewhere else in the code.

dragoonis:
This code gave me nightmares, lol

Trying to remove it but then has other global side effects.

I'm glad you guys are here trying to sort this stuff out 💙

include/do-download.inc
kamil-tekiela: This one is safe to remove. It's not used between declaration and `exit`.

include/layout.inc
kamil-tekiela: This one is used in its current scope and cannot be removed.

include/layout.inc
kamil-tekiela: This one is 99.99% safe. It's not used in `footer.inc` and that file is meant to be the last thing that is executed. However, given how convoluted this codebase is, it's still possible that some code relies on this declaration. I'd say we risk removing it either way and fix side effect if it occurs.

kamil-tekiela (changes requested):
Other than the middle one, the rest is good.

include/layout.inc
cmb69: Good catch! I hope you're doing well after having checked this code mess. :)

kamil-tekiela (approved):
Thanks.

localheinz:

PR #564, opened 2022-06-27
Fix: Heredoc syntax

localheinz:
This pull request

- [x] fixes heredoc syntax issues

### DIFF
--- a/include/layout.inc
+++ b/include/layout.inc
@@ -453,7 +453,7 @@ function print_news($news, $dog, $max = 5, $onlyyear = null, $return = false) {
             continue;
         }
 
-        echo <<< EOT
+        echo <<<EOT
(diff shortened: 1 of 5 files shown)


cmb69:
TIL: https://3v4l.org/8RVCp. Fascinating.

cmb69:
I mean, apparently, having spaces/tabs between the `<<<` and the label is perfectly valid (although undocumented in the PHP manual), so removing this whitespace would be more like a CS fix. However, I can't find a PHP-CS-Fixer fix for this.

cmb69:
Well, let's merge this. I don't think that anybody actually could prefer that notation.

Thank you, @localheinz!

localheinz:
Thank you, @cmb69!

PR #567, opened 2022-06-27
Redesign Docs.php

kamil-tekiela:
The information about PHP 4 & 5 docs while still useful can be moved to "more documentation" section where it belongs more. 

The information about URL wasn't grammatically correct IMHO and also kind of out of place where it was. I moved it to the bottom of the same section.

### DIFF
--- a/docs.php
+++ b/docs.php
@@ -15,11 +15,6 @@ site_header("Documentation", array("current" => "docs"));
 </p>
 
 <p>
- More information about php.net URL shortcuts by visiting our
- <a href="urlhowto.php">URL howto page</a>.
-</p>
-
-<p>
  Note, that many languages are just under translation, and
  the untranslated parts are still in English. Also some translated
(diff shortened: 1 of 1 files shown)


docs.php
```diff
@@ -84,6 +74,13 @@
   The same applies to the <a href="/manual/php3.php">PHP 3 Manual</a>, and
   the <a href="/manual/php4.php">PHP 4 Manual</a>.
  </li>
+ <li>
+  The PHP 4 and PHP 5 documentation has been removed from the
+  manual, but archived versions still exist. For
+  more information, please read <a href="/manual/php4.php">
+  Documentation for PHP 4</a> and <a href="manual/php5.php">
```
cmb69, line 81: I think we should remove that PHP 4 docs duplication right away.

cmb69 (approved):
Thank you!

PR #568, opened 2022-06-28
Refactor compare_version function

kamil-tekiela:
Remove some dead code and add type declarations so it's easier to see what is happening.


cmb69 (approved):
Thank you! Looks like a nice simplification.

I suggest to start bringing parts of the code under test as soon as possible. PHPTs should be okayish for that.

I would also suggest to apply some static analysis, but that brings up the old debate about third-party tools.

PR #570, opened 2022-06-28
Remove duplicated !isset($var)

kamil-tekiela:
There's no need to check the same thing twice. Use either `isset` or `empty`, but not both.


cmb69 (approved):
Thank you! I agree that this is better than before, but it should probably improved in the long run. E.g. `$cm = idate("m")` seems to be more appropriate.

PR #573, opened 2022-06-28
Fix: Remove parameter where argument is never specified

localheinz:
This pull request

- [x] removes a parameter where an argument is never specified


cmb69 (commented):
Thank you! Unfortunately, the PR went stale due to another merge; could you please fix the conflicts?

kamil-tekiela (approved):

localheinz:

PR #576, opened 2022-06-28
Fix: Remove unused local variable

localheinz:
This pull request

- [x] removes an unused local variable `$branches`

### DIFF
--- a/include/gpg-keys.inc
+++ b/include/gpg-keys.inc
@@ -159,8 +159,6 @@ function gpg_key_get_branches(bool $activeOnly): array {
 }
 
 function gpg_key_show_keys(bool $activeOnly): void {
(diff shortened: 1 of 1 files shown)


include/gpg-keys.inc
```diff
@@ -159,8 +159,6 @@ function gpg_key_get_branches(bool $activeOnly): array {
 }
 
 function gpg_key_show_keys(bool $activeOnly): void {
-    $branches = gpg_key_get_branches($activeOnly);
-
     foreach (gpg_key_get_branches($activeOnly) as $branch => $rms) {
```
localheinz, line 162: 🤔

kamil-tekiela (approved):
It's probably safe. Looks like it was a typo during refactoring.

localheinz:

PR #580, opened 2022-06-28
Simplify links

kamil-tekiela:
A function just to print the result of another function already exists in PHP, it's `echo`. So let's remove `print_link` and all unused parameters. Declare types to be string from bottom-up. Warning: it relies on type coercion from false to string just like it did before.

### DIFF
--- a/mirror.php
+++ b/mirror.php
@@ -66,14 +66,14 @@ site_header("Information About This PHP Mirror Site", ["current" => "community"]
 
 <ul>
  <li>This site is <?php echo is_official_mirror() ? "" : "not"; ?> an official PHP.net mirror site</li>
- <li>The mirror site's address is <?php print_link($MYSITE); ?></li>
+ <li>The mirror site's address is <?= make_link($MYSITE)?></li>
 </ul>
 
 <?php if (is_official_mirror()) { ?>
 <h2>Mirror Provider</h2>
 <ul>
  <li>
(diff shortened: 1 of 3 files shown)


localheinz (commented):
👍

cmb69:
> Are the mirror sites still in use?

The mirror program has been retired a few years ago. There still *might* be private mirrors, though.

localheinz:
@cmb69 

Would private mirrors be a problem we need to deal with here?

mirror.php
```diff
@@ -66,14 +66,14 @@
 
 <ul>
  <li>This site is <?php echo is_official_mirror() ? "" : "not"; ?> an official PHP.net mirror site</li>
- <li>The mirror site's address is <?php print_link($MYSITE); ?></li>
+ <li>The mirror site's address is <?php echo make_link($MYSITE); ?></li>
```
cmb69, line 69: What about using short echo tags here (and below):
```suggestion
 <li>The mirror site's address is <?=make_link($MYSITE)?></li>
```

kamil-tekiela:
> > Are the mirror sites still in use?
> 
> The mirror program has been retired a few years ago. There still _might_ be private mirrors, though.

Correct me if I am wrong, but private mirrors don't use https://www.php.net/mirror.php and they won't show up there. This means that we don't really need this functionality anymore, correct? The information about Myra can be moved to a different place on the site.

cmb69 (approved):
Okay, let's leave the mirror stuff there for now.

PR looks good to me. Thank you!

Issue #579, opened 2022-06-28
Should it be possible to download PHP files?


localheinz:
Running

```shell
curl -v https://www.php.net/include/branches.inc
```

allows me to download [`include/branches.inc`](https://github.com/php/web-php/blob/b2ff80507c59fe702ef1f10ad9c255ac9242b07d/include/branches.inc).

derickr:
That check email func file to removed from the repo in 2001:

```
commit 0925042c8942a8de7ef6cf2f822df32c4899ca0e
Author: Jesus M. Castagnetto <jmcastagnetto@php.net>
Date:   Wed Feb 28 12:48:33 2001 +0000

    removed check_email_func.php
```

I'll remove it from the server by hand (I don't think rsync takes care of it).

PR #581, opened 2022-06-28
Fix: Move class into `src/`

localheinz:
This pull request

- [x] moves a previously introduced class into `src/`

💁‍♂️ Given #579, it could make sense to start moving (and extracting additional) classes into `src/`. 

Sooner or later we could even add auto-loading, instead of relying on `require` statements. 

Examples of PSR-4 autoloader implementations can be found [here](https://www.php-fig.org/psr/psr-4/examples/).

### DIFF
--- a/bin/createNewsEntry
+++ b/bin/createNewsEntry
@@ -2,8 +2,9 @@
 <?php
 PHP_SAPI == 'cli' or die("Please run this script using the cli sapi");
 
-require(__DIR__ . '/../include/news_entry.inc');
-use phpweb\news\Entry;
+require_once __DIR__ . '/../src/News/Entry.php';
+
+use phpweb\News\Entry;
 
 $imageRestriction = [
 	'width' => 400,
--- a/bin/createReleaseEntry
+++ b/bin/createReleaseEntry
@@ -2,8 +2,9 @@
 <?php
 PHP_SAPI == 'cli' or die("Please run this script using the cli sapi");
 
-require(__DIR__ . '/../include/news_entry.inc');
-use phpweb\news\Entry;
+require_once __DIR__ . '/../src/News/Entry.php';
+
+use phpweb\News\Entry;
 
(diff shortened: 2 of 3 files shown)


src/news/Entry.php
```diff
@@ -11,7 +11,7 @@ class Entry {
 	];
 
 	const WEBROOT = "https://www.php.net";
-	const PHPWEB = __DIR__ . '/../';
+	const PHPWEB = __DIR__ . '/../../';
```
localheinz, line 14: Run

```shell
bin/createNewsEntry
```

and

```shell
bin/createReleaseEntry
```

to confirm that this still works as expected.

bin/createNewsEntry
```diff
@@ -2,7 +2,7 @@
 <?php
 PHP_SAPI == 'cli' or die("Please run this script using the cli sapi");
 
-require(__DIR__ . '/../include/news_entry.inc');
```
kamil-tekiela, line 5: While you are at it, please remove redundant parentheses. Maybe `news` should be `News`? It would probably be a good idea to conform to PSR-4 and use namespace too.

cmb69: From the [spec](https://www.php-fig.org/psr/psr-4/):

> The subdirectory name MUST match the case of the sub-namespace names.

bin/createNewsEntry
```diff
@@ -2,8 +2,9 @@
...
-use phpweb\news\Entry;
+require_once __DIR__ . '/../src/news/Entry.php';
```
localheinz, line 5: `require_once` should also suffice, right?

cmb69: It's actually better (a second require/include would fail due to class is already defined).

cmb69 (approved):
Thank you!

kamil-tekiela (approved):

localheinz:

PR #583, opened 2022-06-28
Fix: Remove arguments identical to default values of parameters

localheinz:
This pull request

- [x] removes arguments identical to default values of parameters

### DIFF
--- a/include/layout.inc
+++ b/include/layout.inc
@@ -220,9 +220,9 @@ function download_link($file, $title)
     if ($size) {
         echo ' [';
         if ($size < 1024) {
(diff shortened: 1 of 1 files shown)


include/layout.inc
```diff
@@ -220,9 +220,9 @@ function download_link($file, $title)
     if ($size) {
         echo ' [';
         if ($size < 1024) {
-            echo number_format($size, 0, '.', ',') . 'b';
```
localheinz, line 223: See https://www.php.net/manual/en/function.number-format.php:

CleanShot 2022-06-28 at 18 43 42@2x

localheinz:
Thank you, @cmb69!

PR #584, opened 2022-06-28
Fix: Remove useless function

localheinz:
This pull request

- [x] removes a useless function `manual_header()`

### DIFF
--- a/include/shared-manual.inc
+++ b/include/shared-manual.inc
@@ -504,7 +504,6 @@ CHANGE_LANG;
     return trim($r);
 }
 
-function manual_header(){}
 function manual_footer() {
     global $USERNOTES, $__RELATED;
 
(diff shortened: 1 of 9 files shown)


kamil-tekiela (approved):
LGTM I wonder though if it impacts translations.

cmb69:
That would break the manual, since every single manual page calls this function. Note that files under manual/en are just a few examples; the actual files are rendered by PhD, and there we would need to delete this line: https://github.com/php/phd/blob/c3868a838e5029e7cbe39868a1c146e9b81c4986/phpdotnet/phd/Package/PHP/Web.php#L207

After that has been done, and the manual has been successfully rebuilt (happens every 4 or 6 hours), we can merge this PR.

kamil-tekiela:
I thought so. Let's leave it as it is for now.

localheinz:

PR #586, opened 2022-06-28
Fix: Avoid variable function invocation

localheinz:
This pull request

- [x] avoids a variable function invocation

Spotted in #585.


cmb69 (approved):
Ah, good catch! This looks like a hook that has been introduced with acf87cb5bc70ee6007fa6b967757bd96288b1b22, but never used, so I'm fine with removing it. cc @morrisonlevi

morrisonlevi (approved):
I'm not sure what hook you are referring, @cmb69 , but this change looks fine to me.

localheinz:
Thank you, @cmb69 and @morrisonlevi!

PR #587, opened 2022-06-28
Fix: Add type and return type declaration

localheinz:
This pull request

- [x] adds a type and return type declaration to `version_number_to_branch()`

💁‍♂️ See

- http://localhost:8080/eol.php
- http://localhost:8080/index.php
- http://localhost:8080/releases/active.php
- http://localhost:8080/releases/index.php
- http://localhost:8080/supported-versions.php

### DIFF
--- a/include/branches.inc
+++ b/include/branches.inc
@@ -86,11 +86,13 @@ function format_interval($from, $to) {
 	return $eolPeriod;
 }
 
-function version_number_to_branch($version) {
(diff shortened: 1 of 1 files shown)


kamil-tekiela (approved):
Seems ok. There's some dubious type juggling in places where this is called, but that's irrelevant.

localheinz:
Thank you, @kamil-tekiela!

PR #591, opened 2022-06-29
Use accent color for hero buttons :hover and :focus

morrisonlevi:
I've always wanted the accent to be used more, but it's kind of hard.
I think this is a great opportunity for it!

Here's a screenshot of it with one of the buttons being hovered:

PHP.net homepage with hero button in a purple-ish accent color


morrisonlevi:
Definitely, but let's not burden this PR with it, I think?

PR #593, opened 2022-06-29
Move ManualNotesSorter to src/

kamil-tekiela:
Move class into its own file in `src/` dir.

### DIFF
--- a/src/Notes/Sorter.php
+++ b/src/Notes/Sorter.php
@@ -1,0 +1,112 @@
+<?php
+
+namespace phpweb\Notes;
+
+class Sorter {
+    private $maxVote;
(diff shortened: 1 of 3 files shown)


src/Notes/Sorter.php
```diff
@@ -0,0 +1,112 @@
+<?php
+
+namespace phpweb\Notes;
```
cmb69, line 3: Maybe call it `phpweb\UserNotes` to emphasis that it is about the user notes, not `<note>`s in the manual?

cmb69 (approved):
Thank you! Nice one.

It seems that the class has no dependencies, so adding test shouldn't be that hard. :)

PR #594, opened 2022-06-29
Introduce UserNote DTO

kamil-tekiela:
It would be great if someone could retest it before approving. I tested it myself and saw no regressions. 

This PR solves a number of code issues. Mostly it removes nonsense in the Sorter class.

### DIFF
--- a/include/shared-manual.inc
+++ b/include/shared-manual.inc
@@ -125,7 +125,7 @@ function manual_note_display(UserNote $note, $voteOption = true)
     $date = new DateTime("@{$note->ts}");
     $datestr = relTime($date);
     $fdatestr = $date->format("Y-m-d h:i");
-    $text    = clean_note($note->text);
+    $text = clean_note($note->text);
 
     // Calculate note rating by up/down votes
     $vote = $note->upvotes - $note->downvotes;
--- a/include/shared-manual.inc
+++ b/include/shared-manual.inc
@@ -30,7 +30,7 @@ use phpweb\UserNotes\UserNote;
 /**
  * Print out all user notes for this manual page
  *
- * @param UserNote[] $notes
+ * @param array<string, UserNote> $notes
  */
 function manual_notes($notes):void {
     // Get needed values
@@ -85,7 +85,7 @@ END_USERNOTE_HEADER;
 /**
  * Get user notes from the appropriate text dump
  *
- * @return UserNote[]
+ * @return array<string, UserNote>
  */
 function manual_notes_load(string $id): array
 {
--- a/include/shared-manual.inc
+++ b/include/shared-manual.inc
@@ -23,10 +23,16 @@ $PGI = []; $SIDEBAR_DATA = '';
 // =============================================================================
 
 require_once __DIR__ . '/../autoload.php';
-use phpweb\UserNotes\Sorter;
 
-// Print out all user notes for this manual page
-function manual_notes($notes) {
+use phpweb\UserNotes\Sorter;
+use phpweb\UserNotes\UserNote;
+
+/**
+ * Print out all user notes for this manual page
+ *
+ * @param UserNote[] $notes
+ */
+function manual_notes($notes):void {
     // Get needed values
     list($filename) = $GLOBALS['PGI']['this'];
 
@@ -68,20 +74,21 @@ END_USERNOTE_HEADER;
         // If we have notes, print them out
         echo '<div id="allnotes">';
         foreach ($notes as $note) {
-            manual_note_display(
-                $note['xwhen'], $note['user'], $note['note'], $note['id'], $note['votes']
-            );
+            manual_note_display($note);
         }
         echo "</div>\n";
         echo "<div class=\"foot\">$addnotesnippet</div>\n";
     }
     echo "</section>";
 }
-// Get user notes from the appropriate text dump
-function manual_notes_load($id)
+
+/**
+ * Get user notes from the appropriate text dump
+ *
+ * @return UserNote[]
+ */
+function manual_notes_load(string $id): array
 {
-    // Initialize values
-    $notes = [];
     $hash = substr(md5($id), 0, 16);
     $notes_file = $_SERVER['DOCUMENT_ROOT'] . "/backend/notes/" .
                   substr($hash, 0, 2) . "/$hash";
@@ -89,22 +96,15 @@ function manual_notes_load($id)
     // Open the note file for reading and get the data (12KB)
     // ..if it exists
     if (!file_exists($notes_file)) {
-        return $notes;
+        return [];
     }
+    $notes = [];
     if ($fp = @fopen($notes_file, "r")) {
         while (!feof($fp)) {
             $line = chop(fgets($fp, 12288));
             if ($line == "") { continue; }
             @list($id, $sect, $rate, $ts, $user, $note, $up, $down) = explode("|", $line);
-            $notes[$id] = [
-                "id" => $id,
-                "sect" => $sect,
-                "rate" => $rate,
-                "xwhen" => $ts,
-                "user" => $user,
-                "note" => base64_decode($note, true),
-                "votes" => ["up" => (int)$up, "down" => (int)$down]
-            ];
+            $notes[$id] = new UserNote($id, $sect, $rate, $ts, $user, base64_decode($note, true), (int) $up, (int) $down);
         }
         fclose($fp);
     }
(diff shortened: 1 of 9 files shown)


include/shared-manual.inc
```diff
@@ -69,68 +76,63 @@ END_USERNOTE_HEADER;
...
     }
     echo "</section>";
 }
-// Get user notes from the appropriate text dump
-function manual_notes_load($id)
+
+/**
+ * Get user notes from the appropriate text dump
+ *
+ * @return UserNote[]
```
localheinz, line 90:
```suggestion
 * @return array<string, UserNote>
```

include/shared-manual.inc
```diff
@@ -69,68 +76,63 @@ END_USERNOTE_HEADER;
...
-            $notes[$id] = array(
-                "id"    => $id,
-                "sect"  => $sect,
-                "rate"  => $rate,
-                "xwhen" => $ts,
-                "user"  => $user,
-                "note"  => base64_decode($note),
-                "votes"  => array("up"=> (int)$up, "down"=> (int)$down)
-            );
+            $notes[$id] = new UserNote($id, $sect, $rate, $ts, $user, base64_decode($note), (int) $up, (int) $down);
```
localheinz, line 109:
```suggestion
            $notes[$id] = new UserNote(
                $id,
                $sect,
                $rate,
                $ts,
                $user,
                base64_decode($note),
                (int) $up,
                (int) $down
            );
```

src/UserNotes/UserNote.php
```diff
@@ -0,0 +1,44 @@
...
+        string $id,
+        string $sect,
+        string $rate,
+        string $ts,
+        string $user,
+        string $text,
+        int $upvotes = 0,
+        int $downvotes = 0
+    )
+    {
```
localheinz, line 34:
```suggestion
    ) {
```

src/UserNotes/UserNote.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace phpweb\UserNotes;
+
+class UserNote
```
localheinz, line 5:
```suggestion
final class UserNote
```

src/UserNotes/UserNote.php
```diff
@@ -0,0 +1,44 @@
...
+    /** @var string $ts */
+    public $ts;
+    /** @var string $user */
+    public $user;
+    /** @var string $text */
+    public $text;
+    /** @var int $upvotes */
+    public $upvotes;
+    /** @var int $downvotes */
+    public $downvotes;
```
localheinz, line 22: How do you feel about marking these as `private` and adding accessors instead?

```suggestion
-    /** @var string $id */
-    public $id;
-    /** @var string $sect */
-    public $sect;
-    /** @var string $rate */
-    public $rate;
-    /** @var string $ts */
-    public $ts;
-    /** @var string $user */
-    public $user;
-    /** @var string $text */
-    public $text;
-    /** @var int $upvotes */
-    public $upvotes;
-    /** @var int $downvotes */
-    public $downvotes;
+    private $id;
+    private $sect;
+    private $rate;
+    private $ts;
+    private $user;
+    private $text;
+    private $upvotes;
+    private $downvotes;
```

src/UserNotes/UserNote.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        $this->id = $id;
+        $this->sect = $sect;
+        $this->rate = $rate;
+        $this->ts = $ts;
+        $this->user = $user;
+        $this->text = $text;
+        $this->upvotes = $upvotes;
+        $this->downvotes = $downvotes;
+    }
```
localheinz, line 43:
```suggestion
    }
    
    public function id(): string
    {
        return $this->id;
    }

    public function sect(): string
    {
        return $this->sect;
    }

    public function rate(): string
    {
        return $this->rate;
    }

    public function ts(): string
    {
        return $this->ts;
    }

    public function user(): string
    {
        return $this->user;
    }

    public function text(): string
    {
        return $this->text;
    }

    public function upvotes(): int
    {
        return $this->upvotes;
    }

    public function downvotes(): int
    {
        return $this->downvotes;
    }    
```

src/UserNotes/UserNote.php
localheinz, line 22: How do you feel about marking these as `private` and adding accessors instead?

```suggestion
-    /** @var string $id */
-    public $id;
-    /** @var string $sect */
-    public $sect;
-    /** @var string $rate */
-    public $rate;
-    /** @var string $ts */
-    public $ts;
-    /** @var string $user */
-    public $user;
-    /** @var string $text */
-    public $text;
-    /** @var int $upvotes */
-    public $upvotes;
-    /** @var int $downvotes */
-    public $downvotes;
+    private $id;
+    private $sect;
+    private $rate;
+    private $ts;
+    private $user;
+    private $text;
+    private $upvotes;
+    private $downvotes;
```

manual/add-note.php
```diff
@@ -7,6 +7,8 @@
 include_once __DIR__ . '/../include/shared-manual.inc';
 include      __DIR__ . '/spam_challenge.php';
 
+use phpweb\UserNotes\UserNote;
```
localheinz, line 10: How do you feel about importing the relevant parent namespace only?

See https://localheinz.com/blog/2020/05/19/avoiding-imports-and-aliases/.

src/UserNotes/Sorter.php
```diff
@@ -15,6 +15,9 @@ class Sorter {
     private $ratingWeight = 60;
     private $ageWeight = 2;
 
+    /**
+     * @param UserNote[] $notes
```
localheinz, line 19:
```suggestion
     * @param array<string, UserNote> $notes
```

src/UserNotes/Sorter.php
```diff
@@ -30,83 +33,58 @@ function sort(array &$notes) {
...
+        $totalVotes = $note->upvotes + $note->downvotes;
+        return $totalVotes > 0 ? $note->upvotes / $totalVotes : .5;
     }
 
-
-    /*
-     * Not sure why, but using `$b['sort'] - $a['sort']` does not seem to
-     * work properly.
+    /**
+     * @param UserNote[] $notes
```
localheinz, line 58:
```suggestion
     * @param array<string, UserNote> $notes
```

src/UserNotes/Sorter.php
```diff
@@ -30,83 +33,58 @@ function sort(array &$notes) {
...
         }
+        return $prio;
     }
 
-
-    private function findMinMaxValues(array &$notes) {
-        $count = count($notes);
-        if ($count <= 0) {
+    /**
+     * @param UserNote[] $notes
```
localheinz, line 71:
```suggestion
     * @param array<string, UserNote> $notes
```

src/UserNotes/UserNote.php
cmb69: I'd rather go with `readonly` (as annotation for now), if the properties are actually readonly.

include/shared-manual.inc
kamil-tekiela: I am not sure if it improves readability. The line is not that long and monitors are usually wider than taller.

localheinz: See https://localheinz.com/blog/2021/12/31/wrapping-it-up/!

src/UserNotes/UserNote.php
localheinz: @cmb69 

I think they are mutated currently.

kamil-tekiela: I experimented with few different approaches. In the end, I decided that it doesn't have to be immutable. With PHP 8.1, we might easily convert this into readonly and use constructor property promotion. It would make migration easier if we use public properties now. I'd keep it as it is now, optionally adding `readonly` annotation as cmb69 suggests.

manual/add-note.php
kamil-tekiela: I'm not sure what we would gain by doing this. Can you explain? The article you mentioned, talks about code that imports many things, but in our case, it's only two classes. IDEs show the fully qualified name when you hover over it, so I don't think it's ambiguous where the class comes from.

src/UserNotes/UserNote.php
kamil-tekiela: Thinking about it more, I decided to mark the class as `immutable` and `readonly`. I don't think any static analyser respects `readonly` annotation on class, but if we migrate to PHP 8.2, we will know to make the class as readonly.

include/shared-manual.inc
kamil-tekiela: Cool article and I agree with most of what it says. However, in this case, I have to disagree. I see zero advantage for having it in multiple lines and I think it mostly comes down to a matter of choice. Here's a PHPStorm screenshot:
image

As you can see the line doesn't stand out in length and isn't unreadable. It visually matches the line before it. 

@cmb69 What is your opinion? Which style do you like more?

cmb69: I have no strong feelings here. Maybe leave as is, and reconsider, if we adopt some coding standard.

localheinz (commented):
👍

cmb69:
Would it make sense to bring the `UserNotes\Sorter` under test before doing this? I mean we would need to modify the test as well, but still could have more confidence to not inadvertently break anything.

Maybe something like the following PHPT:
````php
--TEST--

--FILE--
<?php
require_once __DIR__ . "/../src/UserNotes/Sorter.php";

$notes = [
    [
        "votes" => ["up" => 2, "down" => 0],
        "xwhen" => 1613487094,
    ], [
        "votes" => ["up" => 0, "down" => 0],
        "xwhen" => 1508180150,
    ], [
        "votes" => ["up" => 14, "down" => 3],
        "xwhen" => 1508179844,
    ],
];

$sorter = new phpweb\UserNotes\Sorter();
$sorter->sort($notes);
var_dump($notes);
?>
--EXPECT--
array(3) {
  [2]=>
  array(6) {
    ["votes"]=>
    array(2) {
      ["up"]=>
      int(14)
      ["down"]=>
      int(3)
    }
    ["xwhen"]=>
    int(1508179844)
    ["score"]=>
    int(11)
    ["total"]=>
    int(17)
    ["rating"]=>
    float(0.82352941176471)
    ["sort"]=>
    float(87.411764705882)
  }
  [0]=>
  array(6) {
    ["votes"]=>
    array(2) {
      ["up"]=>
      int(2)
      ["down"]=>
      int(0)
    }
    ["xwhen"]=>
    int(1613487094)
    ["score"]=>
    int(2)
    ["total"]=>
    int(2)
    ["rating"]=>
    int(1)
    ["sort"]=>
    float(48.236363636364)
  }
  [1]=>
  array(6) {
    ["votes"]=>
    array(2) {
      ["up"]=>
      int(0)
      ["down"]=>
      int(0)
    }
    ["xwhen"]=>
    int(1508180150)
    ["score"]=>
    int(0)
    ["total"]=>
    int(0)
    ["rating"]=>
    float(0.5)
    ["sort"]=>
    float(41.400005811566)
  }
}
````

kamil-tekiela:
Yes, it would make sense. The problem with this is that I couldn't understand what the current implementation does, which is why I rewrote and simplified it. So while a test would be nice to prevent regression, I don't actually know what would make a good test to make sure we catch the edge cases. If you are ok with a simple test like the one proposed above, I am ok to wait for a test before merging this PR. I have checked the note order before and after myself, but I still can't guarantee that all edge cases are covered.

localheinz:
@cmb69 @kamil-tekiela 

In regard to tests, see #605!

cmb69:
See also #606.

kamil-tekiela:
@cmb69  Do we need to adjust the tests more?

localheinz:
@kamil-tekiela 

Can you rebase on top of `master`?

cmb69:
> Do we need to adjust the tests more?

If we can't show the internals (especially `score`, `rating` and `sort`), we should have more tests. I've written the tests assuming that we still can verify these internals, and so didn't look for testing the final sort order thoroughly. And even if the current sort order might not be the best, I think we should stick with it when refactoring.

kamil-tekiela:
@cmb69 I added a regression test which takes the sample list of 46 notes and sorts them. The expected output is the same as is available on master, so if the expected sort and new sort is the same, it's safe to assume the logic hasn't been broken. There's still unlikely chance that some edge case was missed, but I have no idea how to test it more thoroughly than this.

cmb69 (approved):
This looks good to me. Thank you!

PR #595, opened 2022-06-30
Remove security section

kamil-tekiela:
It contains PHP 5 info and the [never-finished security center](https://github.com/php/web-php/commit/04044b071f895f92ec257a85b1333b3b5f462d5f).

I added a redirect which I hope is working correctly.


cmb69 (approved):
Wow, very good catch!

We need to cater to [`url.crypt.blowfish`](https://github.com/php/doc-base/blob/f0bfa37b6ae4fac314a1876fda14c89e19eeb5e5/entities/global.ent#L51) though. This is currently used only on the [`crypt()` manpage](https://www.php.net/crypt), and I think we can remove that sentence there. After translations caught up, we could remove the entity altogether.

localheinz (commented):
👍

kamil-tekiela:
@cmb69 Is it safe to merge this too?

PR #597, opened 2022-07-01
Announce SymfonyWorld Online 2022 Winter Edition conference

javiereguiluz:
Similar to #596, and for your consideration, this announces the last Symfony/PHP conference of 2022.

Thank you.


archive/entries/2022-07-01-1.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
+<entry xmlns="http://www.w3.org/2005/Atom" xmlns:default="http://php.net/ns/news">
+  <title>SymfonyWorld Online 2022 Winter Edition</title>
+  <id>https://www.php.net/archive/2022.php#2022-07-01-1</id>
+  <published>2022-07-01T17:33:43+00:00</published>
+  <updated>2022-07-01T17:33:43+00:00</updated>
+  <link href="https://www.php.net/conferences/index.php#id2022-07-01-1" rel="alternate" type="text/html"/>
+  <link href="https://live.symfony.com/2022-world-winter/" rel="via" type="text/html"/>
+  <default:finalTeaserDate xmlns="http://php.net/ns/news">2022-12-08</default:finalTeaserDate>
```
cmb69, line 9: I gave you wrong information last time (but fixed it before applying); the date is supposed to be the first day for conference announcements, and the last day for call for papers. So in this case:
```suggestion
  <default:finalTeaserDate xmlns="http://php.net/ns/news">2022-12-06</default:finalTeaserDate>
```

cmb69 (commented):
Thank you!

archive/entries/2022-07-01-1.xml
javiereguiluz: Thanks for checking this again. I've updated the date as you suggested. Thanks a lot!

cmb69:
Thank you!

PR #599, opened 2022-07-01
Replace `rand()` calls with `random_int()`

Ayesh:
Switching to a better PRNG now that we no longer need to support PHP 5.

### DIFF
--- a/manual/spam_challenge.php
+++ b/manual/spam_challenge.php
@@ -8,7 +8,7 @@ function plus($a, $b) {
 }
 
 function gen_plus($a) {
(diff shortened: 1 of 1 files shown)


localheinz:
Also see #533.

Ayesh:
Awesome, thanks for pointing to it, I will close this in favor of it because #533 covers the changes I was proposing too.

cmb69:
Hmm, I think a CSPRNG is preferable here, although there is more room for improvement regarding the CAPTCHA generation.

PR #600, opened 2022-07-01
`phpversion()` replacements and code branch collapsing for older PHP versions

Ayesh:
- Removes a `if` branch was PHP < 5, which should be safe now that we no longer support PHP 4. 
 - Replaces `phpversion()` function call with `PHP_VERSION` constant, both of which yield identical results, but the use of the constant is a micro-optimization.

### DIFF
--- a/include/ip-to-country.inc
+++ b/include/ip-to-country.inc
@@ -209,18 +209,9 @@ function i2c_realip()
             // Skip RFC 1918 IP's 10.0.0.0/8, 172.16.0.0/12 and
             // 192.168.0.0/16
             // Also skip RFC 6598 IP's
-            if (!preg_match('/^(?:10|100\.(?:6[4-9]|[7-9]\d|1[01]\d|12[0-7])|172\.(?:1[6-9]|2\d|3[01])|192\.168)\./', $ips[$i])) {
-                if (version_compare(phpversion(), "5.0.0", ">=")) {
-                    if (ip2long($ips[$i]) != false) {
-                        $ip = $ips[$i];
-                        break;
-                    }
-                } else {
-                    if (ip2long($ips[$i]) != -1) {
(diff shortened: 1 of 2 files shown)


mirror-info.php
```diff
@@ -33,7 +33,7 @@ function_exists('hash_file') &&
 
 echo implode('|', array(
     $MYSITE,            	// 0 : CNAME for mirror as accessed (CC, CC1, etc.)
-    phpversion(),       	// 1 : PHP version overview
+    PHP_VERSION,       	// 1 : PHP version overview
```
cmb69, line 36: The indentation of the comment doesn't look quite right, but it's broken anyway, so leave it as is.

include/ip-to-country.inc
```diff
@@ -210,16 +210,9 @@ function i2c_realip()
...
-                    if (ip2long($ips[$i]) != false) {
-                        $ip = $ips[$i];
-                        break;
-                    }
-                } else {
-                    if (ip2long($ips[$i]) != -1) {
-                        $ip = $ips[$i];
-                        break;
-                    }
+                if (ip2long($ips[$i]) != false) {
```
cmb69, line 213: Please cater to the misleading `!= false`; either drop it altogether, or do an `!== false`.

cmb69 (commented):
Thank you for the PR! Makes perfect sense, but the `!= false` should better be addressed right away.

include/ip-to-country.inc
Ayesh: Good call, thank you. I dropped the `!= false` clause, and combined the `if` block with the parent.

cmb69: Ah, even better (the line is a bit long, but we can cater to that when we introduce some coding standard).

PR #603, opened 2022-07-02
Multiple micro-optimizations

Ayesh:
This PR contains multiple commits with micro-optimizations, hopefully to reduce the review burden reviewing multiple fully fledged PRs.
If any of them don't seem to make sense, I'll happily take them off. 

Thank you.

---

- Replace `ob_get_contents();ob_clean()` with `ob_get_clean()`
    `ob_get_clean()` is equivalent to `ob_get_contents()` followed by `ob_clean()`.

 - Replace `intval()` calls with `(int)` type cast
    This is a micro-optimization because `intval()` is a function call, and the type cast is about 6 times fast.

 - Replace `preg_replace` call that could be done with an `rtrim()` call
    In `./error.php`, there is a `preg_replace('!/+$!', '', $URI);` call that essentially is equivalent to `rtrim()`, that both calls removing trailing slash characters in `$URI`.
    The `rtim()` call is more legible and faster.

 - Combine consecutive `str_replace` calls to a single `str_replace` call
 - Use short ternary operator where possible
 - Cascade various `else` statements where possible
     Cleans up the code by removing unnecessary `else` blocks and moving the code to the parent context if the previous `if` block exits the function by either terminating the script, or with a `return` statement.

 - Combine multiple `isset()` calls to a single `isset()`
    `isset()` accepts multiple parameters and returns `true` only if all of the parameters are `isset`. It makes sense to combine multiple individual `isset` calls to a single call for better readability.

 - Replace `for` loop with a `foreach` loop
 -  Remove unnecessary character escapes in regular expressions
     Regular expression special characters are context-sensitive. For example, special characters such as `.` are not considered special within square braces (`[]`).
    This removes several of such instances that certain characters are escaped, but it is not strictly necessary within the context. This improves the readability of the expression.

    A write-up is at [PHP.Watch: Writing better Regular Expressions in PHP](https://php.watch/articles/php-regex-readability#reduce-escape)

 - Remove unnecessary break statement
 - Remove unnecessary PHP close tags

### DIFF
--- a/error.php
+++ b/error.php
@@ -700,4 +700,3 @@ mirror_redirect(
 /*
  * vim: set et ts=4 sw=4 ft=php: :
  */
-?>
--- a/error.php
+++ b/error.php
@@ -116,7 +116,7 @@ if (preg_match("!^distributions/.*!", $URI, $array)) {
 
 // ============================================================================
 // The trailing slash only causes problems from now on
-$URI = preg_replace('!/+$!', '', $URI);
+$URI = rtrim($URI, '/');
 
 // ============================================================================
 // Some nice URLs for getting something for download
--- a/include/langchooser.inc
+++ b/include/langchooser.inc
@@ -102,11 +102,11 @@ function language_choose_code()
         $browser_accept = explode(",", $_SERVER['HTTP_ACCEPT_LANGUAGE']);
 
         // Go through all language preference specs
-        for ($i = 0; $i < count($browser_accept); $i++) {
+        foreach ($browser_accept as $value) {
             // The language part is either a code or a code with a quality
             // We cannot do anything with a * code, so it is skipped
             // If the quality is missing, it is assumed to be 1 according to the RFC
-            if (preg_match("!([a-z-]+)(;q=([0-9\\.]+))?!", trim($browser_accept[$i]), $found)) {
+            if (preg_match("!([a-z-]+)(;q=([0-9\\.]+))?!", trim($value), $found)) {
                 $quality = (isset($found[3]) ? (float) $found[3] : 1.0);
                 $browser_langs[] = array($found[1], $quality);
             }
--- a/manual/vote-note.php
+++ b/manual/vote-note.php
@@ -39,7 +39,7 @@ if ($_SERVER['REQUEST_METHOD'] == 'POST') {
       }
       else {
         $r = json_decode($r);
-        if (json_last_error() == JSON_ERROR_NONE && isset($r->status) && $r->status && isset($r->votes)) {
+        if (isset($r->status, $r->votes) && json_last_error() == JSON_ERROR_NONE && $r->status) {
           $response["success"] = true;
           $response["update"] = (int)$r->votes;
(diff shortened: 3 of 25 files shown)


manual/vote-note.php
```diff
@@ -39,7 +39,7 @@
       }
       else {
         $r = json_decode($r);
-        if (json_last_error() == JSON_ERROR_NONE && isset($r->status) && $r->status && isset($r->votes)) {
+        if (isset($r->status, $r->votes) && json_last_error() == JSON_ERROR_NONE && $r->status) {
```
kamil-tekiela, line 42: That's ... not right. Perhaps remove `json_last_error() == JSON_ERROR_NONE`? `isset((null)->status)` works fine, so checking for JSON error is redundant.  

Ditto for other places in this file.

error.php
```diff
@@ -93,7 +93,7 @@
 
 // ============================================================================
 // BC: Release files moved to subfolder
-if (preg_match("!^release_([^\\.]+)(\\.php$|$)!", $URI, $array)) {
+if (preg_match("!^release_([^.]+)(\\.php$|$)!", $URI, $array)) {
```
kamil-tekiela, line 96: I think this is wrong. A single `\` should still be present in the character class.

include/langchooser.inc
```diff
@@ -106,7 +106,7 @@ function language_choose_code()
             // The language part is either a code or a code with a quality
             // We cannot do anything with a * code, so it is skipped
             // If the quality is missing, it is assumed to be 1 according to the RFC
-            if (preg_match("!([a-z-]+)(;q=([0-9\\.]+))?!", trim($value), $found)) {
+            if (preg_match("!([a-z-]+)(;q=([0-9.]+))?!", trim($value), $found)) {
```
kamil-tekiela, line 109: Ditto

kamil-tekiela (changes requested):

manual/vote-note.php
Ayesh: Thank you. You are right the JSON error check isn't really necessary. Removed.

error.php
Ayesh: This was indeed wrong. Reverted this change (to use `\\`).

include/langchooser.inc
Ayesh: This too.

Ayesh:
Thanks a lot for the review @kamil-tekiela. I made the regex changes as you mentioned (by restoring the original version), and dropped the redundant `JSON_ERROR_NONE` checks too.

Additionally, there is a new change in `!&lt;\?php(<br />)+!`, that I dropped the character escape in `<br />` by changing the delimiter from `/` to `!`, so we don't have to escape `/`.

kamil-tekiela (approved):
LGTM

error.php
cmb69: The double back-slash won't hurt (nor would a single backslash; the regex engine sees `\.` either way), but it is superfluous. From the [docs](https://www.php.net/manual/en/regexp.reference.character-classes.php):

> All non-alphanumeric characters other than \\, -, ^ (at the start) and the terminating ] are non-special in character classes, but it does no harm if they are escaped.

cmb69:
Thank you!

PR #604, opened 2022-07-03
Enhancement: Use PSR-4 autoloader

localheinz:
This pull request

- [x] uses a PSR-4 autoloader

Follows #581.

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -1,0 +1,28 @@
+<?php
+
+/**
+ * @see https://github.com/php-fig/fig-standards/blob/a1a0674a742c9d07c5dd450209fb33b115ee7b40/accepted/PSR-4-autoloader-examples.md#closure-example
+ */
+spl_autoload_register(static function (string $class): void {
+    $prefix = 'phpweb\\';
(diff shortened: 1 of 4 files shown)


kamil-tekiela:
We could also use the PHP native autoloader, but we would have to get rid of the phpweb namespace as the standard one doesn't have a way to map `phpweb` to `src` as far as I know. 

I think the `News` class should also use autoloader.

cmb69 (approved):
Makes perfect sense to me. Thank you!

kamil-tekiela:
Something like this would also work:

```
set_include_path(__DIR__.'/src/'.get_include_path());

spl_autoload_register(static function (string $class): void {
    spl_autoload(str_replace("phpweb\\", "",$class));
});
```

kamil-tekiela (approved):

localheinz:
@kamil-tekiela

I would not bother further optimizing the autoloading situation, especially in the light of #408.

localheinz:
Thank you, @kamil-tekiela!

PR #605, opened 2022-07-03
Enhancement: Use `phpunit/phpunit` to run tests

localheinz:
This pull request

- [x] uses `phpunit/phpunit` to run tests

💁‍♂️ Similar to #559, you can see this in action in https://github.com/localheinz/web-php/pull/2.


cmb69:
I'm not sure whether we should use PHPUnit; that's yet another third-party dependency, and it seems that PHPTs may be sufficient (at least for a while). The nice thing is that these also *can* be run with PHPUnit, what is especially interesting wrt. code coverage reporting.

localheinz:
@cmb69 

`phpunit/phpunit` is the de facto industry standard for testing PHP code. I believe it is time to start using select dependencies.

localheinz:
Closing in favour of #606 and https://github.com/cmb69/web-php/pull/1.

localheinz:
@cmb69 @kamil-tekiela 

I have reopened this pull request because I believe we are at a point where we would benefit from using `phpunit/phpunit`.

What do you think?

TomasVotruba:
Thank you @localheinz for great improvement :clap:

PR #607, opened 2022-07-03
Fix: Remove unused methods

localheinz:
This pull request

- [x] removes unused methods from `phpweb\News\Entry`

### DIFF
--- a/src/News/Entry.php
+++ b/src/News/Entry.php
@@ -21,9 +21,6 @@ class Entry {
 
 
 	protected $title = '';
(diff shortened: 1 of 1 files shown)


cmb69 (commented):
I agree that we shouldn't expose unused getters, but maybe @sgolemon had a good reason to do so.

cmb69:
Well, let's remove them; we can still bring them back if necessary.

Thanks for the PR!

localheinz:
Thank you, @cmb69!

PR #608, opened 2022-07-03
Micro Optimizations - Round #2

Ayesh:
Follow-up to #603, includes a few more optimizations. 
 - [.gitignore: Remove redundant entry](https://github.com/php/web-php/commit/9c77a3cf1293651356ef1e68233d2ace3d2b76d2)
 - [Minor optimizations with ternary operators](https://github.com/php/web-php/commit/24406af9c96edac1fd48cb7cdbd933cf7214a1bd)
 - [Use const instead of define() where appropriate](https://github.com/php/web-php/commit/4b337aea819119792e4e5c889c2a0614620a693e)
    `const` is quite faster because of the compile-time optimizations. Because the replaced statements are not declaring constant conditionally, it's safe to use `const` in all of these places.

Thank you.

### DIFF
--- a/images/logo.php
+++ b/images/logo.php
@@ -1,6 +1,6 @@
 <?php
 
-$refresh = isset($_GET['refresh']) ? true : false;
+$refresh = isset($_GET['refresh']);
 
 // Be 100% sure the timezone is set
 if (ini_get('date.timezone') === '' && function_exists('date_default_timezone_set')) {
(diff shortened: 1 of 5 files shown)


images/logo.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-$refresh = isset($_GET['refresh']) ? true : false;
+$refresh = isset($_GET['refresh']);
```
cmb69, line 3: Thank you for removing this nonsense! :)

kamil-tekiela (approved):

cmb69 (approved):
Thank you!

PR #612, opened 2022-07-05
Minor optimizations in CSS to shed some page size off

Ayesh:
- Removes `px` suffix from `0px`.
 - Combines CSS properties where they are immediately overridden in the same selector block, or can be combined without side effects.
This saves about 50 bytes off the CSS scripts. _it ain't much but it's honest work_

### DIFF
--- a/styles/theme-base.css
+++ b/styles/theme-base.css
@@ -555,8 +555,7 @@ div.classsynopsisinfo_comment {
 
 pre.info {
     border: 1px solid;
-    margin: 1rem 2rem 1.3rem;
-    margin-right: 0.8rem;
+    margin: 1rem 0.8rem 1.3rem 2rem;
 }
 
 #langform {
@@ -686,7 +685,6 @@ table td {
 
 table.standard {
     border-collapse: collapse;
-    border-style: hidden;
     border:1px solid #d9d9d9;
 }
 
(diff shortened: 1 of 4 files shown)


styles/theme-base.css
```diff
@@ -555,8 +555,7 @@ div.classsynopsisinfo_comment {
 
 pre.info {
     border: 1px solid;
-    margin: 1rem 2rem 1.3rem;
-    margin-right: 0.8rem;
+    margin: 1rem 0.8rem 2rem 1.3rem;
```
kamil-tekiela, line 558:
```suggestion
    margin: 1rem 0.8rem 1.3rem 2rem;
```
You swapped bottom with left

cmb69, line 558: Hmm, shouldn't this be:
```suggestion
    margin: 1rem 0.8rem 1.3rm 2rem;
```

cmb69 (commented):
Thank you! Looks overall good to me, but I think the margin isn't quite right.

styles/theme-base.css
Ayesh: Thanks for catching this embarrassing mistake. Fixed now.

Ayesh: You are right this indeed was the one. Thanks for catching this embarrassing mistake. Force-pushed with changes.

cmb69 (approved):

kamil-tekiela (approved):

PR #606, opened 2022-07-03
Test UserNotes\Sorter

cmb69:
We add a copy of the latest run-tests.php from the PHP-8.0 branch, and
tests for the `UserNotes\Sorter` with full code coverage.

PR #614, opened 2022-07-05
Enhancement: Use autoloader

localheinz:
This pull request

- [x] uses the autoloader in tests

Follows #604.
Follows #606.

### DIFF
--- a/tests/sort_notes_001.phpt
+++ b/tests/sort_notes_001.phpt
@@ -4,7 +4,7 @@ sort some notes
 precision=-1
 --FILE--
 <?php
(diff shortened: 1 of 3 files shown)


localheinz:
@kamil-tekiela 

> This is fine, but the tests should work even without autoloader which is why I only made this change as part of my PR.

But why not use the autoloader in tests?

kamil-tekiela:
I only did it as part of the other PR to avoid merge conflicts, but since you proposed it here, let's merge it.

localheinz:
Thank you, @kamil-tekiela!

PR #618, opened 2022-07-06
Remove id prefix

kamil-tekiela:
We stopped adding the prefix in July 2019, but the links are still pointing to anchors with the prefix.


localheinz (commented):
:+1:

cmb69 (approved):
D'oh! Thank you!

PR #619, opened 2022-07-08
Changelog: do not confuse GH-xxx with #xxx bugs

guilliamxavier:
E.g. in https://www.php.net/ChangeLog-8.php#8.1.6, excerpt generated from https://github.com/php/php-src/blob/php-8.1.8/NEWS#L140-L147 (i.e. https://github.com/php/web-php/blob/22ea1e2e6f6569e040ad92445c7a20749a2b0bd0/ChangeLog-8.php#L159-L165):
- before:
  image
- after:
  image

### DIFF
--- a/include/changelogs.inc
+++ b/include/changelogs.inc
@@ -24,7 +24,7 @@ function githubissue($repo, $number) {
 }
 
 function githubissuel($repo, $number) {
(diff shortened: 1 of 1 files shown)


cmb69:
Thank you! This makes perfect sense.

PR #554, opened 2022-06-17
Enhancement: Add `.editorconfig`

localheinz:
This pull request

- [x] adds a basic `.editorconfig`

Follows https://github.com/php/web-php/pull/541#issuecomment-1158935149.

PR #620, opened 2022-07-08
Enhancement: Set charset for all files to UTF-8

localheinz:
This pull request

- [x] sets the charset for all files to `UTF-8`

Follows #554.

💁‍♂️ I would not mind pulling up more attributes!

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,16 +1,15 @@
 root = true
 
 [*]
+charset = utf-8
 insert_final_newline = true
 
 [*.yaml]
-charset = utf-8
 indent_size = 2
 indent_style = space
 trim_trailing_whitespace = true
 
 [Makefile]
-charset = utf-8
 indent_size = 4
 indent_style = tab
 trim_trailing_whitespace = true


.editorconfig
```diff
@@ -1,10 +1,10 @@
 root = true
 
 [*]
+charset = utf-8
 insert_final_newline = true
 
 [*.yaml]
-charset = utf-8
 indent_size = 2
 indent_style = space
```
localheinz, line 9: Still using mixed indentation in PHP files, but could be easily fixed after merging #559 and enabling the [`indentation_type`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/doc/rules/whitespace/indentation_type.rst) fixer.

cmb69:
Hmm, are we sure that all files are UTF-8 encoded? What about binary files (e.g. some images)?

localheinz:
@cmb69 

> Hmm, are we sure that all files are UTF-8 encoded? What about binary files (e.g. some images)?

I am not sure. 

> PS: well, binary files shouldn't be an issue.

👍

cmb69 (approved):
Looks like most, if not all, files are already UTF-8 encoded, and any accidential changes should be caught when reviewing the commit, so let's merge this.

Thank you for the RP!

PR #621, opened 2022-07-08
Enhancement: Add `Makefile`

localheinz:
This pull request

- [x] adds a `Makefile`

💁‍♂️ Run

```shell
make
```

or

```shell
make help
```

to show a list of available commands.

Run

```shell
make tests
```

to run the existing tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,7 @@
+.PHONY: help
+help: ## Displays this list of targets with descriptions
+	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[32m%-30s\033[0m %s\n", $$1, $$2}'
+
+.PHONY: tests
+tests: ## Runs tests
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -0,0 +1,10 @@
+.PHONY: it
+it: tests ## Runs the tests target
```
cmb69, line 2: Not sure if we should have that. I find `make` as shortcut for `make help` quite nice.

cmb69 (commented):
Thank you for the PR! This makes sense to me. (Although Windows users won't be able to use it, but I assume that isn't a real problem.)

Makefile
localheinz: @cmb69 

Adjusted!

cmb69 (approved):
Thank you!

localheinz:
@cmb69 

Once this has been merged, I can update #559 - or the other way around!

cmb69:
Thanks for this PR! I'll have a look at #559 soon.

localheinz:
Thank you, @cmb69

PR #622, opened 2022-07-08
Enhancement: Enable `indentation_type` fixer

localheinz:
This pull request

- [x] enables the `indentation_type` fixer

Follows #559.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/doc/rules/whitespace/indentation_type.rst.

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -10,6 +10,7 @@ $finder = $config->getFinder()
     ->name(__FILE__);
 
 $config->setRules([
+    'indentation_type' => true,
     'no_trailing_whitespace' => true,
     'visibility_required' => true,
     'whitespace_after_comma_in_array' => true,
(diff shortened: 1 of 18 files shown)


cmb69 (approved):
Thank you for the PR! Looks good to me.

I'm fine with using four spaces indent instead of tabs, and generally fine with transitioning to PSR-12 (which is likely one of the most common coding standards for PHP OSS projects).

localheinz:
@cmb69 

I believe that the specifics of a coding standard do not matter as long as they can be enforced automatically!

cmb69:
> I believe that the specifics of a coding standard do not matter as long as they can be enforced automatically!

Right! I should have added: ", so we hopefully all can agree on using PSR-12". I wouldn't want to discuss any single rule individually, but rather get this mess somewhat cleaned up. :)

localheinz:
@cmb69 

As an alternative to enabling fixers individually, I have opened #638!

localheinz:
@cmb69 

Rebased!

PR #623, opened 2022-07-08
Fix: Use `https` instead of `http` as protocol in links to `php.net`

localheinz:
This pull request

- [x] uses `https` instead of `http` as protocol in links to `php.net`

💁‍♂️ Not sure if this is a bit over the top, but I replaced `http://(.+\.)?php.net` with `https://$1php.net` in PhpStorm.


localheinz:
This seems hard to review - not sure if it's worth it, given that most website maintainers will usually redirect from `http` to `https`.

What do you think?

Ayesh:
It sure is a lot to review, but I like the idea behind this, that we gradually upgrade to HTTPS links. Some of the sites (such as `https://thebrainroom.com`) don't seem to support HTTPS, so perhaps we can automate it? I remember some "awesome" GitHub projects that list links to related projects automated a link checker, I will try to look it up.

We could also `sed` it for `*.php.net` and `github.com` as a starting point. These sites are the most common links we have, and seem to support HTTPS on all sub domains as well.

Ayesh:
As far as I can see, php.net sites such as `{pecl|pear|windows|gtk|conf|qa|bugs|news|wiki}.php.net` use the same HTTPS certificate with CN `*.php.net`, so I assume they are safe to use with HTTPS without a doubt because any issues with this certificate will alert pretty much everyone.

Looking at `{windows|downloads}.php.net` certificates on crt.sh, they seem to be automated, so they are safe to use too.

`{bk2|monitoring}.php.net` seem to be automated too, but I have never had any insight into who and how these sites run. Again, the crt.sh data shows the certificates are being renewed correctly.

---

So to summarize, I think we covered all php.net subdomains, and they all seem to be serving correctly over HTTPS (though, we [could improve it](https://www.ssllabs.com/ssltest/analyze.html?d=php.net&s=185.85.0.29&latest)).

That said, the change set looks good :) Thank you.

We might as well serve HSTS headers and preload it, but I'd be digressing, and might be better off discussing in a separate issue.

archive/1998.php
```diff
@@ -58,7 +58,7 @@
  conversion program will help you migrate your PHP 2 scripts to PHP 3.
 </p>
 <p>
- <a href="http://cvs.php.net/co.php/phpweb/Attic/release-3.0.txt?sa=1&amp;login=2&amp;r=1.2">Read
+ <a href="https://cvs.php.net/co.php/phpweb/Attic/release-3.0.txt?sa=1&amp;login=2&amp;r=1.2">Read
```
cmb69, line 61: cvs.php.net is no longer abailable (for years?)

urlhowto.php
```diff
@@ -45,7 +45,7 @@ function a($href): void {
  sites, not just at the main site. If you find that some of these
  shortcuts are not working on your mirror site, please report them
  as a "PHP.net Website Problem" at
- <a href="http://bugs.php.net/">http://bugs.php.net/</a>.
+ <a href="https://bugs.php.net/">http://bugs.php.net/</a>.
```
cmb69, line 48: This should point to issues, but likely more of that stuff can be removed, because the mirror program has been retired some years ago.

cmb69 (commented):
All in all, I think this makes sense, although there might be more general issues here (I've commented on a few). Anyhow, these issues are unrelated, and switching the protocol can be done. Could you please fix the merge conflicts?

samdark:
https://github.com/php/web-php/pull/287

derickr:
> As far as I can see, php.net sites such as {pecl|pear|windows|gtk|conf|qa|bugs|news|wiki}.php.net use the same HTTPS certificate with CN *.php.net, so I assume they are safe to use with HTTPS without a doubt because any issues with this certificate will alert pretty much everyone.

That is no longer true.

PR #628, opened 2022-07-08
Fix: Move and rename tests

localheinz:
This pull request

- [x] moves and renames tests

Follows #606.

### DIFF
--- a/tests/UserNotes/Sorter/sort-no-notes.phpt
+++ b/tests/UserNotes/Sorter/sort-no-notes.phpt
@@ -4,7 +4,7 @@ sort no notes
 precision=-1
 --FILE--
 <?php
-require_once __DIR__ . "/../autoload.php";
+require_once __DIR__ . "/../../../autoload.php";
 
 $notes = [];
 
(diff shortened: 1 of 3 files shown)


tests/UserNotes/Sorter/sort_notes_001.phpt
```diff
@@ -4,7 +4,7 @@ sort some notes
 precision=-1
 --FILE--
 <?php
-require_once __DIR__ . "/../autoload.php";
+require_once __DIR__ . "/../../../autoload.php";
```
localheinz, line 7: If we used `phpunit/phpunit`, I would create a test for `phpweb/UserNotes/Sorter` in one of

- `phpweb/Test/UserNotes/SorterTest
- `phpweb/Test/Unit/UserNotes/SorterTest

to have some sort of 1:1 mapping of class and test case.

Does that make sense?

kamil-tekiela:
How about using more descriptive file names?

kamil-tekiela (approved):

localheinz:
Thank you, @kamil-tekiela!

PR #629, opened 2022-07-08
Fix: Clean up `robots.txt`

localheinz:
This pull request

- [x] removes entries from `robots.txt`

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -7,7 +7,3 @@ Disallow: /search.php
 Disallow: /mod.php
 Disallow: /manual/add-note.php
 Disallow: /manual/vote-note.php
-
-Disallow: /harming/humans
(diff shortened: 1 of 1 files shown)


robots.txt
```diff
@@ -7,7 +7,3 @@ Disallow: /search.php
 Disallow: /mod.php
 Disallow: /manual/add-note.php
 Disallow: /manual/vote-note.php
-
-Disallow: /harming/humans
-Disallow: /ignoring/human/orders
-Disallow: /harm/to/self
```
localheinz, line 13: ❓

kamil-tekiela: This is an easter egg. It refers to the three rows of robotics. It's funny. Leave it.

localheinz: Ha!

PR #630, opened 2022-07-09
Fix: Remove unnecessary closing tags

localheinz:
This pull request

- [x] removes unnecessary closing tags

Follows #606.


cmb69:
I actually prefer closing tags in PHPT files, because of the improved experience with editors/IDEs in PHP mode. It's a bit similar to PHP template files.

PR #631, opened 2022-07-09
Fix: Replace a script with tests

localheinz:
This pull request

- [x] replaces a script with tests

Follows https://github.com/php/web-php/issues/579#issuecomment-1168732201.


cmb69:
Ah, right, thank you! But what about turning this into an actual PHPT test? Basically the same logic and output, but as part of the test suite and not as requestable script.

localheinz:
@cmb69 

Thank you for pointing out, I was not aware this was supposed to be a test!

I split the script into two tests!

cmb69:
Thank you!

localheinz:
Thank you, @cmb69!

PR #626, opened 2022-07-08
Enhancement: Extract constants to avoid mutable global state

localheinz:
This pull request

- [x] extracts constants to avoid mutable global state

PR #632, opened 2022-07-11
Enhancement: Add tests for `gen_challenge()` and `test_answer()`

localheinz:
This pull request

- [x] adds tests for `gen_challenge()` and `test_answer()`

Follows #626.

### DIFF
--- a/tests/gen-challenge.phpt
+++ b/tests/gen-challenge.phpt
@@ -1,0 +1,246 @@
+--TEST--
+gen_challenge() generates a spam challenge
+--FILE--
+<?php
+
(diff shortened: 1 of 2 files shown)


cmb69 (approved):
Thank you for the PR! Testing `test_answer()` definitely makes sense. I'm not sure about the `gen_challenge()` test, since using `rand()` might not be the best idea anyway, so that we needed to mock the random number generator anyway. OTOH, we could still rewrite the test, so I guess this is good as is.

localheinz:
@cmb69 

Are there any objections to merging this pull request?

cmb69:
Thank you!

PR #633, opened 2022-07-11
Enhancement: Trim trailing whitespace in all files

localheinz:
This pull request

- [x] adjusts `.editorconfig` to trim trailing whitespace in all (!) files

Similar to 620.

💁‍♂️ Perhaps we should apply this rule after merging #559 and after enabling the [`no_trailing_whitespace`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/doc/rules/whitespace/no_trailing_whitespace.rst) fixer.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,15 +2,14 @@ root = true
 
 [*]
 insert_final_newline = true
+trim_trailing_whitespace = true
 
 [*.yaml]
 charset = utf-8
 indent_size = 2
 indent_style = space
-trim_trailing_whitespace = true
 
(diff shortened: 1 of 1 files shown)


cmb69:
Yeah, I think we should run the `trailing_whitespace_fixer` *at least* locally before merging this. It is *possible* the a few cases of trailing whitespace are intentional (what we may want to fix then).

cmb69:
Thank you!

localheinz:
Thank you, @cmb69!

PR #634, opened 2022-07-11
Enhancement: Enable `array_indentation` fixer

localheinz:
This pull request

- [x] enables the `array_indentation` fixer
- [x] excludes `manual/en/`

Follows #559.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/doc/rules/whitespace/array_indentation.rst.

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -7,6 +7,7 @@ $config = new PhpCsFixer\Config();
 $finder = $config->getFinder()
     ->ignoreDotFiles(false)
     ->in(__DIR__)
+    ->exclude('manual/en/')
     ->name(__FILE__);
 
 $config->setRules([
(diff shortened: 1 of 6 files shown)


download-docs.php
```diff
@@ -43,8 +43,8 @@
 $formats = array(
     "Single HTML file" => "html.gz",
     "Many HTML files"  => "tar.gz",
-#   "Many PDF files"   => "pdf.tar.gz",
-#   "PDF"              => "pdf",
+    #   "Many PDF files"   => "pdf.tar.gz",
+    #   "PDF"              => "pdf",
```
localheinz, line 47: Perhaps we should remove this instead?

cmb69: Already happened thanks to your PR!

cmb69 (commented):
Thank you! Makes perfect sense to me, but I think we shouldn't fix any files under manual/en, since these are just example files taken from the documentation (we may consider to update these from the docs, but that's another story).

localheinz:
@cmb69 

Updated!

cmb69:
Thank you!

PR #635, opened 2022-07-11
Fix: Remove commented out formats

localheinz:
This pull request

- [x] removes commented out formats


cmb69:
Ah, yeah, it's highly unlikely that PDFs will be made available in the future, since the rendering is broken for ~10years.

Thank you!

localheinz:
Thank you, @cmb69!

PR #638, opened 2022-07-11
Enhancement: Enable `PSR12` rule set

localheinz:
This pull request

- [x] enables the `PSR12` rule set

Follows #559.


localheinz:
@cmb69 

I think this does not work very well:

- changes are hard to review
- mixing PHP code and HTML code is something that is not officially supported by `friendsofphp/php-cs-fixer`

When enabling fixers one by one, we have more control.

cmb69:
Yes, I have to agree!

PR #640, opened 2022-07-11
Enhancement: Enable `single_space_after_construct` fixer

localheinz:
This pull request

- [x] enables the `single_space_after_construct` fixer
- [ ] runs `make coding-standards`

Follows #559.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/doc/rules/language_construct/single_space_after_construct.rst.


releases/feed.php
```diff
@@ -18,7 +18,7 @@
 
 XML;
 
-/* FIX silly editor highlighting */?><?php
+/* FIX silly editor highlighting */ ?><?php
```
cmb69, line 21: I think we should remove this line altogether (probably the highlighting has been fixed in all relevant editors in the meantime). Can be done with a separate commit/PR, though.

cmb69 (commented):
Thank you! This looks good to me, but I wonder why this fixer is not listed as being part of the PSR-12 ruleset.

manual/vote-note.php
```diff
@@ -3,7 +3,7 @@
 include_once __DIR__ . '/../include/prepend.inc';
 include_once __DIR__ . '/../include/posttohost.inc';
 include_once __DIR__ . '/../include/shared-manual.inc';
-include      __DIR__ . '/spam_challenge.php';
+include __DIR__ . '/spam_challenge.php';
```
cmb69, line 6: spam_challenge.php contains only definitions, so `include_once` is more appropriate. That should probably be changed by a separate PR/commit, though.

localheinz:
Thank you, @cmb69!

manual/vote-note.php
cmb69: I already [went with `include_once`](https://github.com/php/web-php/commit/16fd981d27a6d6c6adbaecf06da58e8a35741c0b) for consistency with the others. We probably want to turn that into a class, anyway, and can use autoloading then.

PR #641, opened 2022-07-11
Enhancement: Enable `ordered_class_elements` fixer

localheinz:
This pull request

- [x] enables the `ordered_class_elements` fixer
- [x] excludes `tests/run-tests.php`

Follows #559.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/doc/rules/class_notation/ordered_class_elements.rst.

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -8,7 +8,8 @@ $finder = $config->getFinder()
     ->ignoreDotFiles(false)
     ->in(__DIR__)
     ->exclude('manual/en/')
-    ->name(__FILE__);
+    ->name(__FILE__)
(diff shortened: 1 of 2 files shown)


cmb69:
Thank you! This makes sense to me, but I think we should exclude tests/run-tests.php from *all* fixes, since that file is just a copy of php-src/run-tests.php.

localheinz:
@cmb69 

Updated!

localheinz:
@cmb69 

Rebased!

cmb69:
Great! Thank you!

localheinz:
Thank you, @cmb69!

PR #651, opened 2022-07-12
JavaScript tidying up

kamil-tekiela:
This PR upgrades jQuery from 1.10 to 3.6 and some of the libraries. We also switch from distributing our own copies to using cloudflare CDN. 

UItoTop plugin wasn't used for 9 years and is not even maintained anymore from what I can see. It was a very simple piece of JS code anyway. 

As of  2022 we can finally drop support for Internet Explorer so let's remove all the shims.


cmb69:
Thank you! I'm all for upgrading jQuery, and I see also no problems with removing the IE fallbacks (I don't think they would be necessary for IE 11 anyway). I'm not sure about using Cloudflare CDN, though, for privacy reasons (GDPR etc.) Is there a real issue with delivering these assets as we did before?

cmb69:
> IANAL

Me neither. :)

But since we're already using the MyraCloud CDN, I would stick with it. Only downside I see is that jQuery and the plugins won't be shared for different Websites; doesn't look like a real problem, I guess.

> and storing third-party libraries in our VCS is not very tidy. Should I at least update these libraries?

Having third-party libraries in the repo isn't a problem per se, in my opinion. The fact that we don't care to update these, is, though. Maybe dependabot could help with this? It might make sense to fetch these libraries via a JS dependency manager, so we are informed about available updates.

Anyhow, please update the scripts manually for now.

cmb69 (approved):
Thank you!

morrisonlevi:
A big thanks from me as well! It's been 10+ years since the last major redesign. This cruft needs to go to pave the way for the future. Again, thank you!

PR #655, opened 2022-07-14
Fix mousetrap.js

kamil-tekiela:
Turns out that the file we had stored locally for this library contained an official extension. While I updating, I accidently removed that extension. Let's bring it back. Copied from https://raw.githubusercontent.com/ccampbell/mousetrap/master/plugins/pause/mousetrap-pause.min.js


cmb69 (approved):
Good catch, thank you!

PR #656, opened 2022-07-14
Improve code formatting in JS

kamil-tekiela:
This change only applies more consistent code formatting to the common.js file. I understand it's highly opinionated, but I do find this easier to read. This was done using TSC compiled, automatic formatted, and some manual adjustments. This PR contains no refactoring, code clean-up or any other changes to the code. It's all purely visual.

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -1,24 +1,24 @@
 /* Plugins, etc, are on top. */
 
-String.prototype.escapeSelector = function() {
-  return this.replace(/(.|#)([ #;&,.+*~\':"!^$\[\]\(\)=>|\/])/g, '$1' + '\\\\$2');
+String.prototype.escapeSelector = function () {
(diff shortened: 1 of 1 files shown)


cmb69:
Can we have that formatting automated, like we did with php-cs-fixer? We could start with only this file.

kamil-tekiela:
I don't know if we could automate that. We could think of simply using TSC to compile JS and commit that, but TSC removes blank lines... so in terms of readability, it's not great. I don't know what other automatic formatter we could use that would give us good results. I think there are a number of linters and different coding styles available for JavaScript. I think as a first step, what I propose is sensible. At least the formatting is more uniform in this one file, even if we don't enforce any particular coding style for JS.

cmb69 (approved):
I *guess* we could use [ESLint](https://eslint.org/docs/latest/rules/#layout--formatting), but I'm okay with merging this PR.

TIL about "bork" :)

PR #657, opened 2022-07-15
Enhancement: Enable and configure `concat_space` fixer

localheinz:
This pull request

- [x] enables and configures the `concat_space` fixer

Follows #559.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/doc/rules/operator/concat_space.rst.

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -22,6 +22,9 @@ $config
         'array_indentation' => true,
         'array_syntax' => true,
         'class_attributes_separation' => true,
+        'concat_space' => [
+            'spacing' => 'one',
+        ],
         'constant_case' => true,
         'indentation_type' => true,
         'line_ending' => true,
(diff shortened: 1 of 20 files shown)


.php-cs-fixer.php
```diff
@@ -13,6 +13,9 @@
 
 $config->setRules([
     'array_indentation' => true,
+    'concat_space' => [
+        'spacing' => 'one',
```
localheinz, line 17: We could also use `none` here, but fewer files were modified when using `one`.

kamil-tekiela (approved):

.php-cs-fixer.php
cmb69: I prefer `one`. :)

cmb69 (approved):
Thank you!

Maybe wait a bit for review of others.

cmb69:
Thank you!

localheinz:

PR #658, opened 2022-07-15
Enhancement: Enable `strict_param` fixer

localheinz:
This pull request

- [x] enables the (risky) `strict_param` fixer
- [x] sets the strict param to `false` to ensure the current behavior

Follows #559.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/doc/rules/strict/strict_param.rst.

### DIFF
--- a/error.php
+++ b/error.php
@@ -175,7 +175,7 @@ if (preg_match('/^GH-(\d+)$/', $URI, $matches)) {
 // Redirect if the entered URI was a PHP page name (except some pages,
 // which we display in the mirror's language or the explicitly specified
 // language [see below])
-if (!in_array($URI, array('mirror-info', 'error', 'mod')) &&
+if (!in_array($URI, array('mirror-info', 'error', 'mod'), true) &&
     file_exists($_SERVER['DOCUMENT_ROOT'] . "/$URI.php")) {
     mirror_redirect("/$URI.php");
 }
--- a/error.php
+++ b/error.php
@@ -175,7 +175,7 @@ if (preg_match('/^GH-(\d+)$/', $URI, $matches)) {
 // Redirect if the entered URI was a PHP page name (except some pages,
 // which we display in the mirror's language or the explicitly specified
 // language [see below])
-if (!in_array($URI, array('mirror-info', 'error', 'mod'), true) &&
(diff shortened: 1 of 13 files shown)


error.php
```diff
@@ -175,7 +175,7 @@
 // Redirect if the entered URI was a PHP page name (except some pages,
 // which we display in the mirror's language or the explicitly specified
 // language [see below])
-if (!in_array($URI, array('mirror-info', 'error', 'mod')) &&
+if (!in_array($URI, array('mirror-info', 'error', 'mod'), false) &&
```
localheinz, line 178: As soon as we are sure about the types here (and in all other cases), we could set the strict parameters to `true`.

cmb69: > As soon as we are sure about the types here (and in all other cases),

I guess that is the problem. Wouldn't it be better to apply this fixer only after we're sure? Or do you have new code in mind?

localheinz: @cmb69 

Yes, it could also well be used to prevent new code being added without using the strict parameter explicitly!

include/manual-lookup.inc
```diff
@@ -107,7 +107,7 @@ function find_manual_page($lang, $keyword)
     // when rsync updates the database
     $dbh = false;
     if (class_exists('PDO')) {
-        if (in_array('sqlite', PDO::getAvailableDrivers())) {
+        if (in_array('sqlite', PDO::getAvailableDrivers(), true)) {
```
localheinz, line 110: This looks safe to keep.

manual/help-translate.php
```diff
@@ -28,7 +28,7 @@
 
 foreach ($INACTIVE_ONLINE_LANGUAGES as $cc => $lang) {
     $link = 'no archive';
-    if (in_array($cc, $archived)) {
+    if (in_array($cc, $archived, true)) {
```
localheinz, line 31: This looks safe to keep.

error.php
cmb69: Okay, makes sense.

localheinz:
Thank you, @cmb69!

PR #665, opened 2022-07-15
Update usergroups.json

ehsanamd:


backend/usergroups.json
```diff
@@ -1 +1,1901 @@
...
\ No newline at end of file
+{
```
localheinz, line 1: Perhaps this file is generated somehow?

ehsanamd: I used 'Prettier extension' in VS code to format it.

ehsanamd (commented):
It's formatted by the Prettier extension.

PR #666, opened 2022-07-16
Fix: Extract closure to allow using early returns

localheinz:
This pull request

- [x]x extracts a closure to allow using early returns

Related to #664.

### DIFF
--- a/manual/add-note.php
+++ b/manual/add-note.php
@@ -44,48 +44,49 @@ if ($process) {
         $user = "Anonymous";
     }
 
-    // We don't know of any error now
-    $error = false;
+    $error = (static function (string $note, string $user, array $post): ?string {
+        // No note specified
+        if (strlen($note) == 0) {
+            return "You have not specified the note text.";
+        }
 
-    // No note specified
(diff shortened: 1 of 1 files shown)


manual/add-note.php
```diff
@@ -44,48 +44,49 @@
...
-    // We don't know of any error now
-    $error = false;
-
-    // No note specified
-    if (strlen($note) == 0) {
-        $error = "You have not specified the note text.";
-    }
-
-    // SPAM challenge failed
-    elseif (!test_answer($_POST['func'], $_POST['arga'], $_POST['argb'], $_POST['answer'])) {
```
localheinz, line 56: This expression here is a bit difficult to refactor in the light of #664:

As @kamil-tekiela pointed out, it makes sense to fail early, and in order to fail early, I would need to add at least one condition here and a `try`-`catch` block, to attempt creating a challenge from

- `$_POST['func']`
- `$_POST['arga']`
- `$_POST['argb']`

and if creating a challenge succeeded, compare the expected with the actual answer.

PR #669, opened 2022-07-16
Enhancement: Enable `switch_case_space` fixer

localheinz:
This pull request

- [x] enables the `switch_case_space` fixer

Follows #559.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/doc/rules/control_structure/switch_case_space.rst.

### DIFF
--- a/error.php
+++ b/error.php
@@ -122,9 +122,9 @@ $URI = rtrim($URI, '/');
 // Some nice URLs for getting something for download
 if (preg_match("!^get/([^/]+)$!", $URI, $what)) {
     switch ($what[1]) {
-        case "php"           : $URI = "downloads"; break;
-        case "docs"          : // intentional
-        case "documentation" : $URI = "download-docs"; break;
+        case "php": $URI = "downloads"; break;
+        case "docs": // intentional
+        case "documentation": $URI = "download-docs"; break;
     }
 }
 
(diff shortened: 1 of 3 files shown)


error.php
```diff
@@ -122,9 +122,9 @@
 // Some nice URLs for getting something for download
 if (preg_match("!^get/([^/]+)$!", $URI, $what)) {
     switch ($what[1]) {
-        case "php"           : $URI = "downloads"; break;
-        case "docs"          : // intentional
-        case "documentation" : $URI = "download-docs"; break;
+        case "php": $URI = "downloads"; break;
+        case "docs": // intentional
+        case "documentation": $URI = "download-docs"; break;
```
cmb69, line 127: I'd reformat manually right away:
```suggestion
        case "php":
            $URI = "downloads";
            break;
        case "docs":
        case "documentation":
            $URI = "download-docs";
            break;
```

cmb69 (commented):
Thank you! Adding the switch_case_space fixer looks reasonable to me, but I'd reformat one of the switches (partially to avoid discussion of whether it is better to align the colons).

kamil-tekiela (approved):

localheinz:
@cmb69 @kamil-tekiela 

Is this good to go?

cmb69:
> Is this good to go?

Yes. Thank you!

PR #670, opened 2022-07-16
Fix: Use file name, not path to file

localheinz:
This pull request

- [x] uses the file name, not the path to `.php-cs-fixer.php`

Follows #559.


cmb69:
I guess I would have written it this way, but now wonder if that is really better.

localheinz:
@cmb69 

It does not work with using the full path - when configuring the finder, everything is relative to the directory (or directories) that is (or are) used in the `in()` call.

cmb69:
Thanks for the explanation!

PR #671, opened 2022-07-16
Enhancement: Enable and configure `header_comment` fixer

localheinz:
❗ Blocked by #670.

This pull request

- [x] enables and configures the `header_comment` fixer

Follows #559.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/doc/rules/comment/header_comment.rst.

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -1,5 +1,11 @@
 <?php
 
+$header = <<<TXT
+Any ideas?
+
+@see https://github.com/php/web-php
+TXT;
+
 $finder = PhpCsFixer\Finder::create()->in(__DIR__);
 
 $config = new PhpCsFixer\Config();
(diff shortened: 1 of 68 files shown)


.php-cs-fixer.php
```diff
@@ -1,5 +1,17 @@
 <?php
 
+/**
+ * Any ideas?
+ *
+ * @see https://github.com/php/web-php
+ */
+
+$header = <<<TXT
+Any ideas?
```
localheinz, line 10: ❗

cached.php
```diff
@@ -1,15 +1,11 @@
 <?php
-/*
-  Yes, we know this can be used to view the source for any file
```
localheinz, line 3: Probably makes sense to keep these, but move them.

kamil-tekiela:
I don't like this and I prefer not to do this. I see no real value in this as it's just comments. They are not even phpdoc. I would remove those strange `C++` comments, but I would not add any new header comments.

PR #675, opened 2022-07-22
PHP 8.2 Release page

saundefined:
- [x] Update logo
- [x] Finish the English version
- [x] Put messages in lang-files
- [x] Add links to Documentation (php/doc-en#1799)
- [ ] Waiting for the release of PHP 8.2 🎉 

cc @pronskiy 

Preview: https://php.sergeypanteleev.com/releases/8.2/en.php


bukka:
Would be good if this mentioned https://github.com/php/php-src/pull/8094 as it is one of the bigger changes...

releases/8.2/release.inc
```diff
@@ -0,0 +1,419 @@
...
+            </h2>
+            <div class="php8-compare__main">
+                <div class="php8-compare__block example-contents">
+                    <div class="php8-compare__label">PHP &lt; 8.2</div>
+                    <div class="php8-code phpcode">
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
+class Foo {
+    public function bar(mixed $entity) {
+        if($entity === null) {
```
bukka, line 156:
```suggestion
        if ($entity === null) {
```
same for all other ifs... :)

bukka:
Ah I actually see one line about that: `<li>New <code>Random</code> extension.</li>` so maybe a bit more extend that would be awesome.

cmb69:
At least in the long run, it appears to be sensible to figure out a way to avoid the code duplication and the mix of PHP and HTML in these files.

releases/8.2/release.inc
```diff
@@ -0,0 +1,460 @@
...
+                    <div class="php8-code phpcode">
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
+$array = ['foo', 'bar', 'baz'];
+
+$engine = new Random\Engine\Mt19937(1234, MT_RAND_PHP);
+$randomizer = new Random\Randomizer($engine);
+$single = $randomizer->pickArrayKeys($array, 1)[0]; // (int) 0
+ // or: [$single] = $randomizer->pickArrayKeys($array, 1);
+$multiple = $randomizer->pickArrayKeys($array, 2); // (array) [1, 2]
```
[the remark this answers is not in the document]
TimWolla: Also I would move this upwards in the list. This one is certainly more interesting to the average developer than the deprecation of `${}` interpolation.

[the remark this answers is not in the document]
cmb69: > This example does not paint the random extension in a good light, because it is more complex than the old version.

ACK. It is hard, though, to show the actual advantages of ext/random, since there was no equivalent behavior previously (short off implementing your own RNG).

> If I had a little more time in the elevator (skyscraper!)

You could have used the stairway. :p (SCNR)

Pythone:
Can the new logo have Very Peri as one of the colors used?
https://veryperi.me/
2022 pantone color of the year

releases/8.2/common.php
```diff
@@ -0,0 +1,74 @@
...
+    // We don't use the general language selection of php.net,
+    // so soldier on with this one.
+    return;
+}
+
+function common_header(string $description): void {
+    global $MYSITE;
+
+    $meta_image_path = \htmlspecialchars(
+        \filter_var($MYSITE . 'images/php8/php_8_1_released.png', \FILTER_VALIDATE_URL));
```
TimWolla, line 17: This still references some 8.1.

releases/8.2/release.inc
```diff
@@ -0,0 +1,470 @@
...
+$user = new User();
+$user->last_name = 'Doe'; // Deprecated notice
+PHP
+                        ); ?>
+                    </div>
+                </div>
+            </div>
+        </div>
+        <div class="php8-compare__content">
+            You can use the <code>#[AllowDynamicProperties]</code> attribute on classes that should still allow those
```
pronskiy, line 244: This is quite a major change. How about adding more information?

- Dynamic properties are allowed on `stdClass` and its inheritors.
- `__get()`/`__set()` behaviour is not affected.
- Consider using explicit properties or add the `#[AllowDynamicProperties]` attribute on classes that should still allow non-existing properties access.

saundefined: A bit updated a message, what do you think?

pronskiy: I like the code example 👍 I would still prefer separate points in the description to emphasize that `#[AllowDynamicProperties]` is more like a last resort, not a recommended go-to. 
@cmb69, @TimWolla, @nikic, what do you think?

cmb69: I agree that `AllowDynamicProperties` is meant as last (or temporary) resort, and that we should clarify that here. It seems to me that this new behavior might be the most controversial change in PHP 8.2, so we should do our best to clarify as much as possible.

saundefined: >`stdClass` objects and their children or magic methods `__get`/`__set` are not affected.
>The `#[AllowDynamicProperties]` attribute was added as a way to opt-in to the use of dynamic properties.

What do you think of this wording?

releases/8.2/release.inc
```diff
@@ -0,0 +1,476 @@
...
+                </div>
+            </div>
+        </div>
+        <div class="php8-compare__content">
+            <p>The new <code>random</code> extension organizes and consolidates the existing PHP functionality
+                related to random number generation.</p>
+            <p>One of the most important new additions with the new <code>random</code> extension
+                is the <code>Random\Randomizer</code> class. It is meant to provide an object-oriented API
+                to access all Random Number generation functionality with a choice of Pseudo Random Number Generator
+                algorithm, which can be swapped out with a new implementation.</p>
```
TimWolla, line 401: I'm not happy with this description, how about something like this, including my elevator pitch from here: https://github.com/php/web-php/pull/675/files#r933237161

> The random extension provides a new object-oriented API to random number generation. Instead of relying on a globally seeded random number generator (RNG) using the Mersenne Twister algorithm the object-oriented API provides several classes (“Engine”s) providing access to modern algorithms that store their state within objects to allow for multiple independent seedable sequences. The \Random\Randomizer class provides a high level interface to use the engine's randomness to generate a random integer, to shuffle an array or string, to select random array keys and more.

Use my proposal as you wish. It likely needs be shortened a rephrased a little.

releases/8.2/release.inc
```diff
@@ -0,0 +1,476 @@
...
+
+        </div>
+
+        <div class="php8-column">
+            <h2 class="php8-h2" id="other_new_things">New Classes, Interfaces, and Functions</h2>
+            <div class="php8-compare__content php8-compare__content--block">
+                <ul>
+                    <li>New <code>mysqli_execute_query</code> function and <code>mysqli::execute_query</code> method.
+                    </li>
+                    <li>New <code>#[AllowDynamicProperties]</code> and <code>#[\SensitiveParameter]</code> attributes.
```
TimWolla, line 417:
```suggestion
                    <li>New <code>#[\AllowDynamicProperties]</code> and <code>#[\SensitiveParameter]</code> attributes.
```

For consistency.

releases/8.2/release.inc
```diff
@@ -0,0 +1,484 @@
...
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
+$array = ['foo', 'bar', 'baz'];
+
+$engine = new Random\Engine\Mt19937(1234, MT_RAND_PHP);
+$randomizer = new Random\Randomizer($engine);
+$single = $randomizer->pickArrayKeys($array, 1)[0]; // (int) 0
+
+// or: [$single] = $randomizer->pickArrayKeys($array, 1);
+$multiple = $randomizer->pickArrayKeys($array, 2); // (array) [1, 2]
```
TimWolla, line 391: My [original comment thread](https://github.com/php/web-php/pull/675#discussion_r933237161) already is collapsed, possibly because of some unrelated change marking it as outdated.

The example really should be revisited once more, especially since it uses the horrible horrible `MT_RAND_PHP` option which really no one should be using (I hope to deprecate that one in PHP 8.3).

I've posted a possible example (based on fibers) a while ago in StackOverflow's PHP room. I'm not sure it is a *good example*, but I'd say it's better than the current one:

https://chat.stackoverflow.com/transcript/message/55014282#55014282

saundefined: @TimWolla Thanks for the example, I put it without comparing it with PHP < 8.2, what do you think of this solution? 🤔

TimWolla: @saundefined Thanks, I just cloned the branch locally to see the rendering and it's looking good to me.

One thing I would do though: Move the section for the random extension further up in the page and put the deprecations last (Dynamic Properties, `${}` interpolation). IMO the page should more prominently explain what great new stuff is possible with PHP 8.2, not what no longer works. We want to give the users reasons to upgrade to 8.2 after all.

Suggested order:

- Readonly classes (this is likely the most interesting new feature)
- DNF types (second most interesting)
- null, false, true as standalone types (to keep the two type-related features together, DNF types are more interesting)
- Random extension (this is the "bigger" feature compared to constants in traits)
- Constants in traits
- Deprecate dynamic properties
- Deprecate `${}` interpolation (maybe drop this entirely? The deprecation isn't *that* interesting that it needs such a prominent spot, perhaps the `SensitiveParameter` attribute could be placed below constants in traits instead?)

saundefined: @TimWolla thanks, changed the order 👍

saundefined:
@lex111 if you have a time, can you also check the Russian translation?

releases/8.2/languages/ru.php
```diff
@@ -0,0 +1,40 @@
...
+    'bc_string_interpolation' => 'Устарела интерполяция строк <code>${}</code>.',
+    'bc_utf8' => 'Устарели функции <code>utf8_encode</code> и <code>utf8_decode</code>.',
+    'bc_datetime' => 'У методов <code>DateTime::createFromImmutable</code> и <code>DateTimeImmutable::createFromMutable</code> предварительный тип возвращаемого значения <code>static</code>.',
+    'bc_odbc' => 'Модули <code>ODBC</code> и <code>PDO_ODBC</code> экранирует имя пользователя и пароль.',
+    'bc_str_locale_sensitive' => 'Функции <code>strtolower</code> и <code>strtoupper</code> чувствительны к местным условиям.',
+    'bc_spl_enforces_signature' => 'Методы <code>SplFileObject::getCsvControl</code>, <code>SplFileObject::fflush</code>, <code>SplFileObject::ftell</code>, <code>SplFileObject::fgetc</code> и <code>SplFileObject::fpassthru</code> усиливают свою сигнатуру.',
+    'bc_spl_false' => 'У метода <code>SplFileObject::hasChildren</code> предварительный тип возвращаемого значения <code>false</code>.',
+    'bc_spl_null' => 'У метода <code>SplFileObject::getChildren</code> предварительный тип возвращаемого значения <code>null</code>.',
```
lex111, line 35:
```suggestion
    'bc_spl_null' => 'У метода <code>SplFileObject::getChildren</code> предварительный тип возвращаемого значения задан как <code>null</code>.',
```

releases/8.2/languages/ru.php
```diff
@@ -0,0 +1,40 @@
...
+    'bc_spl_deprecated' => 'Устарел внутренний метод <code>SplFileInfo::_bad_state_ex</code>.',
+    'footer_title' => 'Выше производительность, лучше синтаксис, надёжнее система типов.',
+    'footer_description' => '<p>Для загрузки исходного кода PHP 8.2 посетите страницу <a href="https://www.php.net/downloads">Downloads</a>. Бинарные файлы Windows находятся на сайте <a href="https://windows.php.net/download">PHP for Windows</a>. Список изменений перечислен на странице <a href="https://www.php.net/ChangeLog-8.php#PHP_8_2">ChangeLog</a>.</p>
+        <p><a href="/manual/ru/migration82.php">Руководство по миграции</a> доступно в разделе документации.Ознакомьтесь с ним, чтобы узнать обо всех новых возможностях и изменений, затрагивающих обратную совместимость.</p>',
```
lex111, line 39:
```suggestion
        <p><a href="/manual/ru/migration82.php">Руководство по миграции</a> доступно в разделе документации. Ознакомьтесь с ним, чтобы узнать обо всех новых возможностях и изменениях, затрагивающих обратную совместимость.</p>',
```

releases/8.2/release.inc
```diff
@@ -0,0 +1,423 @@
...
+                        $lang
+                    ) ?></a>
+            </h2>
+            <div class="php8-compare__main">
+                <div class="example-contents example-contents-full">
+                    <div class="php8-compare__label php8-compare__label_new">PHP 8.2</div>
+                    <div class="php8-code phpcode">
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
+$blueprintRng = new \Random\Engine\Xoshiro256StarStar("Example seed string with 256 Bit");
```
TimWolla, line 220:
```suggestion
$blueprintRng = new Xoshiro256StarStar(
    hash('sha256', "Example seed that is converted to a 256 Bit string via SHA-256", true)
);
```

This is a little cleaner: By hashing one can use arbitrarily sized seed strings and the values are better distributed.

releases/8.2/release.inc
```diff
@@ -0,0 +1,423 @@
...
+
+$fibers = [];
+for ($i = 0; $i < 8; $i++) {
+	$fiberRng = clone $blueprintRng;
+	$blueprintRng->jump(); // Moves the blueprint ahead 2**128 steps, giving the Fiber 2**128 unique values without needing to reseed.
+
+	$fibers[] = new Fiber(function () use ($fiberRng, $i): void {
+		$randomizer = new \Random\Randomizer($fiberRng);
+
+		echo "$i: " . $randomizer->getInt(0, 100), PHP_EOL;
```
TimWolla, line 230:
```suggestion
		echo "{$i}: " . $randomizer->getInt(0, 100), PHP_EOL;
```
To highlight the "best practice" with regard to string interpolation.

releases/8.2/release.inc
```diff
@@ -0,0 +1,423 @@
...
+                    <div class="php8-compare__label php8-compare__label_new">PHP 8.2</div>
+                    <div class="php8-code phpcode">
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
+$blueprintRng = new \Random\Engine\Xoshiro256StarStar("Example seed string with 256 Bit");
+
+$fibers = [];
+for ($i = 0; $i < 8; $i++) {
+	$fiberRng = clone $blueprintRng;
+	$blueprintRng->jump(); // Moves the blueprint ahead 2**128 steps, giving the Fiber 2**128 unique values without needing to reseed.
```
TimWolla, line 225:
```suggestion
	// Xoshiro256**'s 'jump()' method moves the blueprint ahead 2**128 steps, as if calling 'generate()' 2**128 times,
	// giving the Fiber 2**128 unique values without needing to reseed.
	$blueprintRng->jump(); 
```

Improved explanation.

releases/8.2/release.inc
```diff
@@ -0,0 +1,423 @@
...
+                        'documentation',
+                        $lang
+                    ) ?></a>
+            </h2>
+            <div class="php8-compare__main">
+                <div class="example-contents example-contents-full">
+                    <div class="php8-compare__label php8-compare__label_new">PHP 8.2</div>
+                    <div class="php8-code phpcode">
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
```
TimWolla, line 219:
```suggestion
                            <<<'PHP'
use Random\Engine\Xoshiro256StarStar;
use Random\Randomizer;

```

releases/8.2/release.inc
```diff
@@ -0,0 +1,423 @@
...
+$blueprintRng = new \Random\Engine\Xoshiro256StarStar("Example seed string with 256 Bit");
+
+$fibers = [];
+for ($i = 0; $i < 8; $i++) {
+	$fiberRng = clone $blueprintRng;
+	$blueprintRng->jump(); // Moves the blueprint ahead 2**128 steps, giving the Fiber 2**128 unique values without needing to reseed.
+
+	$fibers[] = new Fiber(function () use ($fiberRng, $i): void {
+		$randomizer = new \Random\Randomizer($fiberRng);
```
TimWolla, line 228:
```suggestion
		$randomizer = new Randomizer($fiberRng);
```

releases/8.2/release.inc
```diff
@@ -0,0 +1,423 @@
...
+
+		echo "$i: " . $randomizer->getInt(0, 100), PHP_EOL;
+	});
+}
+
+// The randomizer will use a CSPRNG by default.
+$randomizer = new \Random\Randomizer();
```
TimWolla, line 235:
```suggestion
$randomizer = new Randomizer();
```

TimWolla (commented):
I have some improvements for the ext/random example, to improve the explanation and cleanliness. Full code:

```php
<?php

use Random\Engine\Xoshiro256StarStar;
use Random\Randomizer;

$blueprintRng = new Xoshiro256StarStar(
	hash('sha256', "Example seed that is converted to a 256 Bit string via SHA-256", true)
);

$fibers = [];
for ($i = 0; $i < 8; $i++) {
	$fiberRng = clone $blueprintRng;
	// Xoshiro256**'s 'jump()' method moves the blueprint ahead 2**128 steps, as if
	// calling 'generate()' 2**128 times, giving the Fiber 2**128 unique values
	// without needing to reseed.
	$blueprintRng->jump(); 

	$fibers[] = new Fiber(function () use ($fiberRng, $i): void {
		$randomizer = new Randomizer($fiberRng);

		echo "{$i}: " . $randomizer->getInt(0, 100), PHP_EOL;
	});
}

// The randomizer will use a CSPRNG by default.
$randomizer = new Randomizer();

// Even though the fibers execute in a random order, they will print the same value
// each time, because each has its own unique instance of the RNG.
$fibers = $randomizer->shuffleArray($fibers);
foreach ($fibers as $fiber) {
	$fiber->start();
}
```

saundefined:
Hey there!

Preparations for PHP 8.2 release are completed, now we're waiting for the announcement (December, 8th).
I'm going to merge this PR after pushing the announcement.

If there are any improvements, we have 2 days 😃

TimWolla:
@saundefined Can you update the page at https://php.sergeypanteleev.com/releases/8.2/en.php with the very-latest version so one can take a look at the final rendered result?

saundefined:
> @saundefined Can you update the page at https://php.sergeypanteleev.com/releases/8.2/en.php with the very-latest version so one can take a look at the final rendered result?

Exactly, done, thanks!

releases/8.2/release.inc
```diff
@@ -0,0 +1,430 @@
...
+
+$blueprintRng = new Xoshiro256StarStar(
+    hash('sha256', "Example seed that is converted to a 256 Bit string via SHA-256", true)
+);
+
+$fibers = [];
+for ($i = 0; $i < 8; $i++) {
+	$fiberRng = clone $blueprintRng;
+	// Xoshiro256**'s 'jump()' method moves the blueprint ahead 2**128 steps, as if calling 'generate()' 2**128 times,
+	// giving the Fiber 2**128 unique values without needing to reseed.
```
TimWolla, line 231:
```suggestion
	// Xoshiro256**'s 'jump()' method moves the blueprint ahead 2**128 steps, as if calling
	// 'generate()' 2**128 times, giving the Fiber 2**128 unique values without needing to reseed.
```

Minor rewrapping to achieve consistent line lengths. Looks a little prettier.

releases/8.2/release.inc
```diff
@@ -0,0 +1,430 @@
...
+            </div>
+        </div>
+
+        <div class="php8-compare">
+            <h2 class="php8-h2" id="random_extension">
+                <?= message('random_title', $lang) ?>
+                <a class="php8-rfc" href="https://wiki.php.net/rfc/rng_extension">RFC</a>
+                <a class="php8-rfc" href="https://wiki.php.net/rfc/random_extension_improvement">RFC</a>
+                <a class="php8-rfc"
+                   href="https://www.php.net/manual/en/migration82.new-features.php#migration82.new-features.random"><?= message(
```
TimWolla, line 209: Not sure if it might be better to link to https://www.php.net/manual/en/book.random.php for the docs. The info given in the migration guide is very vague and I've tried to get as much docs written as possible for ext/random (unfortunately I wasn't able to completely finish it yet, but all the parameters / return values / classes are explained).

Alternatively perhaps the mentioning `\Random\Randomizer` in the description could be linked to https://www.php.net/manual/en/class.random-randomizer.php?

releases/8.2/languages/en.php
```diff
@@ -0,0 +1,40 @@
...
+    'constants_in_traits_title' => 'Constants in traits',
+    'constants_in_traits_description' => 'You cannot access the constant through the name of the trait, but, you can access the constant through the class that uses the trait.',
+    'deprecate_dynamic_properties_title' => 'Deprecate dynamic properties',
+    'deprecate_dynamic_properties_description' => '<p>The creation of dynamic properties is deprecated to help avoid mistakes and typos, unless the class opts in by using the <code>#[\AllowDynamicProperties]</code> attribute. <code>stdClass</code> allows dynamic properties.</p>
+        <p>Usage of the <code>__get</code>/<code>__set</code> magic methods is not affected by this change.</p>',
+    'new_classes_title' => 'New Classes, Interfaces, and Functions',
+    'new_mysqli' => 'New <code>mysqli_execute_query</code> function and <code>mysqli::execute_query</code> method.',
+    'new_attributes' => 'New <code>#[\AllowDynamicProperties]</code> and <code>#[\SensitiveParameter]</code> attributes.',
```
TimWolla, line 23: Similarly, perhaps the attributes could be linked. Both of them have documentation:

- https://www.php.net/manual/en/class.allow-dynamic-properties.php
- https://www.php.net/manual/en/class.sensitive-parameter

saundefined:
Thank you! Also added more links to documentation.

@TimWolla @cmb69 maybe merge this PR a bit earlier without header changes?
So we can preview and check it, what do you think?

cmb69:
> maybe merge this PR a bit earlier without header changes?

I'd also change the "Released!" to "To be released tomorrow!" or something like that, just in case someone finds the page.

index.php
```diff
@@ -95,7 +95,7 @@
     <img class="hero-logo" src="/images/logos/php-logo-white.svg" alt="php" width="240" height="120">
     <p class="hero-text">A <strong>popular general-purpose scripting language</strong> that is especially suited to web development.<br />Fast, flexible and pragmatic, PHP powers everything from your blog to the most popular websites in the world.</p>
     <div class="hero-actions">
-      <a href="/releases/8.2/index.php" class="hero-btn hero-btn-primary">What's new in 8.2</a>
+      <a href="/releases/8.1/index.php" class="hero-btn hero-btn-primary">What's new in 8.2</a>
```
TimWolla, line 98:
```suggestion
      <a href="/releases/8.1/index.php" class="hero-btn hero-btn-primary">What's new in 8.1</a>
```

saundefined: Yeah, I already fixed it, thanks!

saundefined: Thanks!

saundefined:
>I'd also change the "Released!" to "To be released tomorrow!" or something like that, just in case someone finds the page.

Good idea, thanks!
If there are no objections, I'll merge it a bit later today, and we can check it out.

Issue #676, opened 2022-07-28
Releases RSS/ATOM feed does not validate


AndrewGearhart:
The [PHP Releases RSS/ATOM feed does not pass validation in the W3C Feed Validator](https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fwww.php.net%2Freleases%2Ffeed.php). As a result of the failures, the feed cannot be added to several feed readers (including the Slack RSS feed reader).

cmb69:
Thanks for reporting! Could you please double-check that #678 solves this issue?

AndrewGearhart:
It _looks_ like it will. Can't say for sure without testing... but it looks like it will resolve it!

cmb69:
The fix should be rolled-out to the Website in a few hours. :)

AndrewGearhart:
To follow-up, this did fix it. Thanks again for being so speedy with this!

PR #678, opened 2022-07-29
Fix GH-676: Releases RSS/ATOM feed does not validate

cmb69:
The feed's updated element is required before the entries.

### DIFF
--- a/releases/feed.php
+++ b/releases/feed.php
@@ -18,6 +18,8 @@ echo <<<XML
 
 XML;
 
+ob_start();
+
 // Flatten major versions out of RELEASES.
 $RELEASED_VERSIONS = array_reduce($RELEASES, 'array_merge', []);
 $FEED_UPDATED =  0;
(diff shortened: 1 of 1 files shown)


releases/feed.php
```diff
@@ -80,9 +82,10 @@
...
 
+$entries = ob_get_clean();
+
 $FEED_UPDATED = date(DATE_ATOM, max($FEED_UPDATED));
 
-echo <<<XML
-    <updated>{$FEED_UPDATED}</updated>
-</feed>
-XML;
+echo "    <updated>{$FEED_UPDATED}</updated>\n";
```
kamil-tekiela, line 89: Shouldn't there be a newline before this output?

cmb69: This file uses trailing newlines troughout; the respective linebreak is already there:

https://github.com/php/web-php/blob/c46c30f53b421606849a82047f53af9c5ca01d35/releases/feed.php#L18

kamil-tekiela (approved):
LGTM

Issue #682, opened 2022-08-04
Improve the thanks.php page


rlerdorf:
Could someone take a look at the thanks.php page and make it look a bit more modern?

Perhaps include logos for the current providers and a slightly longer blurb on each one if we have more we can say.

It would help a bit, especially with Digital Ocean, that have been asking for more visibility. For Digital Ocean, we can add that we have edit.php.net, bugs.php.net, museum.php.net, wiki.php.net, downloads.php.net, qa.php.net, pecl.php.net on Digital Ocean droplets.

cmb69:
I'm afraid that only few (if any) people know which of these providers are still active, i.e. which services we still use. It might be best if the providers would file PRs about what they like to change [there](https://github.com/php/web-php/blob/master/thanks.php).

rlerdorf:
The providers usually have no idea what we are running on the servers they provide, so it is not going to come from them.

mvriel:
I just did a very mock -or quick design- as to how the -current- sponsors section could look like; perhaps divided into 2 columns as the number of sponsors is significant enough and the screen real estate allows for it.

Would something like this be interesting?

![image](https://user-images.githubusercontent.com/193704/183056847-7fc29f70-b253-48b0-95fc-9ca7ce69de4a.png)

As for the legacy sponsors, it may be interesting to only show their logo and perhaps a caption with their support method as this is secondary information and to focus on current sponsors?

rlerdorf:
Great, I like it. We don't need to do much for the legacy sponsors at this point. We can leave those are one-liners without a logo for now, although I suppose that might make the page look unbalanced.

mvriel:
Unless someone else beats me to it, I can find some time during the weekend to make a PR for this

mvriel:
During implementation, I discovered that when using 2 columns, some of the texts would become 3 lines high and make the mock above look messy. This is a screenshot of the current implementation
![image](https://user-images.githubusercontent.com/193704/183212047-2fd84e01-004b-491b-ba65-1efff1502771.png)

There are one or two alignment issues to iron out, and I'll be issuing a PR of this without changing the 'Thanks Emiritus' section. I can attack that issue in a subsequent PR to keep things small and incremental

derickr:
This seems to have been merged, and hence this issue can be closed?

PR #684, opened 2022-08-06
Modernize top half of Thanks page

mvriel:
In issue #682 it was requested to modernize the thanks page; as a first
step in this effort I have given the top half of the page with current
sponsors a brush up, including responsive behaviour.

Unfortunately I could not find a logo for ServerCentral; so I left that
empty.

The improvements in this PR does use CSS Grids, meaning it won't 
show up exactly the same way in IE11 compared to other browsers. 
I have taken this as a given and unfortunate thing, but I do not know
whether this is an issue as I do not know the browserlist.

If the above is an issue, let me know and I'll invest more time on how 
it looks on IE11.


mvriel:
Desktop

image

Tablet (Portrait)
image

Mobile
image

As for mobile, I am a bit on the fence if the cards aren't too big. Feedback is appreciated, to show it nicely with logo's is a bit of a thing on mobile if you want to keep it small :)

heiglandreas:
I wouldn't worry too much about the size on mobile.

I'm more asking myself whether there is a possibility to randomize an unordered list via CSS. So that not always the same sponsors are "above the fold". Or at the top of the list on mobile.

cmb69:
This looks great! Thank you!

> If the above is an issue, let me know and I'll invest more time on how
> it looks on IE11.

I wouldn't really worry about IE11; I doubt many PHP developers are using that to browse php.net, and even if someone uses IE11, they still get the info.

> I'm more asking myself whether there is a possibility to randomize an unordered list via CSS.

FWIW, that could be done on the server.

mvriel:
> > I'm more asking myself whether there is a possibility to randomize an unordered list via CSS.
> 
> FWIW, that could be done on the server.

Would be a nice followup contribution to keep scope for this PR limited :)

RobiNN1:
> Unfortunately I could not find a logo for ServerCentral; so I left that
empty.

They use it in the favicon so I assume this is their logo.

https://www.servercentral.net/assets/images/mark.svg

mark

I randomly found this PR, and it looks better than current one.

samdark:
👍

saundefined:
Hi all, I resolved the conflict and added the missing logo.

If there are no objections, I'll merge this PR,
if there are comments — I'll try to quickly fix :)

cmb69 (approved):
Thank you! Looks great!

PR #689, opened 2022-08-30
Use CSS variables for some common colors

morrisonlevi:
Aside from being able to tell what color it is or what purpose it
solves just by looking at the variable name, it helps provide a
foundation for more easily changing the theme in the future.

Hopefully, we'll have a light and dark theme one day :fingerscrossed:

Mostly, colors are unchanged. However, I did change some colors which
were close to the theme colors but not quite.

### DIFF
--- a/styles/theme-base.css
+++ b/styles/theme-base.css
@@ -9,6 +9,14 @@
  * Designed and built with all the love in the world @twitter by @mdo and @fat.
  */
 
+:root {
+  --dark-grey-color: #333;
+  --dark-blue-color: #4F5B93;
+  --medium-blue-color: #7A86B8;
+  --light-blue-color: #E2E4EF;
+  --magenta-color: #793862;
+}
+
 .clearfix {
   *zoom: 1;
 }
@@ -60,7 +68,7 @@ a {
   border-bottom:1px solid;
 }
 a:focus {
-  outline: thin dotted #333;
+  outline: thin dotted var(--dark-grey-color);
   outline-offset: -2px;
 }
 a:hover,
@@ -865,7 +873,7 @@ table.standard th.subr {
 
 #layout .cse .gsc-result-info,
 #layout .gsc-result-info {
-    color: #333;
+    color: var(--dark-grey-color);
     font-size: 0.75rem;
     padding: 0;
     margin: 0;
@@ -1006,11 +1014,11 @@ fieldset {
 }
 
 .tt-dropdown-menu {
-    background: none repeat scroll 0 0 #E2E4EF;
+    background: none repeat scroll 0 0 var(--light-blue-color);
     border-bottom: 1px solid #C4C9DF;
     border-radius: 0 0 2px 2px;
-    box-shadow: 1px 0 1px -1px #C4C9DF inset, -1px 0 1px -1px #C4C9DF inset, 0 0 1px #4F5B93;
-    color: #333333;
+    box-shadow: 1px 0 1px -1px #C4C9DF inset, -1px 0 1px -1px #C4C9DF inset, 0 0 1px var(--dark-blue);
+    color: var(--dark-grey-color);
     padding-top: 3px;
     margin-top: -3px;
     min-width: 100%;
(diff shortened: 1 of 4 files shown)


styles/theme-base.css
```diff
@@ -1006,20 +1014,20 @@ fieldset {
 }
 
 .tt-dropdown-menu {
-    background: none repeat scroll 0 0 #E2E4EF;
+    background: none repeat scroll 0 0 var(--light-blue-color);
     border-bottom: 1px solid #C4C9DF;
     border-radius: 0 0 2px 2px;
-    box-shadow: 1px 0 1px -1px #C4C9DF inset, -1px 0 1px -1px #C4C9DF inset, 0 0 1px #4F5B93;
-    color: #333333;
+    box-shadow: 1px 0 1px -1px #C4C9DF inset, -1px 0 1px -1px #C4C9DF inset, 0 0 1px var(--dark-blue);
```
salathe, line 1020:
```suggestion
    box-shadow: 1px 0 1px -1px #C4C9DF inset, -1px 0 1px -1px #C4C9DF inset, 0 0 1px var(--dark-blue-color);
```

Note: this error is repeated a bunch more times.

styles/theme-medium.css
```diff
@@ -1,36 +1,29 @@
...
+  background-color: var(--background-color);
+  background-image: url('/images/bg-texture-00.svg');
   color: #CCC;
 }
+
 #layout-content {
-  background:#F2F2F2;
-  color:#333;
+  background: var(--foreground-color);
+  color: var(--background-color);
```
salathe, line 14: Using the "foreground color" for the background, and "background color" for the foreground (text) isn't confusing at all... 😕

salathe (commented):
A step in the right direction.  👍

styles/theme-medium.css
morrisonlevi: Yes, this is an issue with the whole concept of "theme-medium" which is neither light nor dark. I can probably come up with better variable names though.

styles/theme-base.css
salathe: Fixed in 870bf66a.

styles/theme-medium.css
morrisonlevi: Take another look, I renamed some things hopefully for the better.

salathe: Thanks, I think those are an improvement.  Naming things is hard! 😏

salathe (approved):
There are a bunch of usages of a slightly different magenta (`#AE508D`) than `--magenta-color` (`#793862`), which could probably either be replaced with `--magenta-color` (my preference, I think) or _maybe_ a new variable.   Discovered by just changing `--magenta-color` to something completely different, and still seeing magenta on various elements.

Other than that, as I said earlier, I think this is a great first step.  For whatever it is worth, I approve. 👍

morrisonlevi:
@salathe What setup are you using to test these changes, btw? In the past (as in, 10 years ago) I had some way to rsync files while not clobbering anything that was in git, and I don't remember how I did that. Wondering if you have any tips for me :)

morrisonlevi:
> There are a bunch of usages of a slightly different magenta (`#AE508D`) than `--magenta-color` (`#793862`), which could probably either be replaced with `--magenta-color` (my preference, I think) or _maybe_ a new variable. Discovered by just changing `--magenta-color` to something completely different, and still seeing magenta on various elements.

I decided to add a new variable `--medium-magenta-color` and rename the previous one `--dark-magenta-color`. These are still colors rather than a light/dark theme agnostic type of name, but that's a bit too far in the future for now. Also, the `--dark-magenta-color` seems too dark in places where it gets used on the dark  background (like the sidebar links on the home page).

Thanks again!

salathe:
> @salathe What setup are you using to test these changes, btw?

I just checked out web-php and ran the built-in web server (e.g. `php -S 127.0.0.1:8080 .router.php`). 

> I decided to add a new variable `--medium-magenta-color` and rename the previous one `--dark-magenta-color`.

Cool, fine by me. 👍

PR #690, opened 2022-08-30
Update GPG key for `carusogabriel`

carusogabriel:
Here's the output of `gpg --fingerprint` for the new key:

```
pub   rsa4096 2022-08-30 [SC] [expires: 2024-08-29]
      2C16 C765 DBE5 4A08 8130  F1BC 4B9B 5F60 0B55 F3B4
uid           [ultimate] Gabriel Caruso <carusogabriel@php.net>
sub   rsa4096 2022-08-30 [E] [expires: 2024-08-29]
```

This affects https://php.net/gpg-keys.php#gpg-8.0 & https://php.net/downloads.php#gpg-8.0


cmb69:
I think you need to *add* the new info, not to *replace* the old one.

carusogabriel:
@cmb69 Thanks for the review - I'm finishing cloning https://github.com/php/web-php-distributions, I'll follow up with a PR in there as well updating the keyring.

PR #695, opened 2022-09-06
Client side validation of form-user

cmb69:
That is supposed to be an email address, and the client side validation
will at least help legitimate users (and maybe reduce some nonsense).

### DIFF
--- a/manual/add-note.php
+++ b/manual/add-note.php
@@ -367,7 +367,7 @@ else {?>
   </tr>
   <tr>
    <th class="subr"><label for="form-user">Your email address (or name)</label>:</th>
-   <td><input id="form-user" type="text" name="user" size="60" maxlength="40" required value="<?php echo clean($_POST['user']); ?>"></td>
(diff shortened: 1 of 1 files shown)


Girgias (approved):

cmb69:
Ah, good catch, @salathe! The point is that a user can enter a name only, but the notification mails then show `USERNAME@mail.php.net` as sender (e.g. https://news-web.php.net/php.notes/219540). This is confusing for readers, and it makes it impossible to reject user notes (so they need to be deleted). If we can change the behavior of the mail server, I'm okay-ish with asking for username or email address (although that still would cause deletions instead of rejections).

localheinz (commented):
👍

manual/add-note.php
```diff
@@ -367,7 +367,7 @@
   </tr>
   <tr>
    <th class="subr"><label for="form-user">Your email address (or name)</label>:</th>
```
saundefined, line 369:
```suggestion
   <th class="subr"><label for="form-user">Your email address</label>:</th>
```

I suppose we could additionally do that?

PR #697, opened 2022-09-11
Enhancement: Enable `new_with_braces` fixer

localheinz:
This pull request

- [x] enables the `new_with_braces` fixer

Replaces #540.
Follows #559.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/doc/rules/operator/new_with_braces.rst.

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -29,6 +29,7 @@ $config
         'constant_case' => true,
         'indentation_type' => true,
         'line_ending' => true,
+        'new_with_braces' => true,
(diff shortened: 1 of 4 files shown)


cmb69:
Thank you for the PR!

This is fine for me, but @kamil-tekiela was against that in #540, and maybe others don't like this either. So let's wait for more feedback.

morrisonlevi:
I'm also fine with this. Not sure how much feedback we'll get ^_^

PR #699, opened 2022-09-11
Enhancement: Enable `random_api_migration` fixer

localheinz:
This pull request

- [x] enables the `random_api_migration` fixer

Replaces #545.
Follows #559.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.4.0/doc/rules/alias/random_api_migration.rst.

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -32,6 +32,7 @@ $config
         'no_extra_blank_lines' => true,
         'no_trailing_whitespace' => true,
         'ordered_class_elements' => true,
+        'random_api_migration' => true,
         'single_space_after_construct' => true,
         'strict_param' => true,
         'switch_case_space' => true,
--- a/download-logos.php
+++ b/download-logos.php
@@ -22,7 +22,7 @@ function print_star(): void
 function random_bgcolor($min, $max): void
 {
     echo "style=\"background-color: #" .
-         sprintf('%02x%02x%02x', rand($min, $max) * 51, rand($min, $max) * 51, rand($min, $max) * 51) .
+         sprintf('%02x%02x%02x', mt_rand($min, $max) * 51, mt_rand($min, $max) * 51, mt_rand($min, $max) * 51) .
          ";\"";
 }
 ?>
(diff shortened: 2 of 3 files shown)


cmb69:
Hmm, replacing `rand()` with `mt_rand()` is okay (to use the canonical function, not the alias), but wouldn't all these occurrences profit from using `random_int()` instead?

localheinz:
@cmb69 

Looks like it is not possible to seed the generator, see https://stackoverflow.com/q/39352599/1172545.

In other words, the tests would be failing.

We could, however, extract (and thus decouple) a `RandomNumberGenerator` or similar with corresponding implementations. What do you think?

TimWolla:
> Looks like it is not possible to seed the generator when using `ramdon_int()`

Yes, this is by design. 

> We could, however, extract (and thus decouple) a `RandomNumberGenerator` or similar with corresponding implementations. What do you think?

Likely not worth it. Use `random_int()` wherever reasonably possible and then switch to [`\Random\Randomizer`](https://wiki.php.net/rfc/rng_extension) once PHP 8.2 is available to use.

localheinz:
@TimWolla 

> Likely not worth it. Use random_int() wherever reasonably possible and then switch to [\Random\Randomizer](https://wiki.php.net/rfc/rng_extension) once PHP 8.2 is available to use.

Sounds great, but the production system currently runs on PHP 7.3 (maybe), and apparently nobody knows who is responsible for it nor how it could be upgraded to use a current PHP version.

TimWolla:
Yes, that's why I put the "once PHP 8.2 is available to use" there. This was not referring to the gold release in November, but to whenever the requirements for web-php are increased.

Adding a wrapper for the pre-PHP-8.2 RNGs is IMO pure busy work, using the CSPRNG would be great, but if it's painful to do, then it's not worth it just for the anti-spam challenges.

morrisonlevi:
IMO, we can merge this as-is. Is there something _better_? Maybe, but this is a clear improvement that we can merge right this moment, yes?

cmb69:
Oh, indeed, I forgot about the tests. So I agree with @morrisonlevi that this is a step forward. Thank you all!

localheinz:
Thank you, @cmb69, @morrisonlevi, and @TimWolla!

PR #702, opened 2022-09-17
Update footer.inc

rongzedong:
manual View Source is a 404 error page.

### DIFF
--- a/include/footer.inc
+++ b/include/footer.inc
@@ -68,7 +68,7 @@
         <li><a href="/contact.php">Contact</a></li>
         <li><a href="/sites.php">Other PHP.net sites</a></li>
         <li><a href="/privacy.php">Privacy policy</a></li>
-<?php if (!empty($_SERVER['BASE_PAGE'])): ?>
+<?php if (!empty($_SERVER['BASE_PAGE']) || substr($_SERVER['BASE_PAGE'],0,5)=='manual'): ?>
         <li><a href="https://github.com/php/web-php/blob/master/<?php echo urlencode($_SERVER['BASE_PAGE']); ?>">View Source</a></li>
 <?php endif; ?>
       </ul>
--- a/include/footer.inc
+++ b/include/footer.inc
@@ -68,7 +68,7 @@
         <li><a href="/contact.php">Contact</a></li>
         <li><a href="/sites.php">Other PHP.net sites</a></li>
         <li><a href="/privacy.php">Privacy policy</a></li>
-<?php if (!empty($_SERVER['BASE_PAGE']) && strncmp($_SERVER['BASE_PAGE'], 'manual', 5) !== 0): ?>
+<?php if (!empty($_SERVER['BASE_PAGE']) && strncmp($_SERVER['BASE_PAGE'], 'manual', 6) !== 0): ?>
         <li><a href="https://github.com/php/web-php/blob/master/<?php echo urlencode($_SERVER['BASE_PAGE']); ?>">View Source</a></li>
 <?php endif; ?>
       </ul>
--- a/include/footer.inc
+++ b/include/footer.inc
@@ -68,7 +68,7 @@
         <li><a href="/contact.php">Contact</a></li>
         <li><a href="/sites.php">Other PHP.net sites</a></li>
         <li><a href="/privacy.php">Privacy policy</a></li>
-<?php if (!empty($_SERVER['BASE_PAGE']) && substr($_SERVER['BASE_PAGE'],0,5)!='manual'): ?>
+<?php if (!empty($_SERVER['BASE_PAGE']) && strncmp($_SERVER['BASE_PAGE'], 'manual', 5) !== 0): ?>
         <li><a href="https://github.com/php/web-php/blob/master/<?php echo urlencode($_SERVER['BASE_PAGE']); ?>">View Source</a></li>
 <?php endif; ?>
       </ul>
--- a/include/footer.inc
+++ b/include/footer.inc
@@ -68,9 +68,9 @@
         <li><a href="/contact.php">Contact</a></li>
         <li><a href="/sites.php">Other PHP.net sites</a></li>
         <li><a href="/privacy.php">Privacy policy</a></li>
-<?php if (empty($_SERVER['BASE_PAGE']) || substr($_SERVER['BASE_PAGE'],0,5)=='manual') {} else {?>
+<?php if (!empty($_SERVER['BASE_PAGE']) && substr($_SERVER['BASE_PAGE'],0,5)!='manual'): ?>
         <li><a href="https://github.com/php/web-php/blob/master/<?php echo urlencode($_SERVER['BASE_PAGE']); ?>">View Source</a></li>
-<?php } ?>
+<?php endif; ?>
       </ul>
       </div>
     </div>
--- a/include/footer.inc
+++ b/include/footer.inc
@@ -68,9 +68,9 @@
         <li><a href="/contact.php">Contact</a></li>
         <li><a href="/sites.php">Other PHP.net sites</a></li>
         <li><a href="/privacy.php">Privacy policy</a></li>
-<?php if (!empty($_SERVER['BASE_PAGE']) || substr($_SERVER['BASE_PAGE'],0,5)=='manual'): ?>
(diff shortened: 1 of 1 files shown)


rongzedong:
link to the manual XML file, will be a Better idea.

kamil-tekiela:
I'm not sure how this fixes the problem. Could you explain?

rongzedong:
visit this url&nbsp;https://www.php.net/manual/en/function.date.php and click view source. will be 404
Use this patch will close this function on the manual page.



---Original---
From: "Kamil ***@***.***&gt;
Date: Sun, Sep 18, 2022 22:02 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [php/web-php] Update footer.inc (PR #702)




 
I'm not sure how this fixes the problem. Could you explain?

kamil-tekiela:
Did you by any chance mean to use `&&` instead of `||`. I think that would make more sense.

rongzedong:
no, i think i'm right. did you see the " ? : "
that mean not set base_url or find manual&nbsp; in base url, not show view source, other will show view source link.



---Original---
From: "Kamil ***@***.***&gt;
Date: Tue, Sep 20, 2022 23:30 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [php/web-php] Update footer.inc (PR #702)




 
Did you by any chance mean to use &amp;&amp; instead of ||. I think that would make more sense.

kamil-tekiela:
Do you mean you inserted a ternary operator there somewhere? I cannot see it in this PR.

cmb69:
We should either remove that link completely, or just use the same URL as for the "Submit a pull request" link.

rongzedong:
it's exist right there.
i just add ' || ......"manual" '







---Original---
From: "Kamil ***@***.***&gt;
Date: Tue, Sep 20, 2022 23:41 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [php/web-php] Update footer.inc (PR #702)




 
Do you mean you inserted a ternary operator there somewhere? I cannot see it in this PR.

kamil-tekiela:
Well, then someone must be wrong; either you or me. @cmb69 Can you see a ternary operator in the code? 
Perhaps, you saw the `:` starting the `if` block and you confused it with a ternary operator... but that still would not explain the flaw in logic.

cmb69:
The patch as is makes no sense; if the 5 first bytes equal "manual", the variable is not empty.

rongzedong:
The original meaning is that if there is not setting baseurl, the view source will not be displayed , set baseurl the view source will be displayed so that you can view the source code. The part I added is that this function is in the manual part, the view source code 404, so I turn off when viewing the manual. If you have a better plan, of course, you can also propose changes and improvements according to your ideas.




---Original---
From: "Kamil ***@***.***&gt;
Date: Tue, Sep 20, 2022 23:42 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [php/web-php] Update footer.inc (PR #702)




 
Do you mean you inserted a ternary operator there somewhere? I cannot see it in this PR.

rongzedong:
I want this expression to mean that if base_url is empty, the view will not be displayed, or if it starts with "manual", it will not be displayed too. i think you can test that part&nbsp; localhost.




---Original---
From: "Christoph M. ***@***.***&gt;
Date: Tue, Sep 20, 2022 23:58 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [php/web-php] Update footer.inc (PR #702)




 
The patch as is makes no sense; if the 5 first bytes equal "manual", the variable is not empty.

rongzedong:
sorry i think i push a error patch.
&nbsp;



---Original---
From: "Christoph M. ***@***.***&gt;
Date: Tue, Sep 20, 2022 23:58 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [php/web-php] Update footer.inc (PR #702)




 
The patch as is makes no sense; if the 5 first bytes equal "manual", the variable is not empty.

rongzedong:
i update the code for my idea.

rongzedong:
@cmb69 @mj @kamil-tekiela , i change the code, please help me,  review and approval. think you.

kamil-tekiela:
Thanks. The code makes more sense now, but still isn't something that we would merge. However, I think cmb69 might be right that the solution you are proposing isn't the ideal solution.

include/footer.inc
```diff
@@ -68,9 +68,9 @@
         <li><a href="/contact.php">Contact</a></li>
         <li><a href="/sites.php">Other PHP.net sites</a></li>
         <li><a href="/privacy.php">Privacy policy</a></li>
-<?php if (!empty($_SERVER['BASE_PAGE'])): ?>
+<?php if (empty($_SERVER['BASE_PAGE']) || substr($_SERVER['BASE_PAGE'],0,5)=='manual') {} else {?>
```
cmb69, line 71: The if-condition should be inverted, so that we don't the else-branch. A first step would be:
```suggestion
<?php if (!(empty($_SERVER['BASE_PAGE']) || substr($_SERVER['BASE_PAGE'],0,5)=='manual')): ?>
```
But we want to go a step further (applying [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)):
```suggestion
<?php if (!empty($_SERVER['BASE_PAGE']) && substr($_SERVER['BASE_PAGE'],0,5)!='manual'): ?>
```
Finally, I suggest to get rid of the `substr()` and use `strncmp()` instead (unfortunately, we can't use `str_starts_with()` for now):
```suggestion
<?php if (!empty($_SERVER['BASE_PAGE']) && strncmp($_SERVER['BASE_PAGE'], 'manual', 5) !== 0): ?>
```

cmb69:
> However, I think cmb69 might be right that the solution you are proposing isn't the ideal solution.

I have overlooked that "view source" as is makes sense for those who want to work on web-php (not doc-en). As such, @rongzedong's suggestion makes sense now, although that code should be refactored.

kamil-tekiela:
We could have an `else` condition too. It would link to doc-en.

rongzedong:
yes. I also think it is a better solution if I can access the corresponding version of doc xml source, my solution is just a transitional solution to solve the 404 problem.



---Original---
From: "Kamil ***@***.***&gt;
Date: Wed, Sep 21, 2022 21:05 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [php/web-php] Update footer.inc (PR #702)




 
We could have an else condition too. It would link to doc-en.

include/footer.inc
```diff
@@ -68,7 +68,7 @@
         <li><a href="/contact.php">Contact</a></li>
         <li><a href="/sites.php">Other PHP.net sites</a></li>
         <li><a href="/privacy.php">Privacy policy</a></li>
-<?php if (!empty($_SERVER['BASE_PAGE'])): ?>
+<?php if (!empty($_SERVER['BASE_PAGE']) && strncmp($_SERVER['BASE_PAGE'], 'manual', 5) !== 0): ?>
```
kamil-tekiela, line 71: should it not be 6?

rongzedong:
@cmb69 i find this,:
BENCHMARK ITERATION RESULT IS:
if (substr($key,0,5 == "HTTP_").... -   0,000481s
if (!strncmp($key, 'HTTP_', 5)).... -     0,000405s

strncmp() is 20% faster than substr() :D

cmb69:
> strncmp() is 20% faster than substr() :D

Yes, I know (`substr()` constructs a temporary string, what is not needed here). Still, that would only be a micro-optimization. The main reason is that we could easily grep for `strncmp` when we want to replace it with `str_starts_with()` in the future.

include/footer.inc
cmb69: Maybe use `strlen('manual')` instead.

cmb69 (approved):
Thank you! This is fine to be merged.

cmb69:
Looks like this has been overlooked. I've merged it now. Thank you!

PR #703, opened 2022-09-19
Also redirect function.curl-file-create.php to curlfile.construct

kocsismate:
In order to fix https://github.com/php/doc-en/issues/1836


cmb69 (commented):
I no longer think this is the right approach. Besides that there are other URLs which lead to the page (e.g. php.net/curl_file_create), the search no longer offers `curl_file_create()` what is probably the biggest issue. Instead, we could properly document `curl_file_create()` on its own page, but not as alias of `CurlFile::__construct()`, but rather with its own documentation (probably referring to `CurlFile::__construct()` where appropriate).

kocsismate:
OK? you are right!

PR #704, opened 2022-09-21
Remove links to previously removed guides

CdrMarks:
The [Unsupported Branches page](https://www.php.net/eol.php) should not link to migration guides that have been previously removed.


kamil-tekiela (approved):
Thank you. LGTM

CdrMarks:
PR updated to use Zend's PHP Legacy docs.

kamil-tekiela (approved):

PR #705, opened 2022-09-22
Update common.php

rmanalu15:
Add Indonesian language to release page

PR #707, opened 2022-09-22
Create id.php to language

rmanalu15:
Add Indonesian language to release page

### DIFF
--- a/releases/8.1/languages/id.php
+++ b/releases/8.1/languages/id.php
@@ -1,0 +1,90 @@
+<?php
+
+return [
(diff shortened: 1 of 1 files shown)


cmb69:
This also requires #706.

I'm not sure that it is a good idea to add translations for languages for which we don't have translations of the manual proper (for maintenance reasons).

@saundefined, @pronskiy, what do you think?

saundefined:
Indeed,

Also, this page contains links to document pages with language-specific URLs,
If we merge it, all links to doc pages will look like 404

pronskiy:
+1, I think it's better to stick to the maintained languages.

cmb69:
> I think it's better to stick to the maintained languages

Okay, so I'm closing this and the related #705 and #706.

Nonetheless, thank you for your constribution, @rmanalu15!

PR #708, opened 2022-09-28
Remove unnecessary code

kamil-tekiela:


include/layout.inc
```diff
@@ -486,10 +486,6 @@ function site_header($title = '', $config = [])
         $config["css_overwrite"] = ["/styles/i-love-markdown.css"];
     }
 
-    if (empty($title)) {
-        $title = "Hypertext Preprocessor";
-    }
```
cmb69, line 491: This is fine, since `site_header()` is apparently never called with an empty string.

cmb69 (approved):
Thank you!

PR #709, opened 2022-09-28
Improve code for language selection

kamil-tekiela:
It just trims down unnecessary code bloat.


cmb69 (approved):
Indeed, looks like the second argument is never passed.

Thank you!

PR #721, opened 2022-11-16
Use French instead of Hungarian

sy-records:
No more Hungarian language in the documentation


saundefined (approved):
LGTM, thanks!

cmb69 (approved):
Good catch, thank you!

PR #725, opened 2022-11-30
Add realeases/next.php endpoint

renanbr:
This Pull Request is actually a feature request.

The goal is to expose an endpoint that users can use to fetch the upcoming release.

The `realeases/next.php` returns a JSON response with the next version, for example:

```json
"8.2"
```

I don't mind if another approach is proposed instead.

### DIFF
--- a/releases/next.php
+++ b/releases/next.php
@@ -1,0 +1,24 @@
+<?php
+$_SERVER['BASE_PAGE'] = 'releases/next.php';
+
+include_once __DIR__ . '/../include/prepend.inc';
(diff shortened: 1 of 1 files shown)


cmb69:
> The goal is to expose an endpoint that users can use to fetch the upcoming release.

Hmm, I don't understand the reason for that. https://www.php.net/releases/active.php already lists all current GA releases, while https://qa.php.net/api.php?type=qa-releases&format=json lists all current QA releases.

renanbr:
Thank you @cmb69 , didn't know the second link. This one solves my problem 🥂

PR #730, opened 2022-12-08
Fix news2html truncation

cmb69:
If there are NEWS entries referring to a certain PHP version, these are interpreted as belonging to that PHP version.  Since the lines actually delimiting PHP versions are supposed to have comma after the date, we enforce that now.


cmb69:
cc @saundefined, @derickr

saundefined (approved):
Looks good, just tested locally, changelog generated fine

Thanks!

sy-records (approved):

PR #731, opened 2022-12-08
Simplify DNF example

cmb69:
As has been pointed out by Thomas Gutbier[1], the PHP 8.2 code doesn't need the special handling for `null`, so we remove it, and accordingly merge the special case handling in the pre PHP 8.2 snippet as well.

[1] <https://externals.io/message/119096#119097>


saundefined (approved):
Thanks!

PR #732, opened 2022-12-08
Fix trait constant example

cmb69:
As has been pointed out by Robert Landers[1], the PHP 8.2 example does not error.  And even the pre PHP 8.2 example does not error where indicated.  It does not appear to be useful to show the pre PHP 8.2 behavior (where it was not allowed to declare constants in traits); so instead we show only a PHP 8.2 example which matches the description.

[1] <https://externals.io/message/119096#119100>


saundefined (approved):
Thanks!

PR #733, opened 2022-12-08
Simplify stand-alone type example

nikic:
These string types confused me -- seems like we can drop them?


sy-records (approved):
LGTM.

saundefined:
Thanks!

PR #739, opened 2022-12-13
Fixed Japanese translation.

issakujitsuk:
The phrase "is not affected by this change" should be translated as "この変更の影響を受けません".

I would also add that translating "Usage of the __get/__set magic methods" as "マジックメソッド __get/__set を使っても" seems a bit unnatural in Japanese. Wouldn't it be more natural to write "マジックメソッド __get/__set を使う場合は"?


mumumu (approved):
looks good to me. @saundefined

saundefined:
Thanks!

PR #747, opened 2023-01-09
Fix redirection rules and add some new rules

kocsismate:


error.php
```diff
@@ -291,13 +291,15 @@
...
+    'splstack.setiteratormode' => 'spldoublylinkedlist.setiteratormode',
+    'splqueue.setiteratormode' => 'spldoublylinkedlist.setiteratormode',
+    'class.allow-dynamic-properties' => 'class.allowdynamicproperties',
+    'class.return-type-will-change' => 'class.returntypewillchange',
+    'class.sensitive-parameter' => 'class.sensitiveparameter',
     'function.curl-file-create' => 'curlfile.construct',
 ];
 
-if (preg_match("!^manual/([^/]+)/([^/]+)$!", $URI, $match) && isset($manual_redirections[$match[2]])) {
+if (preg_match("!^manual/([^/]+)/([^/]+)(\.php)?$!", $URI, $match) && isset($manual_redirections[$match[2]])) {
```
cmb69, line 302: I don't think that can work, since the second group would capture the trailing `.php`. A possible solution would be to make the second group ungreedy:
```suggestion
if (preg_match("!^manual/([^/]+)/([^/]+?)(\.php)?$!", $URI, $match) && isset($manual_redirections[$match[2]])) {
```
It might also make sense to use non-capturing groups (i.e. `(?:…)` instead of `(…)`) where possible.

cmb69 (commented):
Thank you! Handling the `.php` suffix generically is a very good idea.

error.php
kocsismate: Ah right. I was testing the regex on regex101.com, and didn't notice that the 3rd group was not captured. However, neither enabling the ungreedy mode evaluates to the desired result, as the 2nd group only captures a single character. So I thought it's simpler to simply remove the last `.php` suffix from the URI before using it in the regex.

kocsismate:
> I think we can leave this as is, although the ungreedy regex would work as well (([^/]+?) catches as many characters as possible).

Not sure what I messed up last time, but now it really works. I also made the 3rd group non-capturing.

PR #752, opened 2023-01-28
Add woff2 fonts and other font related tweaks

javiereguiluz:
This Pull Request proposes 3 changes related to fonts:

**(1)** Add the WOFF2 variant of web fonts. [WOFF2 standard](https://w3c.github.io/woff/woff2/) reduces the file sizes significantly. This is important to make our website more accessible to folks with limited Internet connections.

96% of browsers support WOFF2 ([source](https://caniuse.com/?search=woff2)) and we still have WOFF and TTF/OTF/EOT formats for those edge-cases which don't support it.

Real examples of the file size reductions for php.net:

```
Fira Sans Regular: 183 KB (woff) -> 129 KB (woff2) --> 30% reduction in size
Fira Mono Regular:  87 KB (woff) ->  65 KB (woff2) --> 35% reduction in size
```

**(2)** Use `font-display: swap` for web fonts (see [MDN Docs for font-display](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display))

This is recommended to not block the rendering the page for a long time if the web font is not available. Again, this is important for folks with suboptimal network connections.

The `swap` value is also the default option in popular font services like [Google Fonts](https://fonts.google.com/).

**(3)** Update the list of alternative font families.

I updated the sans-serif and mono font stack used in the design. The new stacks are much longer (not a problem in practice thanks to CSS variables) but they cover all common browsing devices (including smartphones). I crafted the list from the recommended font stacks of the most popular CSS frameworks and the most popular websites for developers.


javiereguiluz:
If you have some time, please review this proposal. It would improve performance significantly and, to my understanding, there are no drawbacks. Thanks a lot!

oxou:
Any updates on this?

Many people (myself included) browse the manual page on a daily basis and have a slow internet connection, I sometimes wait 2-3 seconds for fonts to load.

javiereguiluz:
I've just rebased this to fix the merge conflicts. Thanks.

Girgias:
Seems reasonable to me, thank you!

PR #753, opened 2023-01-29
Font display swap fix

oxou:
What `font-family: swap` does is `Fallback text is immediately rendered in the next available system typeface in the font stack until the custom font loads, in which case the new typeface will be swap ped in`

This fixes the issue on slow connections where fonts on the page are invisible while the remote font is still downloading, this provides a fallback to the system fonts until the remote font has been downloaded.

More information about it here: 
- https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display
- https://developer.chrome.com/docs/lighthouse/performance/font-display/

### DIFF
--- a/fonts/Fira/fira.css
+++ b/fonts/Fira/fira.css
@@ -8,6 +8,7 @@
          url('/fonts/Fira/ttf/FiraSans-Book.ttf') format('truetype');
     font-weight: 400;
     font-style: normal;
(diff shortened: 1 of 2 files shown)


javiereguiluz:
This is being fixed in #752, so I'd say this is a duplicate of the other pull request. Thanks.

PR #758, opened 2023-02-15
Support serving PHP distribution packages from GitHub releases

ramsey:
https://github.com/php/php-src/pull/10604 introduces a new automated release workflow for building PHP release packages. With this new workflow, we will no longer serve PHP distributions packages from `web-php-distributions`, but rather, from GitHub releases. This PR adds support to `web-php` to create the proper download URLs for the files in GitHub.

Important things to note in this PR:

- All current functionality works and behaves as expected (i.e., museum, links to files in `/distributions/`, etc.)
- The main change is this: if a release contains `'location' => 'github'`, we'll use download links to the files in GitHub
  - Here's an example of what this will look like: https://github.com/ramsey/web-php/commit/776d694f6c5a97dc48e4ea036dc15fc993ffbdec
  - Here's an example of how we'll update `web-qa` with this new process: https://github.com/ramsey/web-qa/commit/b28dfaee92ec9a5f4a2663d46980e79481f6c943
- I've moved `php-keyring.gpg` from `web-php-distributions` into this repository (see also https://github.com/php/web-php-distributions/pull/9)
- With removing `php-keyring.gpg` from `web-php-distributions`, I've set up a redirect in this PR so that `/distributions/php-keyring.gpg` redirects permanently to `/php-keyring.gpg`
- I've added a new GPG signing key for release-managers@php.net (see https://github.com/php/php-src/pull/10604 for more details)


include/layout.inc
```diff
@@ -195,8 +200,11 @@ function download_link($file, $title)
...
 
     if (@file_exists($local_file . ".asc")) {
-        echo " ";
         $sig_link = "/distributions/$file.asc";
+    }
+
+    if (isset($sig_link)) {
+        echo " ";
         echo make_link($sig_link, "(sig)");
     }
```
ramsey, line 210: The code below here (and not shown in the diff) checks the file on local disk (i.e., in `/distributions/`) and calculates the file size to display on the downloads page.

With this new approach, since the files aren't on local disk, none of them will display their sizes, unless we want to make API requests to GitHub to get the file sizes. I chose not to do that in this iteration, since I don't think it's efficient to make those API requests on every page load, and I'm not sure where we would cache that information (if we have a place to cache data).

carusogabriel:
@ramsey Thanks for proposing this. If you don't mind, give me a few days and I'll review it!

PR #762, opened 2023-03-22
fix: upgrate colors to modern colors

SimonCastano:
the PHP documentation is really ugly as a ui designer with knowledge in php I wanted to make a small contribution it is not very big but I hope you review it

I am sure that one of the causes that people say that PHP is dying is that the documentation is very complicated to read and it is not pleasant.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+    "workbench.colorCustomizations": {
+        "activityBar.background": "#043513",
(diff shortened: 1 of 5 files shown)


morrisonlevi:
At the very least, you are going to need to show screenshots of the home page, releases page, and some documentation page to show your changes.

mbabker:
I got curious about this, this is the homepage with the CSS changes applied.  

Screen Shot 2023-04-14 at 11 11 11 AM

PR #772, opened 2023-04-18
Use abbreviated month names

captn3m0:
This changes the API response to always use abbreviated (en) month names. Allows for slightly more stable API clients, which don't have to worry about parsing both kinds of dates.

I noticed this while working on https://github.com/endoflife-date/release-data/pull/91, where I had to write
a [fallback](https://github.com/endoflife-date/release-data/pull/91/files#diff-cf71bc63cf044e0a6b6656eb67cd061b4fa9bd4bd7e0e7d4ee46c7b3bcf95d00R13-R15) where the date format
changes.

Ideally, this should use something non-locale based like `YYYY-MM-DD`, 
happy to file a separate PR for that if that sounds good. (At the expense of breaking the API?)

### DIFF
--- a/include/releases.inc
+++ b/include/releases.inc
@@ -3336,7 +3336,7 @@ $OLDRELEASES = array (
       'tags' => 
       array (
       ),
-      'date' => '11 June 2020',
+      'date' => '11 Jun 2020',
       'source' => 
       array (
         0 => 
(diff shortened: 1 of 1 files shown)


derickr:
This is a generated file, so not wise to edit.

captn3m0:
Closing this, but it would be nice if the generation script could be fixed.

PR #775, opened 2023-05-12
Update Entry.php

oladoyinbov:
Add Strict Types


Girgias:
This doesn't really add any value so closing.

PR #777, opened 2023-05-16
Fix code blocks overflowing content. 

leopoletto:
This PR fixes the code overflowing content by enabling the horizontal scroll. 

The PHP 8.2 release page has the code content overflowing when accessing via mobile, and the same happens on many pages across the documentation, mainly in user comments. 

image
image

The result looks like this:

image
image

### DIFF
--- a/styles/theme-base.css
+++ b/styles/theme-base.css
@@ -1326,7 +1326,8 @@ div.tip p:first-child {
     margin-bottom: 1.5rem;
 }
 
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
Is it the same as https://github.com/php/web-php/pull/770

leopoletto:
Unfortunately, the #770 doesn't fix this page: https://www.php.net/releases/8.2/en.php

PR #788, opened 2023-06-23
Fix autoload redirect failure

### LINKED ISSUE #787: `/autoload` renders a blank page
It seems the intent is to forward [`/autoload`](https://www.php.net/autoload) to `language.oop5.autoload` (like [`/__autoload`](https://www.php.net/__autoload)):

https://github.com/php/web-php/blob/72d69120a8c2860072df4fcbaa856d1d8525342e/error.php#L424-L425

I'm guessing the web server sees [`autoload.php`](https://github.com/php/web-php/blob/72d69120a8c2860072df4fcbaa856d1d8525342e/autoload.php) and runs that instead. In that case renaming `autoload.php` to something else would fix.

sy-records:
Fix #787

### DIFF
--- a/bin/createNewsEntry
+++ b/bin/createNewsEntry
@@ -2,7 +2,7 @@
 <?php
 PHP_SAPI == 'cli' or die("Please run this script using the cli sapi");
 
-require_once __DIR__ . '/../autoload.php';
+require_once __DIR__ . '/../src/autoload.php';
 
 use phpweb\News\Entry;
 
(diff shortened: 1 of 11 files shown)


salathe:
Alternatively, move the `autoload.php` file out of the top-level folder.  Perhaps into `src` since it is concerned with loading with the classes/files in there, or `include` since it's a file to be included (rather than executed standalone).

salathe (approved):

PR #792, opened 2023-07-09
Fix ChangeLog-8.php typo.

Maikuolan:
I think, release date for PHP 8.1.21 should be July 6th, not June 6th.


saundefined (approved):
Indeed, thanks!

PR #794, opened 2023-07-11
Update Longhorn PHP 2023 CFP Deadline

iansltx:
CFP was extended, so bumping the deadline for the post. Will add another post once the schedule is announced.


Girgias:
Thank you!

PR #800, opened 2023-08-11
Remove double underline on abbr

alfsb:
UAs already styles `<abbr>` with a underline, and web-php styles double its:

See, for example: https://www.php.net/manual/en/install.unix.php

Google:
image

Firefox:
image


TimWolla (approved):
This makes sense to me.

alfsb:
In this repo I don't have commit access.

PR #804, opened 2023-08-21
Add redirect for reserved keywords

fredden:
When browsing to https://php.net/yield recently, I was surprised to see a message saying that "yield doesn't exist". The `yield` keyword does exist, but doesn't have a dedicated page so hasn't been indexed. This pull request adds a special case for this keyword so that it can be found by users.

I've also looked into how to get all keywords declared automatically. The best I can come up with today (without significant changes to how the search / look-up process works) is over on the 'systems' repository: https://github.com/php/systems/pull/19

**Update**: I've now gone through all the reserved words listed on https://php.net/reserved and added these to this pull request.


include/errors.inc
```diff
@@ -454,8 +476,70 @@ function is_known_term (string $term): ?string {
         'superglobals' => 'language.variables.superglobals.php',
         'switch' => 'control-structures.switch.php',
         'timestamp' => 'function.time.php',
+        'true' => 'language.types.boolean.php',
         'try' => 'language.exceptions.php',
         'upload' => 'features.file-upload.php',
+        'use' => 'language.namespaces.php',
+        'void' => 'language.types.void.php',
+        'xor' => 'language.operators.logical.php',
+        'yield from' => 'language.generators.syntax.php#control-structures.yield.from',
```
morrisonlevi, line 485: Are spaces allowed in this context?

derickr:
Merged after a rebase, thanks!

PR #807, opened 2023-08-26
PHP 8.3 release page

saundefined:
- [x] Update logo
- [x] Finish the English version
- [x] Put messages in lang-files
- [x] Add links to Documentation
- [x] Merge PR without updates in `include/header.inc` and `index.php`
- [ ] Waiting for the release of PHP 8.3 🎉 
- [ ] Update `include/header.inc` and `index.php`

Any help and suggestions are welcomed :)

Preview: https://php.sergeypanteleev.com/releases/8.3/en.php


releases/8.3/release.inc
```diff
@@ -0,0 +1,269 @@
...
+PHP
+                        ); ?>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="php8-compare">
+            <h2 class="php8-h2" id="null_false_true_types">
+                <code>Randomizer</code> additions
```
TimWolla, line 135: For `Randomizer::getFloat()`:

Before:
```php
// Returns a random float between $min and $max, both including.
function getFloat(float $min, float $max) {
    // This algorithm is biased for specific inputs and may
    // return values outside the given range. This is impossible
    // to work around in userland.
    return (random_int(0, PHP_INT_MAX) / PHP_INT_MAX) * ($max - $min) + $min;
}

$temperature = getFloat(-89.2, 56.7);

$chanceForTrue = 0.1;
// getFloat(0, 1) might return the upper bound, i.e. 1,
// introducing a small bias.
$myBoolean = getFloat(0, 1) < $chanceForTrue;
```

After:
```php
$randomizer = new \Random\Randomizer();

$temperature = $randomizer->getFloat(-89.2, 56.7, \Random\IntervalBoundary::ClosedClosed);

$chanceForTrue = 0.1;
// Randomizer::nextFloat() is equivalent to Randomizer::getFloat(0, 1, \Random\IntervalBoundary::ClosedOpen).
// The upper bound, i.e. 1, will not be returned.
$myBoolean = $randomizer->nextFloat(); < $chanceForTrue;
```

------------------------

In the description consider mentioning the paper by Prof. Goualard:

From the RFC at https://wiki.php.net/rfc/randomizer_additions#getfloat:

> The algorithm used is the γ-section algorithm as published in: Drawing Random Floating-Point Numbers from an Interval. Frédéric Goualard, ACM Trans. Model. Comput. Simul., 32:3, 2022. https://doi.org/10.1145/3503512

releases/8.3/release.inc
```diff
@@ -0,0 +1,269 @@
...
+PHP
+                        ); ?>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="php8-compare">
+            <h2 class="php8-h2" id="null_false_true_types">
+                Improved <code>unserialize()</code> error handling
```
TimWolla, line 197: It probably doesn't make sense to mention this RFC this prominently, because the `UnserializationFailedException` part didn't pass.

releases/8.3/release.inc
```diff
@@ -0,0 +1,269 @@
...
+            <div class="php8-subtitle"><?= message('main_subtitle', $lang) ?></div>
+            <div class="php8-button-wrapper center">
+                <a class="php8-button php8-button_light" href="/downloads"><?= message('upgrade_now', $lang) ?></a>
+            </div>
+        </div>
+    </section>
+
+    <section class="php8-section center">
+        <div class="php8-compare">
+            <h2 class="php8-h2" id="readonly_classes">
```
TimWolla, line 41: Wrong ID here (and for the other `h2` headings).

releases/8.3/release.inc
```diff
@@ -0,0 +1,269 @@
...
+
+    <section class="php8-section center php8-section_light php8-columns">
+        <div class="php8-column">
+            <h2 class="php8-h2" id="other_new_things"><?= message('new_classes_title', $lang) ?></h2>
+            <div class="php8-compare__content php8-compare__content--block">
+                <ul>
+                    <li>New <code>json_validate</code> function</li>
+                    <li>New <code>mb_str_pad</code> function</li>
+                    <li>New <code>posix_sysconf</code>, <code>posix_pathconf</code>, <code>posix_fpathconf</code>, and <code>posix_eaccess</code> functions</li>
+                    <li>New <code>Randomizer::getBytesFromString</code>, <code>Randomizer::nextFloat</code>, and <code>Randomizer::getFloat</code> functions, also <code>Randomizer::IntervalBoundary</code> enum.</li>
```
TimWolla, line 235: This is already mentioned above.

releases/8.3/release.inc
```diff
@@ -0,0 +1,269 @@
...
+        <div class="php8-column">
+            <h2 class="php8-h2" id="deprecations_and_bc_breaks"><?= message('bc_title', $lang) ?></h2>
+            <div class="php8-compare__content">
+                <ul>
+                    <li>More Appropriate Date/Time Exceptions</li>
+                    <li>Negative indices in arrays</li>
+                    <li>Changes to the <code>range</code> function</li>
+                    <li>Traits and static properties</li>
+                    <li>The <code>U_MULTIPLE_DECIMAL_SEPERATORS</code> constant had been deprecated in favor of <code>U_MULTIPLE_DECIMAL_SEPARATORS</code></li>
+                    <li>Changed <code>mt_srand</code> and <code>srand</code> functions</li>
```
TimWolla, line 250: The relevant part that should be spelled out explicitly is the deprecation of `MT_RAND_PHP`.

releases/8.3/release.inc
```diff
@@ -0,0 +1,329 @@
...
+PHP
+                        ); ?>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="php8-compare">
+            <h2 class="php8-h2" id="override_attribute">
+                <code>#[Override]</code> attribute
```
TimWolla, line 104:
```suggestion
                <code>#[\Override]</code> attribute
```
I've intentionally written this with the `\` everywhere to make it copy and paste-safe.

releases/8.3/release.inc
```diff
@@ -0,0 +1,342 @@
...
+                <a class="php8-rfc" href="https://wiki.php.net/rfc/marking_overriden_methods">RFC</a>
+            </h2>
+            <div class="php8-compare__main">
+                <div class="php8-compare__block example-contents">
+                    <div class="php8-compare__label">PHP &lt; 8.3</div>
+                    <div class="php8-code phpcode">
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
+// TODO
```
TimWolla, line 113:
```suggestion
use PHPUnit\Framework\TestCase;
 
final class MyTest extends TestCase
{
    protected $logFile;

    protected function setUp(): void
    {
        $this->logFile = fopen('/tmp/logfile', 'w');
    }

    protected function taerDown(): void
    {
        fclose($this->logFile);
        unlink('/tmp/logfile');
    }
}

// The log file will never be removed, because the
// method name was mistyped (taerDown vs tearDown).
```

Stealing @derickr's example from here: https://github.com/php/presentations/blob/master/slides/internals/php83-marking-overridden-methods.xml (taerDown is better than setUpp as used in the RFC :grin:).

releases/8.3/release.inc
```diff
@@ -0,0 +1,342 @@
...
+                    </div>
+                </div>
+                <div class="php8-compare__arrow"></div>
+                <div class="php8-compare__block example-contents">
+                    <div class="php8-compare__label php8-compare__label_new">PHP 8.3</div>
+                    <div class="php8-code phpcode">
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
+// TODO
+PHP
```
TimWolla, line 126:
```suggestion
use PHPUnit\Framework\TestCase;
 
final class MyTest extends TestCase
{
    protected $logFile;

    protected function setUp(): void
    {
        $this->logFile = fopen('/tmp/logfile', 'w');
    }

    #[\Override]
    protected function taerDown(): void
    {
        fclose($this->logFile);
        unlink('/tmp/logfile');
    }
}

// Fatal error: MyTest::taerDown() has #[\Override] attribute,
// but no matching parent method exists
```

releases/8.3/release.inc
```diff
@@ -0,0 +1,397 @@
...
+    -89.2,
+    56.7,
+    \Random\IntervalBoundary::ClosedClosed,
+);
+
+$chanceForTrue = 0.1;
+// Randomizer::nextFloat() is equivalent to
+// Randomizer::getFloat(0, 1, \Random\IntervalBoundary::ClosedOpen).
+// The upper bound, i.e. 1, will not be returned.
+$myBoolean = $randomizer->nextFloat(); < $chanceForTrue;
```
TimWolla, line 300:
```suggestion
$myBoolean = $randomizer->nextFloat() < $chanceForTrue;
```

releases/8.3/release.inc
```diff
@@ -0,0 +1,397 @@
...
+PHP
+                        ); ?>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="php8-compare">
+            <h2 class="php8-h2" id="randomizer_get_float">
+                New <code>Randomizer::getFloat</code> and <code>Randomizer::nextFloat</code> methods
```
TimWolla, line 253:
```suggestion
                New <code>Randomizer::getFloat()</code> and <code>Randomizer::nextFloat()</code> methods
```

releases/8.3/release.inc
```diff
@@ -0,0 +1,397 @@
...
+PHP
+                        ); ?>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="php8-compare">
+            <h2 class="php8-h2" id="randomizer_get_bytes_from_string">
+                New <code>Randomizer::getBytesFromString</code> method
```
TimWolla, line 189:
```suggestion
                New <code>Randomizer::getBytesFromString()</code> method
```

TimWolla (commented):
Thank you, code blocks are looking good now. I'll probably will write some short description for the Override and Randomizer stuff later, because while the code examples are good, I don't think they are sufficient to properly highlight the new features 😄

releases/8.3/release.inc
```diff
@@ -0,0 +1,464 @@
...
+final class MyTest extends TestCase
+{
+    protected $logFile;
+
+    protected function setUp(): void
+    {
+        $this->logFile = fopen('/tmp/logfile', 'w');
+    }
+
+    protected function taerDown(): void
```
Abijeet, line 191: tearDown?

saundefined: No, it's a special mistyped.

We won't notice the mistyped in PHP < 8.3, but in PHP 8.3 when defining `#[\Override]` we will get an error because the parent class doesn't have `taerDown`, it has `tearDown`.

releases/8.3/release.inc
```diff
@@ -0,0 +1,455 @@
...
+                            <<<'PHP'
+class Foo {
+    const PHP = 'PHP 8.3';
+}
+
+$searchableConstant = 'PHP';
+
+$reflectedClass = new ReflectionClass(Foo::class);
+
+var_dump($reflectedClass->getConstant($searchableConstant));
```
TimWolla, line 383:
```suggestion
var_dump(constant(Foo::class . "::{$searchableConstant}"));
```

releases/8.3/release.inc
```diff
@@ -0,0 +1,455 @@
...
+        }
+
+        return $clone;
+    }
+}
+
+$today = new Foo(new \DateTime());
+$tomorrow = $today->with(
+    dateTime: (new \DateTime())->modify('+1 day')
+);
```
TimWolla, line 87: This is not a good comparison, because it does something entirely different: This is effectively the “clone with” proposal and it allows to externally modify the properties. The PHP 8.3 example just performs deep-cloning.

The correct example for PHP 8.2 would likely be “Fatal Error” and it likely makes sense to move the “Readonly Amendments” down in the list, because while this is a necessary improvement for deep cloning, it isn't really that interesting to most users.

bukka:
I would say that `json_validate` is probably one of the bigger feature in 8.3 so it should go at least above Randomizer ones which I'm not sure we should be mention in such detail. Surely introducing Randomizer was worth to mentioning as it was a big thing but adding some extra less used methods for likely small number of use cases should not be mentioned in such detail IMHO. 

I would keep all the engine features first and then maybe some note about json_validate and Randomizer withtout too detailed examples.

TimWolla:
> but adding some extra less used methods for likely small number of use cases should not be mentioned in such detail IMHO.

For the float methods I agree that the use case is narrow, but `Randomizer::getBytesFromString()` is probably much more interesting to the average developer than `json_validate()`. It's the first time, PHP can natively generate random strings with specific characters and I'd say that's a reasonably common thing to do, e.g. to generate random passwords or identifiers that should not be limited to just hex.

While the float methods are less interesting to the average developer, I still believe it makes sense to list them somewhat prominently, because the example also point out an important fact: It's impossible to generate an unbiased float for arbitrary intervals in userland.

> I would keep all the engine features first

Standard library is equally important as language changes. I believe the order of the highlighted improvements should be the order of interest for the average developer and frankly PHP 8.3 doesn't contain that many interesting engine features, because many interesting things have been postponed to PHP 8.4. In fact I see myself using the new Randomizer methods or even `str_increment()` and `str_decrement()` much more often than the dynamic class constant fetch.

bukka:
> > but adding some extra less used methods for likely small number of use cases should not be mentioned in such detail IMHO.
> 
> For the float methods I agree that the use case is narrow, but `Randomizer::getBytesFromString()` is probably much more interesting to the average developer than `json_validate()`. It's the first time, PHP can natively generate random strings with specific characters and I'd say that's a reasonably common thing to do, e.g. to generate random passwords or identifiers that should not be limited to just hex.

Ok that probably makes it a bit more useful so might make sense to keep it.

It's got a small issue though - the name is so long that it doesn't look good on mobile - maybe smaller font should be used.

> 
> While the float methods are less interesting to the average developer, I still believe it makes sense to list them somewhat prominently, because the example also point out an important fact: It's impossible to generate an unbiased float for arbitrary intervals in userland.

I think his can be done just in docs. We don't need to list it as a top feature for that purpose IMHO. It can be mentioned in other functions potentially. I guess most people won't even know what an unbiased float is or won't care about it.

> 
> > I would keep all the engine features first
> 
> Standard library is equally important as language changes. I believe the order of the highlighted improvements should be the order of interest for the average developer and frankly PHP 8.3 doesn't contain that many interesting engine features, because many interesting things have been postponed to PHP 8.4. In fact I see myself using the new Randomizer methods or even `str_increment()` and `str_decrement()` much more often than the dynamic class constant fetch.

This is very subjective though. We can't really know what's more interesting for users. We can pick some features that we think are useful but not sure if we should do order strictly by that. It might be better to order it in a more logical way for reader so related things follow.

So I would go with this order:

1. Typed class constants
2. Dynamic class constant fetch
3. #[\Override] attribute
4. Readonly amendments
5. json_validate
6. Randomizer::getBytesFromString()

And then mention Random float methods as the first methods in the following list.

releases/8.3/languages/en.php
```diff
@@ -0,0 +1,43 @@
...
+    'new_mb_str_pad' => 'New <code>mb_str_pad()</code> function.',
+    'new_posix' => 'New <code>posix_sysconf()</code>, <code>posix_pathconf()</code>, <code>posix_fpathconf()</code>, and <code>posix_eaccess()</code> functions.',
+    'new_reflection' => 'New <code>ReflectionMethod::createFromMethodName()</code> method.',
+    'new_socket' => 'New <code>socket_atmark()</code> function.',
+    'new_str' => 'New <code>str_increment()</code>, <code>str_decrement()</code>, and <code>stream_context_set_options()</code> functions.',
+    'new_ziparchive' => 'New <code>ZipArchive::getArchiveFlag()</code> function.',
+
+    'bc_title' => 'Deprecations and backward compatibility breaks',
+    'bc_datetime' => 'More Appropriate Date/Time Exceptions.',
+    'bc_arrays' => 'Negative indices in arrays.',
```
bukka, line 33: This is very unclear so it should be probably either removed or extended. Assuming it's for this:
```
Assigning a negative index n to an empty array will now make sure that the
next index is n+1 instead of 0
```

releases/8.3/languages/en.php
```diff
@@ -0,0 +1,43 @@
...
+    'bc_range' => 'Changes to the <code>range()</code> function.',
+    'bc_traits' => 'Traits and static properties.',
```
bukka, line 35:
```suggestion
    'bc_traits' => 'Changes in re-declaration of static properties in traits.',
```

releases/8.3/languages/en.php
```diff
@@ -0,0 +1,43 @@
...
+    'bc_umultipledecimalseparators' => 'The <code>U_MULTIPLE_DECIMAL_SEPERATORS</code> constant had been deprecated in favor of <code>U_MULTIPLE_DECIMAL_SEPARATORS</code>.',
```
bukka, line 36:
```suggestion
    'bc_umultipledecimalseparators' => 'The <code>U_MULTIPLE_DECIMAL_SEPERATORS</code> constant is deprecated in favor of <code>U_MULTIPLE_DECIMAL_SEPARATORS</code>.',
```
looks like present tense is used elsewhere...

nielsdos:
Speaking of standard library, perhaps the "New Classes, Interfaces, and Functions" section should include one bullet point letting users know that there are new DOM manipulation methods added in ext/dom in 8.3?

nielsdos:
Thanks Sergey!
At your live version I still see the php 8.2 logo in the navigation bar, but the 8.3 logo in the hero. I suppose that's because you forgot to upload the header file? Or is there still a filename that needs to be updated?

saundefined:
> Thanks Sergey! At your live version I still see the php 8.2 logo in the navigation bar, but the 8.3 logo in the hero. I suppose that's because you forgot to upload the header file? Or is there still a filename that needs to be updated?

Oh, thanks! Yeah, I forgot to update that, fixed :)

bukka:
Might be also worth to mention https://github.com/php/php-src/commit/0dadd6616a491418871fb0b41590a73b128aa212 in "New Classes, Interfaces, and Functions". Something like `Support for generation EC keys with custom EC parameters in OpenSSL extension`

releases/8.3/languages/en.php
```diff
@@ -0,0 +1,46 @@
...
+    'documentation' => 'Doc',
+    'main_title' => 'Released!',
+    'main_subtitle' => 'PHP 8.3 is a major update of the PHP language.<br class="display-none-md">It contains many new features, including Readonly amendments, Typed class constants, Randomizer additions, performance improvements and more.',
+    'upgrade_now' => 'Upgrade to PHP 8.3 now!',
+
+    'readonly_title' => 'Readonly amendments',
+    'json_validate_title' => 'New <code>json_validate()</code> function',
+    'typed_class_constants_title' => 'Typed class constants',
+    'override_title' => '<code>#[\Override]</code> attribute',
+    'randomizer_getbytesfromstring_title' => 'New <code>Randomizer::getBytesFromString()</code> method',
```
TimWolla, line 14:
```suggestion
    'randomizer_getbytesfromstring_title' => 'New <code>Randomizer<span style="word-break: break-all;">::</span>getBytesFromString()</code> method',
```
This should fix the rendering on mobile:
image

TimWolla:
> It's got a small issue though - the name is so long that it doesn't look good on mobile - maybe smaller font should be used.

Added suggestion to fix this.

releases/8.3/release.inc
```diff
@@ -0,0 +1,475 @@
...
+PHP
+                        ); ?>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="php8-compare">
+            <h2 class="php8-h2" id="readonly_classes">
+                <?= message('readonly_title', $lang) ?>
```
TimWolla, line 201: It probably makes sense to add an explanation outside of the code blocks. Here's my suggestion:

> `readonly` properties may now be modified once within the magic `__clone` method to enable deep-cloning of readonly properties.

releases/8.3/release.inc
```diff
@@ -0,0 +1,475 @@
...
+            <div class="php8-compare__main">
+                <div class="php8-compare__block example-contents">
+                    <div class="php8-compare__label">PHP &lt; 8.3</div>
+                    <div class="php8-code phpcode">
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
+readonly class Foo {
+    public \DateTime $dateTime;
+
+    function __construct(\DateTime $dateTime) {
```
TimWolla, line 213:
```suggestion
    public function __construct(\DateTime $dateTime) {
```
This is a little cleaner.

releases/8.3/release.inc
```diff
@@ -0,0 +1,475 @@
...
+                <div class="php8-compare__arrow"></div>
+                <div class="php8-compare__block example-contents">
+                    <div class="php8-compare__label php8-compare__label_new">PHP 8.3</div>
+                    <div class="php8-code phpcode">
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
+readonly class Foo {
+	public \DateTime $dateTime;
+
+	function __construct(\DateTime $dateTime) {
```
TimWolla, line 241:
```suggestion
	public function __construct(\DateTime $dateTime) {
```

releases/8.3/release.inc
```diff
@@ -0,0 +1,475 @@
...
+		$this->dateTime = $dateTime;
+	}
+
+	public function __clone()
+    {
+        $this->dateTime = clone $this->dateTime;
+    }
```
TimWolla, line 248: Mixed indentation here. It should probably consistently use 4 spaces, instead of tabs.

releases/8.3/release.inc
```diff
@@ -0,0 +1,475 @@
...
+PHP
+                        ); ?>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="php8-compare">
+            <h2 class="php8-h2" id="override_attribute">
+                <?= message('override_title', $lang) ?>
```
TimWolla, line 131: Explanation text:

> By adding the `#[\Override]` attribute to a method, PHP will ensure that a method with the same name exists in a parent class or in an implemented interface. Adding the attribute makes it clear that overriding a parent method is intentional and simplifies refactoring, because the removal of an overridden parent method will be detected.

releases/8.3/release.inc
```diff
@@ -0,0 +1,475 @@
...
+PHP
+                        ); ?>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="php8-compare">
+            <h2 class="php8-h2" id="randomizer_get_bytes_from_string">
+                <?= message('randomizer_getbytesfromstring_title', $lang) ?>
```
TimWolla, line 300: Explanation:

> The [Random Extension](https://www.php.net/releases/8.2/en.php#random_extension) that was added in PHP 8.2 was extended by a new method to generate random strings consisting of specific bytes only. This method allows to easily generate random identifiers, such as domain names, and numeric strings of arbitrary length.

releases/8.3/release.inc
```diff
@@ -0,0 +1,475 @@
...
+PHP
+                        ); ?>
+                    </div>
+                </div>
+            </div>
+        </div>
+
+        <div class="php8-compare">
+            <h2 class="php8-h2" id="json_validate">
+                <?= message('json_validate_title', $lang) ?>
```
TimWolla, line 264: Explanation:

> `json_validate()` allows to check if a string is syntactically valid JSON, while being more efficient than `json_decode()`.

releases/8.3/release.inc
```diff
@@ -0,0 +1,475 @@
...
+                <a class="php8-rfc" href="https://wiki.php.net/rfc/json_validate">RFC</a>
+            </h2>
+            <div class="php8-compare__main">
+                <div class="php8-compare__block example-contents">
+                    <div class="php8-compare__label">PHP &lt; 8.3</div>
+                    <div class="php8-code phpcode">
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
+function json_validate(string $string) {
```
TimWolla, line 273:
```suggestion
function json_validate(string $string): bool {
```

TimWolla (commented):
Some more suggestions

releases/8.3/release.inc
```diff
@@ -0,0 +1,475 @@
...
+        </div>
+
+
+    </section>
+
+    <section class="php8-section center php8-section_light php8-columns">
+        <div class="php8-column">
+            <h2 class="php8-h2" id="other_new_things"><?= message('new_classes_title', $lang) ?></h2>
+            <div class="php8-compare__content php8-compare__content--block">
+                <ul>
```
cookieguru, line 432: Can these be linked to the individual man pages?  Is there a way to link them such that they go to the selected language, and fall back to another if the requested language doesn't exist?  e.g. str_decrement has en and fr translations but not pl.  A link to `https://www.php.net/manual/$lang/function.str-decrement.php` could lead to a 404 page

releases/8.3/release.inc
cookieguru: More *memory efficient?

AFAIK the CPU time is the same

releases/8.3/release.inc
saundefined: Sure, each translation can link to the documentation pages that suit it (see example in 8.2 release page)

releases/8.3/release.inc
TimWolla: Both, because you also save some processing to actually create the structure. It's mainly memory though. I think the simplification is fine for this page.

saundefined:
So, I'm going to merge this PR on Tuesday (November 21),
I've removed the links to the PHP 8.3 page in the https://github.com/php/web-php/pull/807/commits/ac55c2ba6d8827d40a433f6af45dd7cdaf87c085 commit and will add them in a separate PR.

On Thursday (November 23), along with the release of PHP 8.3, we will merge the second PR.

During this time we can check that everything is displaying correctly and make any adjustments if needed.

Feel free to suggest fixes now if you see fit :)

releases/8.3/release.inc
```diff
@@ -0,0 +1,490 @@
...
+            <div class="php8-compare__content">
+                <?= message('readonly_description', $lang) ?>
+            </div>
+        </div>
+
+        <div class="php8-compare">
+            <h2 class="php8-h2" id="json_validate">
+                <?= message('json_validate_title', $lang) ?>
+                <a class="php8-rfc" href="https://wiki.php.net/rfc/json_validate">RFC</a>
+                <a class="php8-rfc" href="/manual/en/function.json-validate.php"><?= message('documentation', $lang) ?></a>
```
cookieguru, line 273:
```suggestion
                <a class="php8-rfc" href="/manual/<?= $lang ?>/function.json-validate.php"><?= message('documentation', $lang) ?></a>
```

saundefined: Good catch, thanks!

releases/8.3/release.inc
```diff
@@ -0,0 +1,490 @@
...
+                    <div class="php8-code phpcode">
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
+interface I {
+    // We may naively assume that the PHP constant is always a string
+    const PHP = 'PHP 8.2';
+}
+
+class Foo implements I {
+    const PHP = []; // But it may be an array...
```
TimWolla, line 57:
```suggestion
    // But implementing classes may define it as an array.
    const PHP = [];
```
For visual consistency and clearer phrasing.

releases/8.3/release.inc
```diff
@@ -0,0 +1,490 @@
...
+                <a class="php8-rfc" href="https://wiki.php.net/rfc/typed_class_constants">RFC</a>
+            </h2>
+            <div class="php8-compare__main">
+                <div class="php8-compare__block example-contents">
+                    <div class="php8-compare__label">PHP &lt; 8.3</div>
+                    <div class="php8-code phpcode">
+                        <?php highlight_php_trimmed(
+                            <<<'PHP'
+interface I {
+    // We may naively assume that the PHP constant is always a string
```
TimWolla, line 52:
```suggestion
    // We may naively assume that the PHP constant is always a string.
```

releases/8.3/release.inc
```diff
@@ -0,0 +1,490 @@
...
+                            <<<'PHP'
+interface I {
+    const string PHP = 'PHP 8.3';
+}
+
+class Foo implements I {
+    const string PHP = [];
+}
+
+// Fatal error: Cannot use array as value for class constant Foo::PHP of type string
```
TimWolla, line 78:
```suggestion
// Fatal error: Cannot use array as value for class constant
// Foo::PHP of type string
```
hard break for visual consistency.

releases/8.3/languages/en.php
```diff
@@ -0,0 +1,50 @@
...
+    'new_ziparchive' => 'New <a href="/manual/en/ziparchive.getarchiveflag.php"><code>ZipArchive::getArchiveFlag()</code></a> method.',
+    'new_openssl_ec' => 'Support for generation EC keys with custom EC parameters in OpenSSL extension.',
+
+    'bc_title' => 'Deprecations and backward compatibility breaks',
+    'bc_datetime' => 'More Appropriate Date/Time Exceptions.',
+    'bc_arrays' => 'Assigning a negative index <code>n</code> to an empty array will now make sure that the next index is <code>n + 1</code> instead of <code>0</code>.',
+    'bc_range' => 'Changes to the <code>range()</code> function.',
+    'bc_traits' => 'Changes in re-declaration of static properties in traits.',
+    'bc_umultipledecimalseparators' => 'The <code>U_MULTIPLE_DECIMAL_SEPERATORS</code> constant is deprecated in favor of <code>U_MULTIPLE_DECIMAL_SEPARATORS</code>.',
+    'bc_mtrand' => 'The <code>MT_RAND_PHP</code> Mt19937 variant is deprecated.',
```
TimWolla, line 44: Perhaps link this to https://www.php.net/manual/en/random.constants.php#constant.mt-rand-php

releases/8.3/languages/en.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+return [
+    'common_header' => 'PHP 8.3 is a major update of the PHP language. Readonly amendments, Typed class constants, #[Override] attribute, Randomizer additions, performance improvements and more',
```
TimWolla, line 4: The contents of this should probably be consistent with `main_subtitle`.

releases/8.3/languages/en.php
```diff
@@ -0,0 +1,50 @@
...
+    'main_title' => 'Released!',
+    'main_subtitle' => 'PHP 8.3 is a major update of the PHP language.<br class="display-none-md">It contains many new features, including Readonly amendments, Typed class constants, Randomizer additions, performance improvements and more.',
+    'upgrade_now' => 'Upgrade to PHP 8.3 now!',
+
+    'readonly_title' => 'Readonly amendments',
+    'readonly_description' => '<code>readonly</code> properties may now be modified once within the magic <code>__clone</code> method to enable deep-cloning of readonly properties.',
+    'json_validate_title' => 'New <code>json_validate()</code> function',
+    'json_validate_description' => '<code>json_validate()</code> allows to check if a string is syntactically valid JSON, while being more efficient than <code>json_decode()</code>.',
+    'typed_class_constants_title' => 'Typed class constants',
+    'override_title' => '<code>#[\Override]</code> attribute',
```
TimWolla, line 15:
```suggestion
    'override_title' => 'New <code>#[\Override]</code> attribute',
```
For consistency.

releases/8.3/languages/en.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+return [
+    'common_header' => 'PHP 8.3 is a major update of the PHP language. Readonly amendments, Typed class constants, #[Override] attribute, Randomizer additions, performance improvements and more',
+    'documentation' => 'Doc',
+    'main_title' => 'Released!',
+    'main_subtitle' => 'PHP 8.3 is a major update of the PHP language.<br class="display-none-md">It contains many new features, including Readonly amendments, Typed class constants, Randomizer additions, performance improvements and more.',
+    'upgrade_now' => 'Upgrade to PHP 8.3 now!',
+
+    'readonly_title' => 'Readonly amendments',
```
TimWolla, line 10:
```suggestion
    'readonly_title' => 'Deep-cloning of readonly properties',
```
This is perhaps better, because “readonly amendments” could be anything.

releases/8.3/languages/en.php
```diff
@@ -0,0 +1,50 @@
...
+    'override_description' => 'By adding the <code>#[\Override]</code> attribute to a method, PHP will ensure that a method with the same name exists in a parent class or in an implemented interface. Adding the attribute makes it clear that overriding a parent method is intentional and simplifies refactoring, because the removal of an overridden parent method will be detected.',
+    'randomizer_getbytesfromstring_title' => 'New <code>Randomizer<span style="word-break: break-all;">::</span>getBytesFromString()</code> method',
+    'randomizer_getbytesfromstring_description' => 'The <a href="/releases/8.2/en.php#random_extension">Random Extension</a> that was added in PHP 8.2 was extended by a new method to generate random strings consisting of specific bytes only. This method allows to easily generate random identifiers, such as domain names, and numeric strings of arbitrary length.',
+    'randomizer_getfloat_nextfloat_title' => 'New <code>Randomizer::getFloat()</code> and <code>Randomizer::nextFloat()</code> methods',
+    'randomizer_getfloat_nextfloat_description' => 'The algorithm used is the γ-section algorithm as published in:
```
TimWolla, line 20: This should get some more explanation, perhaps:

```
Due to the limited precision and implicit rounding of floating point numbers, generating an unbiased float lying within a specific interval is non-trivial and the commonly used userland solutions may generate biased results or numbers outside the requested range. The Randomizer was also extended with two methods to generate random floats in an unbiased fashion. The <code>Randomizer::getFloat()</code> method uses the γ-section algorithm that was published in <a href="https://doi.org/10.1145/3503512" target="_blank" rel="noopener noreferrer">Drawing Random Floating-Point Numbers from an Interval. Frédéric Goualard, ACM Trans. Model. Comput. Simul., 32:3, 2022.</a>
```

releases/8.3/languages/en.php
saundefined: Should we also use that wording in `common_header` and `main_subtitle`, what do you think?

TimWolla: Yes, that would make sense. How about something like this:

> It contains many new features, such as explicit typing of class constants, deep-cloning of readonly properties and additions to the randomness functionality. As always it also includes performance improvements, bug fixes and general cleanup.

For that “elevator pitch” I've focused on the three features that are easily understood with just a few words and no code examples and then mentioned the general improvements that go into every release.

Ayesh: I like this. Concise with a slight positive tone.

pronskiy:
It would be nice to mention the stack overflow detection https://github.com/php/php-src/pull/9104
/cc @arnaud-lb

Flote37:
Hi @saundefined, 
Thanks for this release page, it is really useful to catch up with PHP new feature.

I found a small link issue in the 8.3 release page (https://www.php.net/releases/8.3/en.php)
The link to the new "stream_context_set_options" function redirects to the old one (the one without the ending "s").
We have "https://www.php.net/manual/en/function.stream-context-set-option.php" instead of "https://www.php.net/manual/en/function.stream-context-set-options.php"

As this issue is already "merged", should I create a new issue ?

Flote37:
@cookieguru is this ok ? https://github.com/php/web-php/pull/827

PR #810, opened 2023-08-29
Use `white-space: pre-wrap` for code formatting

TimWolla:
Apparently code formatting is used for the entire user comments. Use `white-space: pre-wrap` to avoid horizontal scrolling of the plaintext description.

see 7d4038829e601bfdff3c434f5d8f20d84b344e27

Fixes php/doc-en#2724

### DIFF
--- a/include/layout.inc
+++ b/include/layout.inc
@@ -20,26 +20,14 @@ function highlight_php($code, $return = false)
     }
 
     // Fix output to use CSS classes and wrap well
-    $highlighted = '<div class="phpcode">' . str_replace(
+    $highlighted = '<div class="phpcode">' . strtr(
+        $highlighted,
         [
-            '&nbsp;',
-            '<br>',
-            '<span style="color: ',
-            '</font>',
-            "\n ",
-            '  ',
-            '  '
+            '&nbsp;' => ' ',
+            "\n" => '',
+
+            '<span style="color: ' => '<span class="',
         ],
-        [
-            ' ',
-            "<br>\n",
-            '<span class="',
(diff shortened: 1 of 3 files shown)


kamil-tekiela:
Also worth pointing out that the notes now have extra newlines.

kamil-tekiela:
This is done on purpose. See https://github.com/php/web-php/pull/770

However, you are right that in user notes there could be normal text that should not be scrollable. I guess that's the reason why it was done using `code span span` selector before. I believe you need to also add this:
```
.phpcode code span span {
    white-space: pre;
}
```

TimWolla:
> Also worth pointing out that the notes now have extra newlines.

Ugh. Are they stored in a pre-rendered version or is the post-processor just not applied to them? They also still have the `&nbsp;`.

> I believe you need to also add this:

You mean to prevent the wrapping from happening within the actual PHP code?

TimWolla:
> or is the post-processor just not applied to them?

Answering my own question: https://github.com/php/web-php/blob/bd160d89cf3e41f5f95719231709b99ebf56b0e5/include/layout.inc#L241.

I'll unify that function with PhD.

TimWolla:
> You mean to prevent the wrapping from happening within the actual PHP code?

This doesn't actually work, because this doesn't match the whitespace within PHP code. I don't really consider soft wrapping the PHP code to be an issue, though. It shouldn't change the semantic.

I've added a commit to unify the highlighting with PhD, though. I've also verified that the PHP 8.0 release page renders properly.

PR #814, opened 2023-09-19
Update contact.php

arnavnagpurkar:
Made formatting of contact.php better


arnavnagpurkar:
Can you tell me why was the PR closed ?

salathe:
Absolutely.  To repeat a reply to another of your recent pull requests elsewhere (python/cpython#109568), "it is needless churn."

arnavnagpurkar:
> Absolutely. To repeat a reply to another of your recent pull requests elsewhere ([python/cpython#109568](https://github.com/python/cpython/pull/109568)), "it is needless churn."

Thanks, but i'm still in class 8th and a learner

PR #842, opened 2023-11-25
FIX #502 Fixed search bar when viewport is 768px-858px

### LINKED ISSUE #502: Search bar breaks layout and hides text on viewports 768-858px
Steps to reproduce:
- browse to https://www.php.net/downloads with a viewport width between 768 and 858px

Observed result:
- search bar is bumped down to a second row in the header bar
- top row of text in the sidebar is hidden
Screen Shot 2022-02-24 at 13 39 33
(note that this is scrolled up all the way on the page)

Expected result:
- search bar scales horizontally to fit on a single header row

AsifR16:
Fixed issue #502
Issue: Search bar breaks layout and hides text on viewports 768-858px 
Solved!!!

### DIFF
--- a/styles/theme-base.css
+++ b/styles/theme-base.css
@@ -1718,6 +1718,9 @@ div.soft-deprecation-notice blockquote.sidebar {
 }
 
 @media (min-width:768px) {
+  #layout-content{
+    padding-top:2.5rem;
+  }
   #topsearch {
     float:right;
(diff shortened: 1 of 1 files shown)


saundefined:
Thanks, but locally I'm having trouble keeping up...
I'm going to close and reopen PR now to create a preview environment

PR #847, opened 2023-11-29
Enhancement: Require and use PHP 8.2

localheinz:
This pull request

- [x] requires and uses PHP 8.2

💁‍♂️ According to @pronskiy, the [website currently runs on PHP 8.2](https://x.com/pronskiy/status/1729819209553641860).

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.3"
+          - "8.2"
 
     steps:
       - name: "Checkout"
@@ -60,7 +60,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.3"
+          - "8.2"
 
     steps:
       - name: "Checkout"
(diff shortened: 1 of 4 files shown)


.github/workflows/integrate.yaml
```diff
@@ -51,6 +51,8 @@ jobs:
 
       - name: "Run friendsofphp/php-cs-fixer"
         run: "vendor/bin/php-cs-fixer fix --ansi --config=.php-cs-fixer.php --diff --dry-run --verbose"
+        env:
+          PHP_CS_FIXER_IGNORE_ENV: "1"
```
localheinz, line 55: We can update `friendsofphp/php-cs-fixer` in a separate pull request.

derickr:
How do we know whether the code can actually run with 8.2 without issues? It's not like we have tests for most of the site.

localheinz:
@derickr 

Which version does the website currently use in production?

See https://twitter.com/pronskiy/status/1729819209553641860:

CleanShot 2023-11-29 at 17 33 51@2x

saundefined:
Hm.. I'm assuming it's already 8.2 in the production host now, isn't it?

https://externals.io/message/121678

PR #856, opened 2023-12-06
Fix: Inline variables

localheinz:
This pull request

- [x] inlines variables


sy-records (approved):

kamil-tekiela (approved):
Even though it's a global, I doubt it was meant to be a variable. I think this is safe.

localheinz:
Thank you, @kamil-tekiela and @sy-records!

PR #861, opened 2023-12-06
Enhancement: Mark class as `readonly`

localheinz:
This pull request

- [x] marks a class as `readonly`

Follows #847.


localheinz:
Thank you, @kamil-tekiela!

PR #864, opened 2023-12-06
Enhancement: Enable `no_mixed_echo_print` fixer

localheinz:
This pull request

- [x] enables the `no_mixed_echo_print` fixer

Follows #559.

💁‍♂️ For reference, see https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/v3.40.2/doc/rules/alias/no_mixed_echo_print.rst.

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -38,6 +38,7 @@ $config
         'line_ending' => true,
         'new_with_parentheses' => true,
         'no_extra_blank_lines' => true,
+        'no_mixed_echo_print' => true,
         'no_trailing_whitespace' => true,
         'ordered_class_elements' => true,
         'random_api_migration' => true,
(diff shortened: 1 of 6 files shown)


.php-cs-fixer.php
```diff
@@ -38,6 +38,7 @@
         'line_ending' => true,
         'new_with_parentheses' => true,
         'no_extra_blank_lines' => true,
+        'no_mixed_echo_print' => true,
```
localheinz, line 41: Note that this fixer can be configured with the `use` option to prefer `echo` or `print`.

Using the default option (`echo`) applies fewer changes across the code base.

kamil-tekiela (approved):

localheinz:
Thank you, @kamil-tekiela!

PR #865, opened 2023-12-06
Fix: Add return type declarations

localheinz:
This pull request

- [x] adds a few return type declarations for functions returning instances of `DateTime` (or `null`)


kamil-tekiela (approved):

localheinz:
Thank you, @kamil-tekiela!

PR #869, opened 2023-12-06
Fix: Do not pass `null` to function that expects `DateTime` or `string`

localheinz:
This pull request

- [x] stops passing `null` to a function that expects `DateTime` or `string`

### Before

See http://localhost:8080/supported-versions.php:

CleanShot 2023-12-06 at 16 59 45@2x

### After

See http://localhost:8080/supported-versions.php:

CleanShot 2023-12-06 at 17 00 03@2x

### DIFF
--- a/include/branches.inc
+++ b/include/branches.inc
@@ -43,11 +43,10 @@ $BRANCHES = [
  * page. (Currently 28 days.) */
 $KEEP_EOL = new DateInterval('P28D');
 
-function format_interval($from, $to) {
+function format_interval($from, DateTime $to) {
     try {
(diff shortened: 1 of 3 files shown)


include/branches.inc
```diff
@@ -43,11 +43,10 @@ $BRANCHES = [
  * page. (Currently 28 days.) */
 $KEEP_EOL = new DateInterval('P28D');
 
-function format_interval($from, $to) {
+function format_interval($from, DateTime $to) {
```
localheinz, line 46: Alternatively, we could create `$to` in the function, but that would make it harder to test if we wanted to test it.

kamil-tekiela (approved):

PR #872, opened 2023-12-06
Enhancement: Extract `Version`

localheinz:
This pull request

- [x] extracts `Version` as a value object

Related to #871.

### DIFF
--- a/include/branches.inc
+++ b/include/branches.inc
@@ -1,6 +1,10 @@
 <?php
+
+require_once __DIR__ . '/../src/autoload.php';
 include_once __DIR__ . '/releases.inc';
 include_once __DIR__ . '/version.inc';
+
+use phpweb\Release\Version;
 
 /* Branch overrides. For situations where we've changed the exact dates for a
  * branch's active support and security fix EOLs, these can be reflected here.
@@ -85,13 +89,18 @@ function format_interval($from, DateTime $to) {
     return $eolPeriod;
 }
 
-function version_number_to_branch(string $version): ?string {
-    $parts = explode('.', $version);
-    if (count($parts) > 1) {
-        return "$parts[0].$parts[1]";
+function version_number_to_branch(string $value): ?string {
+    try {
+        $version = Version::fromString($value);
+    } catch (\InvalidArgumentException) {
+        return null;
     }
 
-    return null;
+    return sprintf(
+        '%s.%s',
+        $version->major()->toString(),
+        $version->minor()->toString(),
+    );
 }
 
 function get_all_branches() {
--- a/src/Release/Major.php
+++ b/src/Release/Major.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpweb\Release;
+
+final readonly class Major
+{
+    private function __construct(private string $value)
+    {
+    }
+
+    /**
+     * @throws \ValueError
+     */
+    public static function fromString(string $value): self
+    {
+        if (1 !== preg_match('/^(0|[1-9]\d*)$/', $value)) {
+            throw new \ValueError(\sprintf(
+                'Value "%s" does not appear to be a valid value for a major version.',
+                $value,
+            ));
+        }
+
+        return new self($value);
+    }
+
+    public function toString(): string
+    {
+        return $this->value;
+    }
+}
--- a/src/Release/Patch.php
+++ b/src/Release/Patch.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpweb\Release;
+
+final readonly class Patch
+{
+    private function __construct(private string $value)
+    {
+    }
+
+    /**
+     * @throws \ValueError
+     */
+    public static function fromString(string $value): self
+    {
+        if (1 !== preg_match('/^(0|[1-9]\d*)$/', $value)) {
+            throw new \ValueError(\sprintf(
+                'Value "%s" does not appear to be a valid value for a patch version.',
+                $value,
+            ));
+        }
+
+        return new self($value);
+    }
+
+    public function toString(): string
+    {
+        return $this->value;
+    }
+}
--- a/src/Release/Version.php
+++ b/src/Release/Version.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace phpweb\Release;
+
+final readonly class Version
+{
+    private function __construct(
+        private Major $major,
+        private Minor $minor,
+        private Patch $patch,
+    ) {
+    }
+
+    public static function create(
+        Major $major,
+        Minor $minor,
+        Patch $patch,
+    ): self {
+        return new self(
+            $major,
+            $minor,
+            $patch,
+        );
+    }
+
+    /**
+     * @throws \ValueError
+     */
+    public static function fromString(string $value): self
+    {
+        if (1 !== preg_match('/^(?P<major>(0|[1-9]\d*))\.(?P<minor>(0|[1-9]\d*))\.(?P<patch>(0|[1-9]\d*))$/', $value, $matches)) {
+            throw new \ValueError(\sprintf(
+                'Value "%s" does not appear to be a valid value for a version.',
+                $value,
+            ));
+        }
+
+        return new self(
+            Major::fromString($matches['major']),
+            Minor::fromString($matches['minor']),
+            Patch::fromString($matches['patch']),
+        );
+    }
+
+    public function major(): Major
+    {
+        return $this->major;
+    }
+
+    public function minor(): Minor
+    {
+        return $this->minor;
+    }
+
+    public function patch(): Patch
+    {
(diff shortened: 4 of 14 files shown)


src/Release/Patch.php
```diff
@@ -0,0 +1,32 @@
...
+     * @throws \InvalidArgumentException
+     */
+    public static function fromString(string $value): self
+    {
+        if (1 !== preg_match('/^(0|[1-9]\d*)$/', $value)) {
+            throw new \InvalidArgumentException(\sprintf(
+                'Value "%s" does not appear to be a valid value for a patch version.',
+                $value
+            ));
+        }
```
Girgias, line 23: Use ``ctype_digit()`` ?

src/Release/Version.php
```diff
@@ -0,0 +1,71 @@
...
+
+    public function minor(): Minor
+    {
+        return $this->minor;
+    }
+
+    public function patch(): Patch
+    {
+        return $this->patch;
+    }
```
Girgias, line 60: Do we really need those if we have readonly properties?

src/Release/Version.php
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * @throws \InvalidArgumentException
+     */
+    public static function fromString(string $value): self
+    {
+        if (1 !== preg_match('/^(?P<major>(0|[1-9]\d*))\.(?P<minor>(0|[1-9]\d*))\.(?P<patch>(0|[1-9]\d*))$/', $value, $matches)) {
+            throw new \InvalidArgumentException(\sprintf(
+                'Value "%s" does not appear to be a valid value for a version.',
+                $value
+            ));
```
Girgias, line 37:
```suggestion
            throw new \ValueError(\sprintf(
                'Value "%s" does not appear to be a valid value for a version.',
                $value
            ));
```
We could just throw a ValueError

Girgias (commented):
Just some comments after a quick glance

include/branches.inc
```diff
@@ -85,13 +89,18 @@ function format_interval($from, DateTime $to) {
     return $eolPeriod;
 }
 
-function version_number_to_branch(string $version): ?string {
-    $parts = explode('.', $version);
-    if (count($parts) > 1) {
-        return "$parts[0].$parts[1]";
+function version_number_to_branch(string $value): ?string {
+    try {
+        $version = Release\Version::fromString($value);
```
kamil-tekiela, line 94: Why not `$version = Version::fromString($value);` What am I missing?

src/Release/Version.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        return $this->minor;
+    }
+
+    public function patch(): Patch
+    {
+        return $this->patch;
+    }
+
+    public function toString(): string
```
kamil-tekiela, line 62: Why not just use the magic method `__toString`?

src/Release/Version.php
```diff
@@ -0,0 +1,71 @@
...
+        Major $major,
+        Minor $minor,
+        Patch $patch,
+    ): self {
+        return new self(
+            $major,
+            $minor,
+            $patch,
+        );
+    }
```
kamil-tekiela, line 26: Why not just make the constructor public?

tests/UserNotes/Release/Version/from-string-returns-version.phpt
```diff
@@ -0,0 +1,17 @@
+--TEST--
+Version::fromString() returns Version
+--FILE--
+<?php
+
+declare(strict_types=1);
+
+use phpweb\Release;
+
+require_once __DIR__ . '/../../../../src/autoload.php';
```
kamil-tekiela, line 10: Why are the tests in `UserNotes` directory?

src/Release/Version.php
localheinz: Depends on whether we want to stick with the internal implementation: using accessors instead of `public` properties has the advantage that we can change the internals any time without changing the consumers.

I can adjust, though.

src/Release/Version.php
localheinz: I prefer to explicitly invoke a method to cast a type to another type instead of relying on a magic method.

src/Release/Version.php
localheinz: Using one or more `public` named constructors and a single `private` constructor in conjunction with accessors instead of `public` properties allows us to change the internals of the class as needed.

From the call-site it does not matter whether someone invokes a method or accesses a property, but relying on a `public` property makes it harder to change internals.

For example, we could change the internals as follows and would not need to make any changes to call-sites:

```diff
diff --git a/src/Release/Version.php b/src/Release/Version.php
index 47975096..6b5664fd 100644
--- a/src/Release/Version.php
+++ b/src/Release/Version.php
@@ -6,11 +6,9 @@ namespace phpweb\Release;

 final readonly class Version
 {
-    private function __construct(
-        private Major $major,
-        private Minor $minor,
-        private Patch $patch,
-    ) {
+    private function __construct(private string $value)
+    {
+
     }

     public static function create(
@@ -18,11 +16,12 @@ final readonly class Version
         Minor $minor,
         Patch $patch,
     ): self {
-        return new self(
-            $major,
-            $minor,
-            $patch,
-        );
+        return new self(sprintf(
+            '%s.%s.%s',
+            $major->toString(),
+            $minor->toString(),
+            $patch->toString(),
+        ));
     }

     /**
@@ -37,35 +36,32 @@ final readonly class Version
             ));
         }

-        return new self(
-            Major::fromString($matches['major']),
-            Minor::fromString($matches['minor']),
-            Patch::fromString($matches['patch']),
-        );
+        return new self($value);
     }

     public function major(): Major
     {
-        return $this->major;
+        $parts = explode('.', $this->value);
+
+        return Major::fromString($parts[0]);
     }

     public function minor(): Minor
     {
-        return $this->minor;
+        $parts = explode('.', $this->value);
+
+        return Minor::fromString($parts[1]);
     }

     public function patch(): Patch
     {
-        return $this->patch;
+        $parts = explode('.', $this->value);
+
+        return Patch::fromString($parts[2]);
     }

     public function toString(): string
     {
-        return sprintf(
-            '%s.%s.%s',
-            $this->major->toString(),
-            $this->minor->toString(),
-            $this->patch->toString(),
-        );
+        return $this->value;
     }
 }
```

tests/UserNotes/Release/Major/from-string-rejects-invalid-value.phpt
```diff
@@ -0,0 +1,41 @@
...
+
+$invalidValues = array_filter($values, static function (string $value): bool {
+    try {
+        Release\Major::fromString($value);
+    } catch (\InvalidArgumentException) {
+        return true;
+    }
+
+    return false;
+});
```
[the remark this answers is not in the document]
localheinz: @Girgias @kamil-tekiela 

Do you have any suggestions how to work with data providers in `phpt` tests?

This is what I am trying to achieve here.

tests/UserNotes/Release/Version/from-string-returns-version.phpt
localheinz: Adjusted!

src/Release/Major.php
```diff
@@ -0,0 +1,32 @@
...
+            ));
+        }
+
+        return new self($value);
+    }
+
+    public function toString(): string
+    {
+        return $this->value;
+    }
```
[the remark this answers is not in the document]
localheinz: It's a matter of preference - when I allow constructing an object from a simple type, I typically use a named constructor `fromSimpleType()` and a corresponding method `toSimpleType()`. This allows to construct an object from and cast it to other simple types when necessary, and hides the internals.

I have written about [Naming constructors in PHP](https://localheinz.com/articles/2022/03/26/naming-constructors-in-php/), where I attempt to explain my reasoning.

Also see https://github.com/php/web-php/pull/872/files#r1418528779.

tests/UserNotes/Release/Version/from-string-rejects-invalid-value.phpt
```diff
@@ -0,0 +1,53 @@
...
+    try {
+        Release\Version::fromString($value);
+    } catch (\InvalidArgumentException) {
+        return true;
+    }
+
+    return false;
+});
+
+var_dump($invalidValues);
```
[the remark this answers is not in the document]
localheinz: @kamil-tekiela @Girgias 

I can adjust if you prefer!

include/branches.inc
kamil-tekiela: Thanks. I am rather against importing the namespace. IMHO every class should be imported unless it conflicts with another one. Only in case of conflicts, one should import namespace. Importing the full name makes for a cleaner code. The line is shorter, there's less visual bloat and my eyes can focus on what's really important: the name of the class. For the most part, I am not interested from which namespace the class comes. I can always hover over it to see the full namespace. The import list is maintained automatically by IDEs so I never look at it. If the import list is too long, it can always be collapsed using the alternative syntax.

PR #876, opened 2023-12-07
Fix: Remove emacs instruction

localheinz:
This pull request

- [x] removes what appear to be instruction for emacs to open a file in C++ mode

Spotted in #875.

💁‍♂️ For reference, see https://stackoverflow.com/questions/3312114/how-to-tell-emacs-to-open-h-file-in-c-mode.


sy-records (approved):

derickr (approved):
LGTM

PR #880, opened 2023-12-07
Fix: Remove vim instruction

localheinz:
This pull request

- [x] removes instructions for vim


sy-records (approved):

kamil-tekiela (approved):

localheinz:
Thank you, @kamil-tekiela and @sy-records!

PR #887, opened 2023-12-07
Fix: Use `mt_srand()` instead of `srand()`

localheinz:
This pull request

- [x] uses `mt_srand()` instead of `srand()`

Follows #699.
Somewhat related to #664.


localheinz:
Thank you, @kamil-tekiela!

PR #888, opened 2023-12-07
Fix: Remove dead code

localheinz:
This pull request

- [x] removes a few bits of dead code


localheinz:
Thank you, @saundefined!

PR #900, opened 2023-12-14
Enhancement: Run end-to-end tests asserting an HTTP response status code

localheinz:
This pull request

- [x] runs end-to-end tests asserting an HTTP response status code

Also see #605.

PR #901, opened 2023-12-15
Fix: Do not pass `null` to function that expects `string`

localheinz:
This pull request

- [x] stops passing `null` to a function that expects `string`

Spotted in #900.

💁‍♂️ In the first iteration of the inner `foreach`-loop, `$version` is `null`:

https://github.com/php/web-php/blob/e1c61689ade5ce4264e575a0a80fe03d349d670a/include/version.inc#L85-L89

### DIFF
--- a/include/version.inc
+++ b/include/version.inc
@@ -82,7 +82,7 @@ $RELEASES = (function () {
 function release_get_latest() {
     global $RELEASES;
 
-    $version = null;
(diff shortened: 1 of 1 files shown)


saundefined (approved):
Thanks!

localheinz:
Thank you, @saundefined!

PR #910, opened 2024-01-01
Add PHP Conference Kansai 2024.

Y-Kanoh:
Add PHP Conference Kansai 2024 in Japan.
https://2024.kphpug.jp

It is introducing of PHP Conference Kansai 2024 that is held on February 11th in Osaka Japan.
Please review and merge.


saundefined (approved):
Thanks!

PR #911, opened 2024-01-05
add: Drupal Mountain Camp 2024 conference

dan2k3k4:
[Drupal Mountain Camp 2024](https://drupalmountaincamp.ch/) will take place between March 7-10, 2024 in Davos, Switzerland.

We're branching out to focus on gathering more non-Drupal sessions such as PHP, Laravel, and Symfony as well as modern JS (React) and accessibility, inclusivity and diversity related talks.

This is why I figured this conference would be a good addition to the php.net community.


saundefined (approved):
Looks good, thanks!

PR #914, opened 2024-01-07
Add PHP Conference Fukuoka 2024

akase244:
Add PHP Conference Fukuoka 2024
https://phpcon.fukuoka.jp/2024/

It is introducing of "PHP Conference Fukuoka 2024" that is held on June 22, 2024.
Please review it.


saundefined (approved):
Thanks!

PR #925, opened 2024-01-15
Fix German translation of the PHP 8.3 release page

TimWolla:
The current translation is frankly terrible. It contains several spelling and grammar mistakes and it is very liberal in the translation, completely rephrasing (and dropping) important aspects of the English original. 

The updated translation matches the English original pretty closely. For the ext/random and Override stuff I was the original author of the English original.

/cc @msamesch @joshuaruesweg as the maintainers of doc-de. I've attempted to use the terminology that is used within the German documentation (e.g. Elternklasse instead of Basisklasse), I'd appreciate another pair of eyes :-)

-------------

see #902


TimWolla:
Note: https://www.php.net/releases/8.2/de.php / #903 is similarly terrible, but as it is not as prominent, given it's not the newest release, I did not yet bother fixing the translation. A short term fix might be simply deletion the German version for 8.2 / reverting #903.

saundefined:
I'm going to merge it, if there'll any improvement, let's do it in the new PR :)

Thanks a lot!

PR #934, opened 2024-02-01
Fix: Run `friendsofphp/php-cs-fixer` on `src/` and `tests/` only

localheinz:
This pull request

- [x] runs `friendsofphp/php-cs-fixer` on `src/` and `tests/` only


saundefined (approved):
Looks good, thank you!

localheinz:
Thank you, @saundefined!

PR #944, opened 2024-02-13
Enhancement: Collect and report code coverage from running unit tests

localheinz:
This pull request

- [x] collects and reports code coverage from running unit tests to https://about.codecov.io

PR #945, opened 2024-02-13
Fix: Provide token

localheinz:
This pull request

- [x] provides a token when reporting code coverage

Follows #944.

💁‍♂️ Needs a `CODECOV_TOKEN` secret to be proved, ideally for both

- https://github.com/php/web-php/settings/secrets/actions
- https://github.com/php/web-php/settings/secrets/dependabot

For reference, see 

- https://github.com/codecov/codecov-action?tab=readme-ov-file#v4-release 
- https://github.com/php/web-php/pull/945#issuecomment-1941704779

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -58,6 +58,7 @@ jobs:
         uses: "codecov/codecov-action@v4.0.1"
         with:
           files: ".build/phpunit/logs/clover.xml"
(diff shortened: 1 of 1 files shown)


saundefined:
@localheinz I guess we need to register on codecov.io, and install application for PHP org.
I don't have privileges to do that, and it's better to discuss this with Derick.

Until we do some preliminary setup, maybe disable this action for now?

localheinz:
@saundefined 

See #951.

PR #946, opened 2024-02-13
Fix: Use rolling tag

localheinz:
This pull request

- [x] uses a rolling tag for `codecov/codecov-action`

Follows #944.

💁‍♂️ My bad, wasn't aware they provide rolling tags.


saundefined (approved):
Thank you!

PR #955, opened 2024-02-17
Fix: Do not bother ignoring non-existent file

localheinz:
This pull request

- [x] stops ignoring a non-existent file

Follows #605.


saundefined (approved):
Thanks!

PR #958, opened 2024-02-23
Add PHPerKaigi 2024

muno92:
Add PHPerKaigi 2024
https://phperkaigi.jp/2024/

It is introducing of "PHPerKaigi 2024" that is held on Mar 7, 2024.
Please review it.


saundefined (approved):
Thank you!

PR #961, opened 2024-02-27
Remove section on edit.php.net

haszi:
Remove section on edit.php.net as edit.php.net has been unavailable for quite a while.


saundefined (approved):
Thanks!

PR #985, opened 2024-04-30
Update the end of release cycle based on approved RFC

drealecs:
Update the end of release cycle based on the recently approved RFC https://wiki.php.net/rfc/release_cycle_update


derickr:
Thanks!

PR #992, opened 2024-05-18
Add new PHP 8.4 gpg key branch

ericmann:
Waiting for keys from Saki and Calvin before merging.


NattyNarwhal (approved):
I think we're good signing wise.

PR #994, opened 2024-05-23
Add visual regression tests

saundefined:
/cc @pronskiy


pronskiy:
Going through the list of PRs and issues, we can see that there are a bunch of UI/CSS things. The problem is that we currently do not have any experts in this area, and moreover, no QA experts to verify if the changes do not break something.

With this PR Sergey and I propose to introduce screenshot regression testing.

It works like this:
- Creates snapshots of the website.
- Then for (almost) every PR creates a new screenshot after changes and compares it to a snapshot.
- If there is an unexpected difference between two screenshots, then this is a regression introduced and PR should be fixed.

We'd like your feedback on this idea. And if it makes sense, we'll implement the missing parts.

Girgias (commented):
I think this makes sense, as at least it is possible to see the difference between the current website render and the proposed change.

saundefined:
Example: https://github.com/saundefined/web-php/pull/21
Report: https://web-php-snapshot-pr-21.preview.thephp.foundation/

.github/workflows/pr-preview.yml
```diff
@@ -33,3 +33,98 @@ jobs:
...
+          comment-author: 'github-actions[bot]'
+
+      - uses: peter-evans/create-or-update-comment@v4
+        if: ${{ !cancelled() }}
+        with:
+          issue-number: ${{ github.event.number }}
+          comment-id: ${{ steps.snapshot.outputs.comment-id }}
+          edit-mode: 'replace'
+          body: |
+            🚀 Commit ${{ github.sha }} Regression report on https://web-php-regression-report-pr-${{ github.event.number }}.preview.thephp.foundation
```
derickr, line 130: Maybe use:
```
Regression report for commit ${{ github.sha }} is at https://web-php-regression-report-pr-${{ github.event.number }}.preview.thephp.foundation
```

.github/workflows/pr-preview.yml
```diff
@@ -33,3 +33,98 @@ jobs:
           edit-mode: 'replace'
           body: |
             🚀 Commit ${{ github.sha }} Deployed on https://web-php-pr-${{ github.event.number }}.preview.thephp.foundation
```
derickr, line 35: Maybe change this to:

```
Preview for commit ${{ github.sha }} can be found at https://web-php-pr-${{ github.event.number }}.preview.thephp.foundation
```

sy-records (approved):

saundefined:
I'm going to merge this PR later this week if there are no objections.
We'll see how useful it is in real life and adjust as needed.

PR #995, opened 2024-05-26
Create devcontainer.json

Graysonbarton:
Fix dependencies


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,5 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/universal:2",
```
saundefined, line 2: Thanks for the PR!

Shouldn't we use `mcr.microsoft.com/devcontainers/php:8.2` instead?
(I haven't used this file before, so it's just a question)

PR #1007, opened 2024-06-13
Fuzzy search of docs

Roy-Orbison:
Allow partial terms and multiple search tokens instead of rigid 'begins-with' substring matching.

This isn't perfect and could probably do with some tuning of the [FuzzySearch](https://github.com/jeancroy/FuzzySearch/blob/36e3ca507d0712af01f07a61f9734bfe22f281b6/dist/FuzzySearch.js#L27) parameters, but it's better than the current, very unforgiving, method.

It also adds the `.php` suffix to selected autocomplete results (to match the canonical tags), stops the results list from becoming inaccessible when taller than the viewport, and drops a couple of unused variables.

php quick search


Roy-Orbison:
Some inline styles still need to go into CSS.

Seeking feedback before marking PR as ready.

sy-records (approved):
LGTM.

PR #1008, opened 2024-06-14
Make replaceable class text italic

haszi:
Make replaceable class text italic to visually differentiate content that may or must be replaced.

@Girgias this is a minimal implementation of what I mentioned in https://github.com/php/doc-en/pull/3442.


saundefined (approved):
Thanks!

PR #1012, opened 2024-06-21
Add CascadiaPHP 2024 CFP entry

iansltx:


iansltx:
Should the above show the entry in the news list? Not seeing it and not sure whether that's a limitation of the preview environment or something broken with my code.

saundefined (approved):
Looks good, thanks!

PR #1017, opened 2024-06-22
Use HTTPS URL for event submission to main.php.net, match spam check expectation to web-master

### LINKED ISSUE #999: Event submission form error: "missing some parameters"
Hello everyone. My name is Bob Bloom. 

I am lead organizer of GTA PHP (Greater Toronto Area PHP), York Region PHP, and Laravel Toronto. In lieu of an actual website, I post stuff at https://github.com/local-toronto-php-groups. Our meetup.com links are there. If you are in Toronto this upcoming Thursday, join us at the heart of Toronto's financial district, at a fantastic corporate host. Larry Garfield is joining us remotely to talk about Property Hooks, new to PHP 8.4. 

Thank you to the PHP team for making Groups high visibility on PHP.net. I know there are too many monthly user group events to list on the home page. However, IMO, the link to a calendar of events is even better. 

Unfortunately, I am not able to contribute my Groups' events to this calendar. The reason is that I am get the "missing some parameters error".

image

There could be something very obvious that I am overlooking, and as a Canadian, let me say that I am sorry to trouble you about this. 

I did try contacting people via email whom I thought might be associated with PHP.net, or could recommend to me whom to contact. I was ultimately successful, and appreciative of the thoughtful reply. So I am here posting this issue as a result. 

Also, while I am here, is there a list of Groups on PHP.net? Something like what Go has at [https://go.dev/wiki/GoUserGroups](https://go.dev/wiki/GoUserGroups)? My thinking is that Toronto area PHP devs would go to PHP.net, so our group, and then click to our GitHub organization. 

Thank you for your help, it is much appreciated.

Regards from Toronto!
-Bob Bloom

iansltx:
Resolves #999 / GH-999

At some point in the past 12 years (this line was modified in the last 3 years but I doubt it got tested when modified) main.php.net started redirecting insecure HTTP to HTTPS, including for POSTs. The catch with those redirects is that POSTs won't get resubmitted when redirected, so when submitting an event the redirect would result in a GET with no parameters to the event submission endpoint, hence "Missing parameters." So event submission has been broken since main.php.net started redirecting HTTP to HTTPS.

Back in 2012 there was an attempt to switch this and other URLs to HTTPS, but it got rolled back because "there could be mirrors without ssl support." (see blame for the line this commit modifies). Since then, mirrors have been phased out, so we can safely assume we're calling HTTPS endpoints now (and that's the only way this will work anyway).

Verified by hitting the mentioned endpoint both on HTTP and HTTPS. HTTP gets redirected and fails due to missing parameters, HTTPS makes it through to the next step.

Additionally swaps the spam check value back to matching the web-master repo's expected value, as once the above was fixed that became the issue for calling the endpoint through this form. Optimizing for having only one repo needing to be fixed here rather than both this repo and the web-master one, hence not changing the expected antispam value there.

### DIFF
--- a/submit-event.php
+++ b/submit-event.php
@@ -110,7 +110,7 @@ if ($process) {
     }
 
     // Spam question
-    if ($_POST["sane"] != 4) {
+    if ($_POST["sane"] != 3) {
         $errors[] = "It's OK. I'm not real either";
     }
 
@@ -257,7 +257,7 @@ if ($process && count($errors) === 0) {
  </tr>
  <tr>
   <th class="subr">Are you real?</th>
-  <td><select name="sane"><?php display_options(["I, Robot", "I used to be", "WTF?", "No, but I'd still want to submit this", "Yes"], "2"); ?></select></td>
+  <td><select name="sane"><?php display_options(["I, Robot", "I used to be", "WTF?", "Yes", "No, but I'd still want to submit this"], "2"); ?></select></td>
  </tr>
 </table>
 </form>
--- a/submit-event.php
+++ b/submit-event.php
@@ -116,7 +116,7 @@ if ($process) {
 
     if (isset($_POST['action']) && $_POST['action'] === 'Submit' && empty($errors)) {
         // Submit to main.php.net
-        $result = posttohost("http://main.php.net/entry/event.php", $_POST);
+        $result = posttohost("https://main.php.net/entry/event.php", $_POST);
         if ($result) {
             $errors[] = "There was an error processing your submission: $result";
         }


iansltx:
Verified in the preview environment by @tiffany-taylor

tiffany-taylor:
I tested here https://web-php-pr-1017.preview.thephp.foundation/submit-event.php with dummy form data. First attempt, the [spam check in event.php](https://github.com/php/web-master/blob/main/public/entry/event.php#L37) had caught my manual test. @iansltx modified the spam check and array order in `submit-event.php`, which enabled me to reach the form submission success message on my second manual test. As far as I can tell from testing on my phone, it seems correct.

sy-records (approved):

saundefined (approved):
Thanks a lot!

PR #1019, opened 2024-06-23
Switch "today" marker on supported versions image from red to dark grey to differentiate from EOL versions

### LINKED ISSUE #1018: "Supported version" chart is unclear
The fact that the "today" dotted line has the same colour (red) as the EOL versions makes the entire chart unclear.

The dotted line and the date below should be drawn in a different colour, possibly neutral. I suggest to colour it in black.

Schermata del 2024-06-23 17-24-29

iansltx:
Resolves #1018 / GH-1018

### DIFF
--- a/images/supported-versions.php
+++ b/images/supported-versions.php
@@ -98,13 +98,13 @@ $height = $header_height + $footer_height + (count($branches) * $branch_height);
 			}
 
 			.today line {
-				stroke: #f33;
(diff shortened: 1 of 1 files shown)


saundefined (commented):
Thanks!

I have nothing against both red and gray colors,
so I'd listen to others' opinions.

PR #1022, opened 2024-06-26
Add CODEOWNERS

saundefined:
@pronskiy and I are suggesting to create a `CODEOWNERS` file
to automatically assign people to review PRs and triage issues.

We have a @php/web-team, but I'm not sure if all people on that team want to be notified every time,
and also it's easier to add people who want to be added via PR rather than team admin.

If anyone wants to add themselves to this file as well, feel free to add yourself :)
(but you must have write access to that repository)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @saundefined
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
 * @saundefined
+* @sy-records


derickr (commented):
LGTM

.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @saundefined
```
sy-records, line 1:
```suggestion
* @saundefined
* @sy-records
```

sy-records (commented):
You can add me 😀

Girgias:
I think the idea makes sense, I have people nudging me in private sometimes to do stuff about web when I don't really know the repo.

saundefined:
So, if there are no objections, I'll merge it later today.

If anyone with write permission would like to add themselves to this list,
they can do it themselves at any time.

PR #1026, opened 2024-07-05
Fix CODEOWNERS syntax error

sy-records:
Each line is a file pattern followed by one or more owners.

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file


saundefined (approved):
Thanks!

PR #1027, opened 2024-07-06
OG meta tags by default

### LINKED ISSUE #981: OG image is missing for some pages
When sharing pages like these: https://www.php.net/archive/2024.php on social media, there is image missing such as og:image and Twitter image src. Perhaps it could be inherited from the front page's meta image.

saundefined:
OG meta tags are available for all pages by default,
and can be overridden if needed (such as for PHP 8.0-8.3 release pages).

Should fix #981

### DIFF
--- a/include/layout.inc
+++ b/include/layout.inc
@@ -424,6 +424,9 @@ function site_header(string $title = 'Hypertext Preprocessor', array $config = [
 {
     global $MYSITE;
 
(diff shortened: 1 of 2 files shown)


sy-records (approved):
LGTM.

petk:
Nice. This looks very neat :+1: Thanks.

PR #1035, opened 2024-07-20
Remove docs.php.net, it's not user-visible any longer (closes #996)

### LINKED ISSUE #996: docs.php.net - still in use ?
As I was going through the PHP stubs of the Solr Extension, I found links to docs.php.net, which apparently is severely outdated, that says the latest php is 8.1 beta, and still has docs en, which I imagine could be as outdated. I'm gonna update the links to https://www.php.net/manual/*

From:
https://www.php.net/sites.php
Screenshot 2024-05-26 at 11 09 48

it also says "Documentation is built here four times a day."

Questions that popped shall it be named dev, if its dev. so no reliability expected ? or have redirector to php.net/manual ? or is it useful the way it is? or sites.php needs to get updated to reflect the current status?

I'm just wondering and thought to bring that to your attention,

Cheers,
Omar

jimwins:
If we do set up a site for previewing docs releases, it should probably have a more distinct name to avoid confusion with doc.php.net. Especially one that doesn't invite external links over the released docs (maybe docs-preview or docs-test or something).

### DIFF
--- a/sites.php
+++ b/sites.php
@@ -117,15 +117,6 @@ site_header("A Tourist's Guide", ["current" => "help"]);
  and contributors.
 </p>
 
-<h2 id="docs" class="content-header"><a href="http://docs.php.net/">docs.php.net</a>: Documentation dev server</h2>
-
-<p class="content-box">
- The documentation developmental server is a PHP mirror that contains upcoming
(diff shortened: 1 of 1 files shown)


saundefined (approved):
Thanks!

PR #1036, opened 2024-07-21
Trim mirroring docs

jimwins:
Just gets rid of old information that was only relevant for "official" mirrors for PHP.net, which aren't a thing any longer.


sy-records (approved):

saundefined (approved):
Looks good, thanks!

PR #1038, opened 2024-07-26
Link to doc.php.net and /my.php

jimwins:
Provide easier on-ramps!

Note that https://doc.php.net doesn't actually work quite yet (it's http-only), working with the systems team to get that fixed.

### DIFF
--- a/docs.php
+++ b/docs.php
@@ -17,8 +17,8 @@ site_header("Documentation", ["current" => "docs"]);
 <p>
  Note, that many languages are just under translation, and
  the untranslated parts are still in English. Also some translated
- parts might be outdated. The translation teams are open to
- contributions.
+ parts might be outdated. The translation teams are
+ <a href="https://doc.php.net/">open to contributions</a>.
 </p>
 </div>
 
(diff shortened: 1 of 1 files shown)


docs.php
```diff
@@ -17,8 +17,8 @@
 <p>
  Note, that many languages are just under translation, and
  the untranslated parts are still in English. Also some translated
- parts might be outdated. The translation teams are open to
- contributions.
+ parts might be outdated. The translation teams are
+ <a href="https://doc.php.net/">open to contributions</a>.
```
sy-records, line 21: Contributions should not need to be directed to doc.php.net?

jimwins: The idea here is to build out more information on doc.php.net about how to contribute (both to translations and the English version) but I think all we need on this page is a pointer over there so people can actually find it!

saundefined: After php/web-doc#49 merge, this could be a good starting point.

But I guess this PR should be delayed until https is working, or merge it now with http, and do another commit after configuring the certificate..

Girgias: Yeah, there isn't really an issue with it not being HTTPS, it is just a static website.

jimwins: Okay, I've changed it to just http: because doc.php.net is on a decrepit host for now and I'd rather not wait for that to be addressed.

saundefined (approved):
Thanks!

PR #1039, opened 2024-07-26
Purge references to PHP-GTK, it was abandoned

jimwins:


cmb69 (approved):
Indeed, removing references to gtk.php.net appears to be reasonable; after all, even the relaunch from almost 10 years ago didn't really take off.

sy-records (approved):

PR #1040, opened 2024-07-26
Remove some obsolete mailing lists

jimwins:


cmb69 (commented):
This looks good to me, but maybe @derickr will have a look.

saundefined (approved):
I suggest we wait until the end of the week, if there are no objections, we'll merge it.

Also, I'm not sure if we should keep the partition «Non-English language mailing lists»,
I guess we should either add the other languages or remove all of them :)

derickr:
@saundefined I don't always manage to read things before "the end of the week", but this looks OK to me too.

PR #1052, opened 2024-08-21
Validate all of the XML files in a GitHub Action

jimwins:
Adding to the "Coding Standards" action because it seemed to fit there instead of running a whole new job

### DIFF
--- a/archive/2007.xml
+++ b/archive/2007.xml
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
-blorp
 <feed xmlns="http://www.w3.org/2005/Atom" xmlns:php="http://php.net/ns/news">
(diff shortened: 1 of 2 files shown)


jimwins:
Will leave this broken so reviewers can see it works. [6d984a8](https://github.com/php/web-php/pull/1052/commits/6d984a82c61d8dc240b216dd5c62657d4c58d448) should be reverted before merging.

archive/2007.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
+blorp
```
sy-records, line 2: LGTM.

```suggestion

```

sy-records (approved):

PR #1055, opened 2024-08-27
API Platform Con 2024

dunglas:
Announcement for [the API Platform Conference](https://api-platform.com/con).


saundefined (approved):
Thanks!

PR #1057, opened 2024-09-01
Add function for translation of UI elements

haszi:
Add a function similar to the one in PhD that can be used to translate words or phrases, and use it to translate UI elements of the doc pages. If translation of a certain text is not defined for a language, the function will use the original English text.

Closes https://github.com/php/phd/issues/145

As noted in [this comment](https://github.com/php/phd/issues/145#issuecomment-2308983253) (by @derickr ?) the "User Contributed Notes" related parts of the UI may have not been translated on purpose.

Pinging @leonardolara and @cmb69 from the original [issue](https://github.com/php/phd/issues/145).


haszi:
> 🚀 Regression report for commit [3ced7a0](https://github.com/php/web-php/commit/3ced7a0f797188f89e1328c7ab7dfa30ae49e664) is at https://web-php-regression-report-pr-1057.preview.thephp.foundation

I don't quite know what to make of this test failure and visual diff in the above link as both pages seem the same to me. If I understand this correctly, they're off by 40 pixels somewhere. Any pointers on what to do with this would be welcome.

saundefined:
> I don't quite know what to make of this test failure and visual diff in the above link as both pages seem the same to me. If I understand this correctly, they're off by 40 pixels somewhere. Any pointers on what to do with this would be welcome.

really weird, the `.hero-versions` block was a bit shorter than in the test sample.
I'll try this case locally, maybe we'll just make the tests less accurate by a few pixels

>https://web-php-pr-1057.preview.thephp.foundation/

since the test environment only uses English, maybe make some other language as default to see how it works,
and then revert to English?

saundefined:
@haszi regression test fail fixed in 0361d004178272ed7b4ed53e8bd86eb8cece3abc.
Thanks!

include/ui_translation/ru.ini
```diff
@@ -0,0 +1,9 @@
+change_language = ""
+improve_this_page = ""
+how_to_improve_this_page = ""
+contribution_guidlines_on_github = ""
+submit_a_pull_request = ""
+report_a_bug = ""
+add_a_note = ""
+user_contributed_notes = ""
+no_user_notes = ""
```
saundefined, line 9: @mmalferov please take a look,
you might want to correct something :)

```suggestion
change_language = "Изменить язык"
improve_this_page = "Улучшить страницу"
how_to_improve_this_page = "Посмотреть инструкцию"
contribution_guidlines_on_github = "Посмотритеть рекомендации по внесению вклада на GitHub"
submit_a_pull_request = "Прислать исправление"
report_a_bug = "Сообщить об ошибке"
add_a_note = "Добавить примечание"
user_contributed_notes = "Примечания пользователей"
no_user_notes = "На этой странице нет примечаний, добавленных пользователями."
```

<details><summary>Example</summary>

image

</details>

mmalferov: Наверное, это влияние каких-то книг об оформлении страниц, форм и ссылок. Почему-то оставлять подписи с ответом на вопрос «Что сделать?» хочется кнопкам, а ссылкам я бы добавил подписи с ответом на вопрос «Что?»:

```suggestion
change_language = "Язык"
improve_this_page = "Улучшение страницы"
how_to_improve_this_page = "Инструкция"
contribution_guidlines_on_github = "Github-инструкция по улучшению страниц"
submit_a_pull_request = "Исправление"
report_a_bug = "Сообщение об ошибке"
add_a_note = "Добавить примечание"
user_contributed_notes = "Примечания пользователей"
no_user_notes = "Пользователи ещё не добавляли примечания для страницы"
```

suggest

P. S. Is a 'guidlines' (without 'e' that follows the 'd') correct? The slug `contribution_guidlines_on_github` should be the `contribution_guidelines_on_github` maybe?

saundefined: Looks good, thanks!

include/shared-manual.inc
```diff
@@ -59,17 +62,19 @@ function manual_notes($notes):void {
         $noteCountHtml = "<span class=\"count\">$num_notes note" . ($num_notes == 1 ? '' : 's') . "</span>";
```
leonardolara, line 62: Maybe
`$noteCountHtml = "<span class=\"count\">$num_notes " . ($num_notes == 1 ? $note : $notes) . "</span>";`
to allow translation of the number of notes beside the title?
```
$note = 'note';
$notes = 'notes';
```

cmb69: Yeah, that would be much better! Appending an "s" works for few languages. However, pluralization is more difficult than distinguishing between singular and plural (@mvorisek can probably confirm for the Czech language), but it might be okay for the currently active languages (and we could still extend).

mvorisek: I can confirm appending `s` to make words plular is NOT how Czech language works. We change the word ending based on many rules.

saundefined: Or use the more universal format «notes: N»,

because, for example, in Russian (I guess it's not the only language),
different endings will be for 1, 3, and 5 notes.

cmb69: > because, for example, in Russian (I guess it's not the only language),
> different endings will be for 1, 3, and 5 notes.

That's what I was referring to regarding the Czech language. :)

jimwins: This is why `gettext` has [`ngettext()`](https://www.php.net/function.ngettext.php), but maybe using that would be overkill for this issue. If there was interest in doing more widespread internationalization of the PHP website, that could be a way to go.

haszi: > Maybe `$noteCountHtml = "<span class=\"count\">$num_notes " . ($num_notes == 1 ? $note : $notes) . "</span>";` to allow translation of the number of notes beside the title?

I think that doing this is a good idea but I would prefer not to do it as part of this PR. The changes in this PR address the simpler issue of translating the static elements of the documentation UI. Looking at the above discussion on the intricacies of translating the number of notes to various languages (and we haven't even heard from the Chinese or Japanese doc maintainers on the subject), I think all that needs a bit more thought put into it and should be split into its own PR.

So my suggestion is for this PR to only address the translation of the static elements of the doc pages and to open a new PR for the dynamic elements. Maybe even open an issue first and ask maintainers of every language repo to discuss how the dynamic elements (the total note count, the number of upvotes/downvotes and how long ago a note was submitted) can be translated to each of their respective languages and open a PR once it is clear how to adequately implement it for each of those languages.

haszi:
If merging this PR with the translation of only the static UI elements is acceptable, should the actual translations be done as part of this PR (as was done with the Russian translation by @saundefined and @mmalferov) or as separate PRs?

cmb69:
> If translation of a certain text is not defined for a language, the function will use the original English text.

In my opinion, this is good enough.

haszi:
Rebased the PR so it can be merged which, if there are no objections, I'd like to do around the end of next week.

Issue #1058, opened 2024-09-02
sha265sum of 8.3.11 is incorrect on downloads.php


patrickpoortman:
On the  page https://www.php.net/downloads.php

It currently lists
[php-8.3.11.tar.gz](https://www.php.net/distributions/php-8.3.11.tar.gz) [(sig)](https://www.php.net/distributions/php-8.3.11.tar.gz.asc) [19,330Kb] 29 Aug 2024 b93a69af83a1302543789408194bd1ae9829e116e784d578778200f20f1b72d40

The checksum has a trailing zero that is not supposed to be there, the cecksum is;
`sha256:b93a69af83a1302543789408194bd1ae9829e116e784d578778200f20f1b72d4`

https://github.com/php/web-php/blob/0361d004178272ed7b4ed53e8bd86eb8cece3abc/include/version.inc#L27

PR #1059, opened 2024-09-02
Correct tar.gz sha256 checksum in version.inc

patrickpoortman:
The published checksum contains a trailing 0.
fix https://github.com/php/web-php/issues/1058

Removed the trailing `0`


sy-records (approved):
Thanks!

saundefined (approved):
Good catch, thanks!

PR #1067, opened 2024-09-12
Fix comment overwriting

sy-records:
see https://github.com/php/web-php/pull/1067#issuecomment-2345077976

`Preview` and `Regression report` in one comment.

The comment is overwritten.


saundefined (approved):
Thanks!

PR #1071, opened 2024-09-15
Replace MYPHPNET global var with the UserPreferences class

MauricioFauth:
- Replaces the `$MYPHPNET` global variable with static properties of the `phpweb\UserPreferences` class
- Moves the `MYPHPNET_URL_*` global constants to the `phpweb\UserPreferences` class
- Removes the `src/autoload.php` include in `shared-manual.inc` as it is already included in `prepend.inc`

Using static properties for this is not great, but it is an improvement over the global variable.

### DIFF
--- a/include/prepend.inc
+++ b/include/prepend.inc
@@ -101,7 +101,7 @@ include __DIR__ . '/last_updated.inc';
 function myphpnet_load(): void
 {
     UserPreferences::$languageCode = '';
-    UserPreferences::$searchType = MYPHPNET_URL_NONE;
+    UserPreferences::$searchType = UserPreferences::URL_NONE;
     UserPreferences::$isUserGroupTipsEnabled = false;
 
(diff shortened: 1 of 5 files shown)


include/prepend.inc
```diff
@@ -96,91 +100,83 @@ include __DIR__ . '/last_updated.inc';
...
-    }
-
-    if (isset($MYPHPNET[3])) {
-        return $MYPHPNET[3];
+    if (is_bool($enable)) {
+        UserPreferences::$isUserGroupTipsEnabled = $enable;
     }
 
+    // Show the ug tips to lucky few, depending on time.
     if ($_SERVER["REQUEST_TIME"] % 10) {
```
MauricioFauth, line 154: Reference: https://github.com/php/web-php/commit/46ce29d9d2231acdc038b24a3afc280a6e613ff3

saundefined (approved):
Looks good, thanks!

sy-records (approved):

PR #1072, opened 2024-09-15
Remove `LAST_LANG` cookie as it is not used anymore.

takaram:
The "last seen language" is no longer respected since 4 years ago.
So we should stop setting the `LAST_LANG` cookie.

See 7691b44067db646287064df92ad6bd10718ba9a6 and https://bugs.php.net/bug.php?id=79524


sy-records (approved):
LGTM.

Thank you!

saundefined (approved):
Looks good, thanks!

PR #1075, opened 2024-09-18
Move myphpnet_* functions to the UserPreferences class

MauricioFauth:
Move all `myphpnet_*` functions to the `UserPreferences` class and changes its static properties to instance properties.


saundefined (approved):
Thanks for resolving the conflicts!

LGTM, we can wait for the opinions of the others,
if there are no objections, we will merge it

cmb69:
In my opinion, a very good change, especially since we now have more code under test.

PR #1076, opened 2024-09-18
Link to installation documentation and Windows downloads

jimwins:


downloads.php
```diff
@@ -39,6 +39,30 @@
...
+    Microsoft Windows</a>. The PHP project does not currently release binary packages
+    for other platforms such as Linux or macOS, but they are packaged by distributions
+    and other providers. For more information, see:
+
+    <ul>
+        <li>
+            <a href="/manual/en/install.unix.php">Installation instructions for Unix systems</a>
+        </li>
+        <li>
+            <a href="/manual/en/install.macosx.php">Installation instructions for macOS</a>
```
cmb69, line 60: Maybe directly link to https://www.php.net/manual/install.macosx.packages.php?

downloads.php
```diff
@@ -39,6 +39,30 @@
...
+
+<p>
+    <a href="https://windows.php.net/download/">Binaries are available for
+    Microsoft Windows</a>. The PHP project does not currently release binary packages
+    for other platforms such as Linux or macOS, but they are packaged by distributions
+    and other providers. For more information, see:
+
+    <ul>
+        <li>
+            <a href="/manual/en/install.unix.php">Installation instructions for Unix systems</a>
```
cmb69, line 57: Most of the installation instructions for Unix refer to manual builds. I think only https://www.php.net/manual/en/install.unix.debian.php is about packages. Besides that page needs an overhaul, we should at least have a respective page about Fedora/RHEL packages (`yum`). Maybe @remicollet is interested in this?

cmb69 (commented):
Thank you! This makes generally sense to me.

downloads.php
```diff
@@ -39,6 +39,30 @@
     ],
 );
 ?>
+<p>
+    <a href="/manual/en/install.general.php">Installing PHP</a> is covered
+    thoroughly in the PHP documentation.
+</p>
```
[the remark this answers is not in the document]
jimwins: Pushed a fix for both of these issues.

downloads.php
jimwins: I think we want to link to the install page for macOS and Unix instead of direct to the packages, just to not push users too deep into the weeds. Those pages in the documentation should be cleaned up to put information about downloading binaries/packages before instructions on building from source.

downloads.php
```diff
@@ -39,6 +39,32 @@
...
+
+<p>
+    <a href="/manual/install.general.php">Installing PHP</a> is covered
+    thoroughly in the PHP documentation.
+</p>
+
+<h2>Binaries</h2>
+
+<p>
+    <a href="https://windows.php.net/download/">Binaries are available for
```
NattyNarwhal, line 52: Should probably remove the Windows download links below if we link it up here.

downloads.php
```diff
@@ -39,6 +39,32 @@
...
+    Microsoft Windows</a>. The PHP project does not currently release binary packages
+    for other platforms such as Linux or macOS, but they are packaged by distributions
+    and other providers. For more information, see:
+
+    <ul>
+        <li>
+            <a href="/manual/install.unix.php">Installation instructions for Unix systems</a>
```
NattyNarwhal, line 59: This is for building from source and should be put below, and probably made more obvious on the linked page itself that it's for building from source and not likely the user's first choice.

downloads.php
```diff
@@ -39,6 +39,32 @@
...
+    thoroughly in the PHP documentation.
+</p>
+
+<h2>Binaries</h2>
+
+<p>
+    <a href="https://windows.php.net/download/">Binaries are available for
+    Microsoft Windows</a>. The PHP project does not currently release binary packages
+    for other platforms such as Linux or macOS, but they are packaged by distributions
+    and other providers. For more information, see:
```
NattyNarwhal, line 55: We should be mentioning them here perhaps?

NattyNarwhal (approved):

downloads.php
jimwins: Yeah, we can at least have a unix.yum that corresponds to unix.debian without much controversy, I hope, and then actually link to the more comprehensive third-party repositories if we decide to start calling out the third-party things.

NattyNarwhal:
Forgot to check for other reviews before I hit send; I see Christoph has covered a lot of the same points I have at least.

downloads.php
jimwins: I'm not worried about the redundancy, and the later links go to the specific `#id` for the corresponding version.

downloads.php
jimwins: Yeah, see my response to @cmb69 earlier. This part of the documentation should be re-jiggered a bit to be more binary/package forward.

downloads.php
jimwins: I'm not sure what you mean by "them".

NattyNarwhal: The distributions for PHP from third parties.

jimwins:
Started working on the documentation side of this here: https://github.com/php/doc-en/pull/3763

downloads.php
jimwins: I don't think we want to link from this page to those, but I hope we can get agreement to start putting some of that sort of information in the documentation.

jimwins:
This was inspired by [this GitHub issue](https://github.com/php/php-src/issues/15916) and related discussion on the Foundation Slack channel.

jimwins:
Documentation changes landed, so merging this.

PR #1078, opened 2024-09-19
Link to GH issues instead of bugs.php.net for documentation issues

jimwins:
Make sure we're linking to the GH issue tracker instead of bugs.php.net.

(Also one little layout tweak that was bugging me.)


sy-records (approved):
LGTM

PR #1081, opened 2024-09-23
Move `language_choose_code()` into new `LangChooser` class

takaram:
This PR introduces a new class, `LangChooser`, replacing the `language_choose_code()` function.

The goal is to make `/releases/8.3/index.php` respect users' language preferences, similar to the manual pages. I initially considered reusing `language_choose_code()`, but its reliance on global variables made that challenging. As a result, I refactored the logic into a new class to simplify reuse in the `/releases` pages.

A follow-up PR to update `/releases/8.3/index.php` will be submitted once this PR is merged.


saundefined (commented):
At first glance, looks good, thanks!
I'll take another closer look soon, and maybe someone else will have some comments to make.

saundefined (approved):
Thanks!
If there's no objection, I'll merge it later today.

sy-records (approved):

takaram:
Thank you!

PR #602, opened 2022-07-02
Open for discussion: design updates

mvriel:
As a design experiment, I have made a re-imagining of the PHP website's visual design. After posting that on Twitter (https://twitter.com/mvriel/status/1542898919142789123) I received thus much support that I opened a topic on the mailing list to see if the webmaster team is interested as well (https://news-web.php.net/php.webmaster/30216).

However, I was having fun with the design and wondered how much effort it would be to apply the changes to the website. This Pull Request can be considered a basis for discussion, to demonstrate what is needed and to include screenshots for the team to see what it looks like.

As a side note: I am well aware that the conclusion of this adventure may be that the presented work will not be included. This is fine. I am more than happy to have the conversation and it would be a shame to let this merely rot on my harddisk :)

At the time of writing, the changes have been applied to the homepage alone and look like this,

**Desktop**

image

**Mobile**

PR #1084, opened 2024-10-04
Update navbar design and improve search UI

lhsazevedo:
> [!TIP]
> You can view this proposal live at: https://php-navbar.lhsazevedo.dev/

## Intro
Over the years, the PHP webpage received various design proposals, some didn't gain traction due to their drastic nature or departure from PHP's established branding style. However, our community has shown that we can successfully implement design improvements when they're incremental – the PHP 8 [release pages](https://github.com/php/web-php/pull/350), [homepage hero](https://github.com/php/web-php/pull/459) and [thanks page](https://github.com/php/web-php/pull/684) are great examples of this approach.

Inspired by these successful updates and the the discussions they sparked, I'd like to propose some updates to the navbar design and introduce a new search dialog to enhance a bit the user experience on php.net. These changes are inspired by modern web design principles while respecting PHP's branding and build upon [previous community discussions](https://github.com/php/web-php/pull/602).

By focusing on small enhancements rather than a full redesign, we can improve the site's usability and aesthetics without disrupting its familiar layout and branding.

## UI Scope
It's worth noting that while the current client-side search implementation could be considered outdated, replacing it would be beyond the scope of this design improvement. Such a change would require a separate, focused effort. This proposal aims to enhance the user interface while maintaining the existing search functionality, setting the stage for potential future improvements to the search implementation itself.

## Enhancements

1. **Modernized navbar design**:
   - A responsive navbar with improved mobile navigation.
   - Ensures a consistent experience across devices, addressing the growing mobile user base while maintaining PHP's established visual identity.

2. **New search dialog**:
   - A user-friendly search modal with keyboard navigation.
   - Improves accessibility with proper ARIA attributes and keyboard navigation, making php.net more inclusive.

3. **Retained search implementation**:
   - Integration of the existing FuzzySearch with the new UI.
   - Maintains familiar search behavior while setting the stage for future improvements to the search implementation.

4. **Code improvements**
   - CSS refactoring and removal of outdated libraries (Hogan.js, typeahead.js).
   - Improves maintainability, making future enhancements easier to implement.

5. **Enhanced reliability**
   - Introduction of Playwright E2E tests for the new search modal.
   - Ensures the changes don't introduce regressions, maintaining the site's reliability.

## Impact
These changes are designed to be minimally disruptive:
- Existing users will find a familiar but enhanced experience
- New users will benefit from improved navigation and search UI

## Preview
Please note that the preview site uses an improved search index generated by a companion change I'm proposing to the PHP documentation generator (phd). While both PRs can be merged independently, I've captured the screenshots and recordings using the new index to showcase the full potential of these improvements.

You can find the companion PR for the phd changes here: https://github.com/php/phd/pull/154
### Preview Site
You can view this proposal live at: https://php-navbar.lhsazevedo.dev/

### Desktop preview

https://github.com/user-attachments/assets/8137fb65-1e90-4b93-a185-708aa766e79b

### Mobile preview

https://github.com/user-attachments/assets/ad25d7e4-8ad2-4ca6-99a5-f92c49266d75

## Comparison
<details>
  <summary>View GIF comparisons (image heavy)</summary>

#### Navbar
desktop
#### Search modal dialog
desktop_search

#### Navbar
mobile
#### Navigation
mobile_nav
#### Search
mobile_search

</details>

## Final thoughts
I believe these improvements will benefit the PHP community while respecting the site's established design principles. I look forward to your feedback and collaboration in refining this proposal.

If you find this proposal valuable or have any concerns, I'd greatly appreciate your feedback. Please consider leaving a 👍 or 👎 to help measure the community opinion.


lhsazevedo:
Suggested reviewers: @morrisonlevi @dragoonis @cmb69 @kamil-tekiela @ramsey  @mvriel

kamil-tekiela:
I like it. I like the changed color, I like that the text of the buttons is more readable, I like that the PHP logo is now white, and I like that the search results list is simpler. There are a couple of things I don't like though. The mobile view looks really good.

1. I am not of a fan of the search opening up to cover the page. Why is it that when I click on an input box a new input box pops up? The placeholder text also doesn't match suggesting that the popup does something different than the box I actually clicked on. 
2. If the popup stays, I would not disable the page scroll. I never liked this as it creates a jarring visual experience. At least on Chrome the page shifts to the right and the scroll bar is now in the middle box which means that if I want to scroll I need to hover my mouse over the thin box. 
3. If the popup stays, there should be no load animation. It's an unnecessary gimmick that only delays loading of the box. 
4. Even though the text is more readable, I still think it's not contrast enough from the background color. Especially the "Search" placeholder is visually blending in with the surroundings. Make it stand out more, either by changing the text color or giving the search box lighter border. I wouldn't mind if all text in the navbar was pure white or whitesmoke.
5. IMHO there is no need to make the navbar taller. The previous height was ok. On most monitors the vertical space is more expensive than horizontal, so every padding you add makes the actual content push down to a place with less priority. You made it shorter in mobile view so why make it taller in desktop view.
6. Not sure if related but the code formatting on `/releases/8.3/en.php` is borked.
7. The navbar is now almost the same color as the main div of `/releases/8.3/en.php`. Everything is purple now. I don't hate the color, but it's just too much. Something needs to change.
8. The logograms in mobile view have too low contrast with the background. You made them `#D2D9FF` but even if they were `#FFFFFF` they would be still too dark. I would suggest making it `#F5F5F5` with opacity 1, the same as all text in the navbar.

cmb69:
Thank you! Overall, this looks like a nice improvement, but I don't like that the search now won't work if JavasScript is not available.

simonrjones:
The popup search results looks very smart. A few thoughts:

Accessibility:
- Contrast of primary navigation items fails WCAG accessibility (so does the current site it seems), text colour #DBE2FF or lighter would fix this
- Contrast of search text fails WCAG accessibility, text colour #CBCCF1 or lighter would fix this
- ALT text of PHP logo is "PHP Logo" - there's no need for the text "logo" here for accessibility reasons. Ideally it would be "PHP home" to indicate it links to the homepage.

Other:
- I'd strongly recommend using progressive enhancement so the search works without JS. JS can fail for many reasons and the search should be accessible to all. The search box is a button, if this switched back to input then the form would work without JS (since would submit to the search results page)
- When you do just press return it loads this full website search results page instead of the manual lookup page. It's probably more useful, but the styling is very basic on the full website search results page (I appreciate this is the same on the current site). Compare:
   - https://www.php.net/manual-lookup.php?pattern=function&scope=quickref
   - https://php-navbar.lhsazevedo.dev/search.php?lang=en&q=function

include/footer.inc
```diff
@@ -108,5 +108,51 @@ if (!empty($_SERVER['BASE_PAGE'])
...
+    class="search-modal"
+  >
+    <div class="search-modal-header">
+      <div class="search-modal-form">
+        <div class="search-modal-input-icon">
+          <!-- https://feathericons.com search -->
+          <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><circle cx="11" cy="11" r="8"></circle><line x1="21" y1="21" x2="16.65" y2="16.65"></line></svg>
+        </div>
+        <input
+          type="text"
```
Girgias, line 125: Why not the correct type?
```suggestion
          type="search"
```

include/footer.inc
```diff
@@ -108,5 +108,51 @@ if (!empty($_SERVER['BASE_PAGE'])
...
+      aria-label="Search results"
+      id="search-modal-results"
+      class="search-modal-results"
+    ></div>
+    <div class="search-modal-helper-text">
+      <div>
+        <span class="search-modal-helper-key">↑ ↓</span> to navigate •
+        <span class="search-modal-helper-key">Enter</span> to select •
+        <span class="search-modal-helper-key">Esc</span> to close
+      </div>
```
Girgias, line 148: Use the relevant HTML `<kbd>` tag.

include/footer.inc
```diff
@@ -108,5 +108,51 @@ if (!empty($_SERVER['BASE_PAGE'])
...
+      <div>
+        Press <span class="search-modal-helper-key">Enter</span> without
+        selection to search using Google
```
Girgias, line 151: Ditto kdb tag.

include/footer.inc
```diff
@@ -108,5 +108,51 @@ if (!empty($_SERVER['BASE_PAGE'])
 
 <a id="toTop" href="javascript:;"><span id="toTopHover"></span><img width="40" height="40" alt="To Top" src="/images/to-top@2x.png"></a>
 
+<div id="search-modal-backdrop" class="search-modal-backdrop">
+  <div
+    role="dialog"
+    aria-label="Search modal"
+    id="search-modal"
+    class="search-modal"
+  >
```
Girgias, line 117: Indent here is off.

Girgias (changes requested):
The search bar not working any more with JS is a massive issue.

lhsazevedo:
Thanks for the quick review and feedback! I've addressed many of your points and would like to discuss a few of them:

### Progressive enhancement
@cmb69 @Girgias @simonrjones
The search was updated to work without JavaScript, using a form by default with the modal as a progressive enhancement.

Searching on mobile sized screens will still require JS. Given that it's common for mobile users to have JavaScript enabled, I think this is a reasonable trade-off. Some major language require JavaScript for searching (even on desktop): [Kotlin](https://kotlinlang.org/docs/home.html), [Rust](https://doc.rust-lang.org/book/), [Swift](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/), [MDN](developer.mozilla.org), [Dart](https://dart.dev/). All Algolia-powered docs as well, and MkDocs.

### Accessibility and contrast
@kamil-tekiela @simonrjones

All elements in the navbar and search modal have been updated to comply with WCAG AA contrast ratios. This includes the primary navigation items, search text, and icons in mobile view.

### Layout
#### Page covered, no scroll
@kamil-tekiela The popup is inspired by Algolia that used in popular documentation sites (e.g. [Composer](https://getcomposer.org/doc/), [PHPStan](https://phpstan.org/), [Laravel](https://laravel.com/), [Vue](https://vuejs.org/), [React](https://react.dev/) etc). It felt it a bit awkward the first time I saw it, but got used to it eventually.

The ideal UI for me would be something like [Psalm docs](https://psalm.dev/docs/) (Material for MkDocs), but it is a bit more complicated to implement. Given the current search UI situation (5 different scrolling elements nested in a thin menu) maybe we could proceed with the Algolia-like modal as a first step and further improve it in the future. Would you consent to this approach?

#### Fade animation
@kamil-tekiela I believe this is a bit subjective and a matter of taste, but I can remove it if it's a common opinion. For now I halved the animation duration to 1/10th of a second, is this still a deal breaker? You can test it [here](https://php-navbar.lhsazevedo.dev/) (may need to disable cache).

#### Navbar height
@kamil-tekiela Forgot to mention but I followed @morrisonlevi's suggestion to [make the navbar static](https://github.com/php/web-php/pull/602#issuecomment-1228911015), so now it doesn't take up any space after scrolling down. Note that the navbar is 64px tall on both desktop and mobile, which seems a common height for navbars, even sticky ones.

### Other fixes
- Matched the placeholder text between the search button and the modal.
- Darkened the release page background a bit for better contrast with the navbar.
- Hid navbar logo from the accessibility tree, as the anchor aria-label is already "PHP home".

### Unrelated or out of scope
#### Release page formatting
@kamil-tekiela I think this is now fixed https://github.com/php/phd/pull/156

#### Search page styling
@simonrjones Given that this is the current behavior and is related to the page content itself, I think it would be out of scope for this PR. What do you think of discussing this in a separate issue?

include/footer.inc
lhsazevedo: Fixed

include/footer.inc
lhsazevedo: Done

include/footer.inc
lhsazevedo: Hm, I can't spot the wrong indent... Would you mind suggesting the fix?
(BTW this was formatted using Prettier)

kamil-tekiela:
Thanks, good work! 

The animation speed is better and less noticeable now. The height is just my opinion and if you think it looks better taller than let's leave it. I didn't even notice it's not sticky anymore. 

The colors have improved, but I would still like to see the text stand out more from the background. A tiny bit lighter would suffice.

Girgias:
I still don't see why you require JS on mobile.

I know people that disable JS on desktop and on mobile. Surely it can "just" be a search bar that has nothing dynamic like it currently does on mobile.

Moreover, I don't see how us using something like [MeiliSearch](https://github.com/meilisearch/meilisearch) in the future should require JS. Other websites might do that, and I argue that they are wrong.
The fact I can't even *load* the Kotlin docs without JS is *bonkers*, the MDN search looks like it should work but just leaves you on a blank page (which is also jank). At least the Rust docs don't *offer* you the option to search if JS is disable.

lhsazevedo:
@Girgias 
> I still don't see why you require JS on mobile.

The navigation menu itself now requires JavaScript to function (open/close) on small screens. If we decide to support users with JS disabled on such devices, it’s not just the search that needs a fallback—the entire navigation menu would need one too.

Given that various doc pages, and even governmental portals don’t prioritize JS-disabled support on mobile ([gov.ie](https://www.gov.ie/en/), [gov.sg](https://www.gov.sg/), [denmark.dk](https://denmark.dk/), [usa.gov](https://www.usa.gov/), [europa.eu](https://european-union.europa.eu/index_en)), I initially viewed this as an acceptable compromise.

However, if this is key concern, I’d suggest following the smart approach used by [gov.uk](https://www.gov.uk/), where both the menu and search buttons default to links (/browse and /search) when JS is disabled, and are dynamically enhanced when JS is available. This avoids hacks needed to simulate the enhanced experience without JS, while simplifying the overall implementation.

If this sounds good, I could work on it this week.

@kamil-tekiela Ok, will do :)

Girgias:
Yes the `gov.uk` approach is the one I'd like.

There is a reason many countries use their template, it is because (for once) the UK gov does something actually right.

simonrjones:
I agree supporting non JS is a better and more reliable solution. 

This is a good page on GOV.UK on building a robust front end: https://www.gov.uk/service-manual/technology/using-progressive-enhancement

lhsazevedo:
Just pushed some commits addressing both points.

@Girgias @simonrjones 
Both search and menu buttons are progressively enhanced on mobile as well, using the same approach as [gov.uk](https://gov.uk). I've included end-to-end tests for them as well.

@kamil-tekiela 
Contrast increased a bit for all menu items.

<details><summary>Fun fact:</summary>

These changes were finished and pushed during an [PHPRio](https://www.meetup.com/php-rio/) Pub Meetup event :beers: 

IMG_20241011_213642091

 
</details>

include/footer.inc
```diff
@@ -108,5 +108,71 @@ if (!empty($_SERVER['BASE_PAGE'])
...
+    </div>
+    <div
+      role="listbox"
+      aria-label="Search results"
+      id="search-modal__results"
+      class="search-modal__results"
+    ></div>
+    <div class="search-modal__helper-text">
+      <div>
+        <kbd>↑ ↓</kbd> to navigate •
```
Girgias, line 165: The kdb tag represents a *single* keyboard key.
```suggestion
        <kbd>↑</kbd> and <kbd>↓</kbd> to navigate •
```

Girgias (approved):
Some minor nits, but LGTM.

I will let other people more in charge with the Website take over.

js/search.js
```diff
@@ -1,378 +1,394 @@
...
-            });
+    /**
+     * Render the search results.
+     *
+     * @param {Array} results The search results.
+     */
+    const renderResults = (results) => {
+        const escape = (html) => {
+            var div = document.createElement("div");
+            var node = document.createTextNode(html);
```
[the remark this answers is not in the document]
lhsazevedo: Oops, nice catch. Forgot to refactor after copying from [here](https://stackoverflow.com/a/25396011).

js/common.js
```diff
@@ -253,28 +252,25 @@ function globalsearch(txt) {
...
-                        return;
-                    }
+        for (const node of cache.data) {
+            if (
+                node.description.toLowerCase().contains(term) ||
+                node.name.toLowerCase().contains(term)
+            ) {
+                $("#goto .results ul").append(`
+                    <li>
+                        <a href='/manual/en/${node.id}.php'>
```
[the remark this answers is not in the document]
lhsazevedo: Yup, it probably should. Actually, this code is part of the "global search" feature that triggers with <kbd>g</kbd> then <kbd>a</kbd>. I wasn’t aware of it until I had to adjust the code to prevent breaking due to changes in the search index structure.

Screen Shot 2024-10-21 at 20 29 51

TBH, I believe that we could just remove it and use <kbd>g</kbd> <kbd>a</kbd> as another shortcut to the new search dialog, but may be out of scope for this PR.

For now I've pushed a commit for the anchor href you pointed and the cache key a few lines before it.

ddevsr:
I think keep header position fixed, but after scrolling resize to small is good

Girgias:
@saundefined can you have another look at this?

lhsazevedo:
Sounds good! No rush, it would be great to have this ready for the 8.4 release though.

lhsazevedo:
Fixes #502

theodorejb:
Can the mobile nav sidebar be made to close when tapping outside of it? I instinctively tried to do this when testing on my phone, and it feels annoying that I instead have to tap a small X in the corner.

saundefined:
@lhsazevedo can you take a look?
https://github.com/lhsazevedo/web-php/pull/1

>Can the mobile nav sidebar be made to close when tapping outside of it?

Should solve it

lhsazevedo:
Thanks @saundefined, merged.

I noticed some common code between the offcanvas menu and the search modal. 

https://github.com/php/web-php/blob/eb7a39d86eae7b1d608bbb7def3d84e50255b40a/js/search.js#L315-L321

In the future, we could consider refactoring this into a helper to keep things DRY. But hey, it's always good to go [WET](https://dev.to/wuz/stop-trying-to-be-so-dry-instead-write-everything-twice-wet-5g33) first!

saundefined:
So, no more comments have come in,
I guess we can merge this PR 🎉 

Thanks a lot for the work done!

haszi:
@lhsazevedo Thanks for this! The UI and search in general look and feel much more natural and polished.

One question I have that I couldn't find an answer to here is whether the removal of the manual lookup page is deliberate or accidental? While in it's current form it is of limited use, I did like it taking me to a function's page when I typed the full function name into the search box and hit enter.

lhsazevedo:
@haszi Great to hear that :)

The Google Search was a conscious decision (now the lookup it is only used when JavaScript is not available), but I also thought that this was the previous behavior. I just checked the previous implementation, and indeed, hitting enter would trigger the lookup.

While I think the Google search would benefit most of the users, I can see the value of the lookup use as well. Considering only those two implementations, I see two options:

- Restore the old behavior, were <kbd>Enter</kbd> triggers the lookup and the Google search option is the last entry in the result list.
- Keep it as is, with the lookup alternative being to navigate to the desired result with <kbd>↓</kbd>, then <kbd>Enter</kbd>.

Both sound reasonable to me, so I don't have a strong preference. I can implement the first option if desired, plus some minor quality-of-life adjustments, by next week.

haszi:
While I liked the search bar immediately taking me to a function's documentation, I don't have any strong preferences on this. And since I'm not active in `web-php`, neither do I think my opinion should carry too much weight. :-) That I leave up to those who do take care of the repo, like @saundefined . :-)

PR #1095, opened 2024-10-16
Remove useless repo parameter

sy-records:
Missing `$repo`

https://github.com/php/web-php/blob/bdd2b4c6ecfc59467368329a03453b3ecc9ec975/include/shared-manual.inc#L49-L52


saundefined (approved):
Thanks!

PR #1097, opened 2024-10-16
Run examples with php wasm

soyuka:
Adds a PHP Wasm binary to run documentation examples in the browser. We discussed this with @pronskiy (
fixes https://github.com/php/doc-en/issues/3259). 

https://github.com/user-attachments/assets/527aaa1d-4ba1-4098-a7e0-3b407e386937

I've builtin `PHP wasm 8.4.0-dev` and I'm not sure what would be preferred to keep this up to date. If you want to reproduce the build see:

https://github.com/soyuka/php-wasm/pull/7

Clone that branch and execute: 

```
docker buildx bake
cp build/* ../phpdoc/web-php/js
```

Work sponsored by [Les-Tilleuls.coop](https://les-tilleuls.coop/en) inline with what we built at the [API Platform Playground](https://api-platform.com/playground/) more informations on a [blog post](https://soyuka.me/how-to-create-a-php-playground-for-your-documentation/)


soyuka:
https://github.com/user-attachments/assets/527aaa1d-4ba1-4098-a7e0-3b407e386937

js/interactive-examples.js
```diff
@@ -0,0 +1,61 @@
...
+  const pre = document.createElement('pre');
+  pre.classList.add('examplescode');
+  pre.innerText = output;
+  div.appendChild(pre)
+  return container;
+}
+
+async function main() {
+  const buffer = [];
+	const {ccall} = await phpBinary({
```
dunglas, line 21:
```suggestion
  const {ccall} = await phpBinary({
```

(mixed tabs and spaces)

js/interactive-examples.js
```diff
@@ -0,0 +1,61 @@
...
+      buffer.push(data);
+    }
+  })
+
+  console.log("PHP wasm %s loaded.", ccall("phpw_exec", "string", ["string"], ["phpversion();"]));
+  let lastOutput = null
+
+	document.querySelectorAll('.example').forEach((example) => {
+		const button = document.createElement('button');
+		const phpcode = example.querySelector('.phpcode');
```
dunglas, line 39:
```suggestion
  document.querySelectorAll('.example').forEach((example) => {
    const button = document.createElement('button');
    const phpcode = example.querySelector('.phpcode');
```

js/interactive-examples.js
```diff
@@ -0,0 +1,61 @@
...
+
+    const code = phpcode.querySelector('pre code')
+    code.setAttribute('contentEditable', true)
+
+		button.innerText = 'Run code';
+		button.onclick = function() {
```
dunglas, line 45:
```suggestion
    button.innerText = 'Run code';
    button.onclick = function() {
```

js/interactive-examples.js
```diff
@@ -0,0 +1,61 @@
...
+      }
+
+      ccall("phpw_run", null, ["string"], ['?>' + phpcode.innerText]);
+      lastOutput = createOutput(buffer.join(''))
+      phpcode.parentNode.appendChild(lastOutput);
+      buffer.length = 0;
+		};
+
+		phpcode.parentNode.insertBefore(button, phpcode);
+	});
```
dunglas, line 57:
```suggestion
    };

    phpcode.parentNode.insertBefore(button, phpcode);
  });
```

js/interactive-examples.js
```diff
@@ -0,0 +1,61 @@
...
+  title.innerText = 'Output: ';
+  container.appendChild(title)
+  const div = document.createElement('div');
+  div.classList.add('examplescode');
+  container.appendChild(div);
+  const pre = document.createElement('pre');
+  pre.classList.add('examplescode');
+  pre.innerText = output;
+  div.appendChild(pre)
+  return container;
```
dunglas, line 16:
```suggestion
  const container = document.createElement('div');
  container.classList.add('screen', 'example-contents');

  const title = document.createElement('p');
  title.innerText = 'Output: ';
  container.appendChild(title);

  const div = document.createElement('div');
  div.classList.add('examplescode');
  container.appendChild(div);

  const pre = document.createElement('pre');
  pre.classList.add('examplescode');
  pre.innerText = output;
  div.appendChild(pre);

  return container;
```

CS

js/interactive-examples.js
```diff
@@ -0,0 +1,61 @@
...
+		const button = document.createElement('button');
+		const phpcode = example.querySelector('.phpcode');
+
+    const code = phpcode.querySelector('pre code')
+    code.setAttribute('contentEditable', true)
+
+		button.innerText = 'Run code';
+		button.onclick = function() {
+      if (lastOutput && lastOutput.parentNode) {
+        lastOutput.parentNode.removeChild(lastOutput)
```
stof, line 47: must simpler: `lastOutput.remove()` (there is no reason to care about unmaintained browsers not supporting the `remove` method when you are adding a feature requiring WebAssembly to work)

js/interactive-examples.js
```diff
@@ -0,0 +1,61 @@
...
+      }
+      buffer.push(data);
+    }
+  })
+
+  console.log("PHP wasm %s loaded.", ccall("phpw_exec", "string", ["string"], ["phpversion();"]));
+  let lastOutput = null
+
+	document.querySelectorAll('.example').forEach((example) => {
+		const button = document.createElement('button');
```
stof, line 38: the created button should use the `button` type instead of the default `submit` type, to avoid issues in case this ends up being used in a place being inside a form.

js/interactive-examples.js
```diff
@@ -0,0 +1,61 @@
...
+    const code = phpcode.querySelector('pre code')
+    code.setAttribute('contentEditable', true)
+
+		button.innerText = 'Run code';
+		button.onclick = function() {
+      if (lastOutput && lastOutput.parentNode) {
+        lastOutput.parentNode.removeChild(lastOutput)
+      }
+
+      ccall("phpw_run", null, ["string"], ['?>' + phpcode.innerText]);
```
stof, line 50: should this use `.innerText` or `.textContent` here ? `innerText` is impacted by the styles of the page

js/interactive-examples.js
```diff
@@ -0,0 +1,61 @@
...
+      lastOutput = createOutput(buffer.join(''))
+      phpcode.parentNode.appendChild(lastOutput);
+      buffer.length = 0;
+		};
+
+		phpcode.parentNode.insertBefore(button, phpcode);
```
stof, line 56: use `phpcode.before()` instead

stof:
Disabling many extensions as you did in https://github.com/soyuka/php-wasm/pull/7 is likely to make many examples non working though (in your build, even JSON functions are not available)

js/interactive-examples.js
soyuka: no idea 🤷

soyuka:
> Disabling many extensions as you did in https://github.com/soyuka/php-wasm/pull/7 is likely to make many examples non working though (in your build, even JSON functions are not available)

Definitely, there are a few things to discuss, I can enable most extensions, some will require additional libraries (for example dom or xml require libxml) and it'll add weight to the build (now it's "only" 3.6 MB).

include/footer.inc
```diff
@@ -105,6 +105,13 @@ if (!empty($_SERVER['BASE_PAGE'])
    echo '<script src="/cached.php?t=' . @filemtime($path) . '&amp;f=/js/' . $filename . '"></script>' . "\n";
  }
 ?>
+<?php
+ $jsfiles = ["interactive-examples.js"];
+ foreach ($jsfiles as $filename) {
+   $path = dirname(__DIR__) . '/js/' . $filename;
+   echo '<script type="module" src="/cached.php?t=' . @filemtime($path) . '&amp;f=/js/' . $filename . '"></script>' . "\n";
+ }
+?>
```
soyuka, line 114: maybe that this deserves only one line?

js/interactive-examples.js
```diff
@@ -0,0 +1,62 @@
...
+
+  console.log("PHP wasm %s loaded.", ccall("phpw_exec", "string", ["string"], ["phpversion();"]));
+  let lastOutput = null
+
+  document.querySelectorAll('.example').forEach((example) => {
+    const button = document.createElement('button');
+    button.setAttribute('type', 'button')
+    const phpcode = example.querySelector('.phpcode');
+
+    const code = phpcode.querySelector('pre code')
```
pronskiy, line 42:
```suggestion
    const code = phpcode.querySelector('code')
```

I think `pre` is not needed here, is it?

On the preview site https://web-php-pr-1097.preview.thephp.foundation/manual/en/function.var-dump.php
I see the following error in the console:
image

pronskiy:
As I understand, php binary is always loaded now, right? If yes, does it make sense to load it only when the user clicks "Run" for the first time?

js/interactive-examples.js
soyuka: weird I have a different HTML on the local version but it works

soyuka:
done

soyuka:
Note that we should use brotli or gzip though, I'm unfamiliar with the server architecture behind the php docs website, should I add this directly to the cached.php file?

Girgias:
@derickr is likely the person that would know this.

jimwins:
It looks like php.net isn't particularly well set-up for encoding content. You can see how CSS/JS/etc is served up by the `cached.php` file, and testing with `curl --compressed -I` shows that it doesn't get encoded. HTML content does get gzip-encoded, at least.

I'm not sure how much of the encoding happens on our server vs. what Myra does on the edge, but there's probably room for improvement there even outside of encoding the WASM bundle appropriately. The current web configuration isn't in the systems repo yet, though.

lwlwilliam:
If the modified code can still be highlighted, that would be even better

stof:
@lwlwilliam this would require bringing an actual code editor (Codemirror or Monaco) instead of just using `contenteditable`

pronskiy:
@lwlwilliam, @stof, it makes sense for sure, but perhaps as step two in a separate PR? Just to keep this one smaller and delivered sooner. I'm pretty sure some unforeseen bugs will be discovered when this is deployed, and it'll be easier to fix them with smaller steps.

pronskiy:
@soyuka, what do you think about adding the PHP version in the output label? 
Screenshot 2024-10-21 at 11 42 10 AM

derickr:
Myra already caches `cached.php` with a 2 minute TTL (and 1 minute in case of a 404), but there are no settings for specific compressions:

```
prefix	/cached.php	2 minutes	1 minute
```

I can ask whether they can do on the fly-compression, but I am not sure whether they do.

To answer @jimwins, the config isn't in the systems repo, as nothing on php-web4 has been "php-ified" yet. I think Sascha set it up with nginx in his own way. There is no `/local/systems` either. The nginx configuration does not have any compression options set either.

I believe adding it to `cached.php` with the right headers (content-type **and** compression-type), with storing the compressed file should work. You should be able to test that yourself on your local development environment too.

js/interactive-examples.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+
+async function main() {
+  let lastOutput = null;
+
+  document.querySelectorAll(".example").forEach((example) => {
+    const button = document.createElement("button");
+    button.setAttribute("type", "button");
+    const phpcode = example.querySelector(".phpcode");
```
lwlwilliam, line 54:
```suggestion
  document.querySelectorAll(".example .example-contents").forEach((example) => {
    const button = document.createElement("button");
    button.setAttribute("type", "button");
    const phpcode = example.querySelector(".phpcode");
    if (phpcode === null) {
      return;
    }
```

There may be multiple `.phpcode` tags under `.example` tag.

image

derickr:
I don't quite understand whether your "try" was succesful or not. I can only see an image.

The javascript should surely be calling `/cached.php?....` and not the `.wasm` file directly...?

And we deploy on nginx, so you should test with that, and not with something else. I have just added the nginx configuration to the systems repository: https://github.com/php/systems/tree/master/php-web4/nginx

shaedrich:
> @pronskiy added version:

Should there be only one version or all of the supported in a dropdown/select?

pronskiy:
@shaedrich, for the sake of simplicity of this first step, it's better to stick to just one version. The dropdown sounds like a good idea for the next step.

cookieguru:
Probably want to add
```gitattributes
js/php-web.wasm filter=lfs diff=lfs merge=lfs binary
```
to `.gitattributes`

pronskiy:
@derickr, @saundefined, @sy-records, folks, do you think this is ready to merge?

saundefined (approved):

derickr:
I am getting failures when I try to check out this branch:

```
remote: Enumerating objects: 31, done.
remote: Counting objects: 100% (31/31), done.
remote: Compressing objects: 100% (15/15), done.
remote: Total 31 (delta 17), reused 27 (delta 16), pack-reused 0 (from 0)
Unpacking objects: 100% (31/31), 1.20 MiB | 1.79 MiB/s, done.
From https://github.com/php/web-php
 * [new ref]           refs/pull/1097/head -> pr/1097
M	distributions
Switched to branch 'pr/1097'
Encountered 1 file that should have been a pointer, but wasn't:
	js/php-web.wasm
error: cannot rebase: You have unstaged changes.
error: Please commit or stash them.
```

It is perhaps because something is on LFS, or not? The file just *looked* checked in here, so what is the deal with adding `lfs` to `.gitattributes`?

If it *is* LFS, then you need to provide instructions as to how to check it out properly.

derickr (changes requested):
I am getting failures when I try to check out this branch:

```
remote: Enumerating objects: 31, done.
remote: Counting objects: 100% (31/31), done.
remote: Compressing objects: 100% (15/15), done.
remote: Total 31 (delta 17), reused 27 (delta 16), pack-reused 0 (from 0)
Unpacking objects: 100% (31/31), 1.20 MiB | 1.79 MiB/s, done.
From https://github.com/php/web-php
 * [new ref]           refs/pull/1097/head -> pr/1097
M	distributions
Switched to branch 'pr/1097'
Encountered 1 file that should have been a pointer, but wasn't:
	js/php-web.wasm
error: cannot rebase: You have unstaged changes.
error: Please commit or stash them.
```

It is perhaps because something is on LFS, or not? The file just *looked* checked in here, so what is the deal with adding `lfs` to `.gitattributes`?

If it *is* LFS, then you need to provide instructions as to how to check it out properly.

derickr:
Merged, thanks so much!

It should be live in the next hour or so.

bohwaz:
hey guys, the WASM file doesn't appear to be compressed by the server?

soyuka:
@derickr to what script may I add the `php.wasm` build on https://github.com/php/systems/ ? 

> hey guys, the WASM file doesn't appear to be compressed by the server?

Would https://github.com/php/systems/pull/67 work? IMO we should get rid of `cached.php` and leave assets compression and cache management to nginx (or caddy).

cookieguru:
I meant to comment on this before it got merged, but removing it from LFS was not the way to go.  Running a git clone/pull now needs to download an additional 3.6 MB for this commit alone, and that will increase every time the wasm file is updated.  By placing it in LFS, only the version tied to the commit would have ever need to be fetched.  I think a simple `git lfs install` would have fixed the original issue

derickr:
@cookieguru We don't want to use LFS just yet - it's probably better to not have the WASM file in the repo at some point, but that causes problems too for deployment.

derickr:
Let's continue this discussion in https://github.com/php/systems/pull/67 (only).

saundefined:
@soyuka @pronskiy @Girgias 

What do you think to add selector with active PHP versions (e.g. 8.1-8.4)?

if the code behavior differs between versions, we display these differences in the documentation,
I think interactive examples would be more illustrative.

I suppose this way we can do away with `&example.outputs.*;` blocks altogether in the future.

pronskiy:
@saundefined, it totally makes sense to me.
How do you think it would work from a user’s perspective?

For example if the dropdown defaults to PHP 8.4, I hit Run, and it shows the 8.4 output. Then I switch the dropdown to 8.3 and hit Run again. Would it replace the output or show the diff somehow?

Girgias:
@derickr part of your concern seems to be fixed by: https://github.com/php/web-php/pull/1180

But not all examples produce output, which is a different issue.

shaedrich:
> @soyuka @pronskiy @Girgias
> 
> What do you think to add selector with active PHP versions (e.g. 8.1-8.4)?

I suggested the same thing:

> > @pronskiy added version:
> 
> Should there be only one version or all of the supported in a dropdown/select?

so, I'd love to have that as well. Maybe, now is the right time to implement it, now that the initial implementation is merged

pronskiy:
> Having multiple versions also means you need to have users download the wasm binary multiple times. I really don't think we should focus on that. 
@derickr, not necessarily. Any version of Wasm should be downloaded only when the user clicks “Run” for the first time.

For example:
* I visit php.net, and 8.4 is selected by default. No Wasm is loaded initially.
* I press “Run,” and the 8.4 Wasm binary is loaded and cached in the browser.
* I switch to version 8.3 using the dropdown and press “Run.” The 8.3 Wasm binary is then loaded and cached in the browser.
* Now, if I switch between 8.4 and 8.3, the examples run smoothly because the corresponding Wasm binaries are already cached.

PR #1103, opened 2024-10-24
Additional redirects for mongodb set-to-book conversion

jmikola:
The first redirect was added in php/web-php#1100, but additional sections that no longer have their own page were missed. The original conversion was done in php/doc-en#3627 and php/doc-base#138.

https://jira.mongodb.org/browse/PHPC-1247

### DIFF
--- a/error.php
+++ b/error.php
@@ -252,8 +252,14 @@ $manual_page_moves = [
     'regexp.reference' => 'regexp.introduction',
     "security" => "manual/security",
 
-    // Set to book
+    // MongoDB converted from set to book
     'set.mongodb' => 'book.mongodb',
+    'mongodb.installation.homebrew' => 'mongodb.installation',
+    'mongodb.installation.manual' => 'mongodb.installation',
+    'mongodb.installation.pecl' => 'mongodb.installation',
+    'mongodb.installation.windows' => 'mongodb.installation',
+    'mongodb.persistence.deserialization' => 'mongodb.persistence',
+    'mongodb.persistence.serialization' => 'mongodb.persistence',
 ];
 
 if (isset($manual_page_moves[$URI])) {
--- a/error.php
+++ b/error.php
@@ -254,12 +254,12 @@ $manual_page_moves = [
 
     // MongoDB converted from set to book
     'set.mongodb' => 'book.mongodb',
-    'mongodb.installation.homebrew' => 'mongodb.installation',
-    'mongodb.installation.manual' => 'mongodb.installation',
-    'mongodb.installation.pecl' => 'mongodb.installation',
-    'mongodb.installation.windows' => 'mongodb.installation',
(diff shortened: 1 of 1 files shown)


error.php
```diff
@@ -252,8 +252,14 @@
     'regexp.reference' => 'regexp.introduction',
     "security" => "manual/security",
 
-    // Set to book
+    // MongoDB converted from set to book
     'set.mongodb' => 'book.mongodb',
+    'mongodb.installation.homebrew' => 'mongodb.installation',
```
jmikola, line 257: I considered redirecting to the URL fragment ([#mongodb.installation.homebrew](https://www.php.net/manual/en/mongodb.installation.php#mongodb.installation.homebrew)), which appears to be [supported](https://stackoverflow.com/a/66101035), but the logic in `elseif` below always appends `.php`.

cmb69: Maybe (*untested*):
````diff
 error.php | 7 ++++++-
 1 file changed, 6 insertions(+), 1 deletion(-)

diff --git a/error.php b/error.php
index 56a375fe..70f8e460 100644
--- a/error.php
+++ b/error.php
@@ -259,7 +259,12 @@ if (isset($manual_page_moves[$URI])) {
 } elseif (preg_match("!^manual/([^/]+)/([^/]+).php$!", $URI, $match) &&
           isset($manual_page_moves[$match[2]])) {
     status_header(301);
-    mirror_redirect("/manual/$match[1]/" . $manual_page_moves[$match[2]] . ".php");
+    $parts = explode("#", $manual_page_moves[$match[2]], 2);
+    if (count($parts) === 1) {
+        mirror_redirect("/manual/$match[1]/" . $parts[0] . ".php");
+    } else {
+        mirror_redirect("/manual/$match[1]/" . $parts[0] . ".php" . "#" . $parts[1]);
+    }
 }
 
 $manual_redirections = [
````

File: error.php [hunk 2]
```diff
@@ -262,7 +268,13 @@
 } elseif (preg_match("!^manual/([^/]+)/([^/]+).php$!", $URI, $match) &&
           isset($manual_page_moves[$match[2]])) {
     status_header(301);
-    mirror_redirect("/manual/$match[1]/" . $manual_page_moves[$match[2]] . ".php");
+
+    $parts = explode('#', $manual_page_moves[$match[2]], 2);
+    if (count($parts) === 1) {
+        mirror_redirect("/manual/{$match[1]}/{$parts[0]}.php");
+    } else {
+        mirror_redirect("/manual/{$match[1]}/{$parts[0]}.php#{$parts[1]}");
```
jmikola, line 276: Technically not required, but I opted for [curly syntax](https://www.php.net/manual/en/language.types.string.php#language.types.string.parsing.advanced) here for readability. I'm not sure why concatenation was used in the original code (seemed inconsistent since interpolation was already used for `$match[1]`).

sy-records (approved):
LGTM.

PR #1116, opened 2024-11-08
Fix offcanvas menu open bug

### LINKED ISSUE #1112: Fix navbar offcanvas menu bug
Resolve an issue on mobile screens where the offcanvas menu disappears when opened after closing the search dialog. At the moment, users can work around it by tapping the screen to close the menu and reopening it.

I am currently investigating the cause to push a fix.

### LINKED ISSUE #1118: Offcanvas nav opens with focus on close button instead of nav links
When the offcanvas navigation is opened, the initial focus is set on the close button instead of the navigation menu. This behavior may disrupt the user experience, especially for keyboard or screen reader users who expect the focus to start on the navigation items.

Reported in https://github.com/php/web-php/pull/1116#issuecomment-2463957847

lhsazevedo:
Add the click handler to the backdrop element instead of the document to close the offcanvas menu on mobile devices.

Fixes #1112, fixes #1118

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -465,8 +465,8 @@ $(document).ready(function () {
         offcanvasElement.querySelectorAll("input, button, a");
     const backdropElement = document.getElementById("navbar__backdrop");
 
-    const documentWidth = document.documentElement.clientWidth
-    const scrollbarWidth = Math.abs(window.innerWidth - documentWidth)
+    const documentWidth = document.documentElement.clientWidth;
+    const scrollbarWidth = Math.abs(window.innerWidth - documentWidth);
 
     const offcanvasFocusTrapHandler = (event) => {
         if (event.key != "Tab") {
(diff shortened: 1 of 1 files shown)


sy-records:
It's displaying properly, but there's a small problem:



The close button is selected by default after displaying

sy-records (approved):
LGTM. Thank you!

js/common.js
```diff
@@ -523,13 +523,10 @@ $(document).ready(function () {
...
-        if (
-            !offcanvasElement.contains(event.target) &&
-            !menuButton.contains(event.target)
-        ) {
-            closeOffcanvasNav()
+        if (event.target === backdropElement) {
+            closeOffcanvasNav();
         }
-    };
+    }
```
sy-records, line 529: Should this semicolon be retained?

```suggestion
    };
```

lhsazevedo: Yes! Some other lines were also missing semicolons, so I just ran this navbar section through Prettier again.

cmb69: Maybe we should employ a linter?

saundefined (approved):
Thanks!
Linter is a great proposal, I guess we can add it in a separate PR

PR #1119, opened 2024-11-08
Hosting jQuery locally

### LINKED ISSUE #1117: JQuery should not be loaded from external site
Right now, the JQuery library is loaded from an external site (code.jquery.com). Instead, we should self-host this, and place it in the `/cached.php` system: https://github.com/php/web-php/blob/master/include/footer.inc#L99

Previously it was probably on the jquery.com site due to it having a CDN, but we have that ourselves now, there is no reason to not self-host any more (and leak our user's IP addresses).

sy-records:
Close #1117


saundefined (approved):
Thanks!

dragoonis:
Hey @saundefined and @sy-records 

What's the rationale behind this ticket ?

When we originally added this via a CDN was because the browser could open multiple sockets via different domains, whereas there is a limit of sockets per domain. 

What has changed? Have you benchmarked the before/after on this? If so what are the results. 

Thanks.

cmb69:
See #1117.

dragoonis:
Thanks @cmb69 I've read that now, and @derick's comments.

What is our CDN now? Cloudfront?

jimwins:
Most/all (working towards all) of the PHP.net web services are fronted by https://www.myrasecurity.com/, which was co-founded by a long-time PHP community member.

dragoonis:
Thanks for replying @jimwins, I read your comment when you originally wrote it, but didn't reply. Cheers 👍

PR #1120, opened 2024-11-09
Update to actions/cache@v4

cmb69:
v3 is meant for nodejs 16, but that is no longer supported by hosted GH runners[1].  To avoid potential instabilities, we update to v4.

[1] <https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/>


sy-records (approved):

saundefined (approved):
Thanks!

PR #1121, opened 2024-11-10
Replace languages.inc globals with I18n\Languages consts

MauricioFauth:
The include/languages.inc file was not removed as it's used in other repositories. It should be removed after migration.

Tests were added to ensure that the global variables and the constants are in sync with each other.

### DIFF
--- a/docs.php
+++ b/docs.php
@@ -1,4 +1,7 @@
 <?php
+
+use phpweb\I18n\Languages;
+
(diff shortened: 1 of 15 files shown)


saundefined (approved):
Looks good, thanks, again!

Let's also give some time to other reviewers, and if there are no objections, merge it

sy-records (approved):

PR #1135, opened 2024-11-20
Change asymmetric visibility example to show how it simplifies code

theodorejb:
The previous example wasn't very helpful, since it only showed invalid code that will error (and could just as easily be replaced with a `readonly` property).

### DIFF
--- a/releases/8.4/release.inc
+++ b/releases/8.4/release.inc
@@ -158,12 +158,20 @@ PHP
                             <<<'PHP'
 class PhpVersion
 {
-    public string $version = '8.3';
-}
+    private string $version = '8.3';
 
-$phpVersion = new PhpVersion();
-var_dump($phpVersion->version); // string(3) "8.3"
-$phpVersion->version = 'PHP 8.4'; // No error
+    public function getVersion(): string
+    {
+        return $this->version;
+    }
+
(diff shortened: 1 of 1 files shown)


saundefined:
LGTM, thanks!

@TimWolla can you also take a look?

releases/8.4/release.inc
```diff
@@ -178,11 +185,13 @@ PHP
...
     public private(set) string $version = '8.4';
-}
 
-$phpVersion = new PhpVersion();
-var_dump($phpVersion->version); // string(3) "8.4"
-$phpVersion->version = 'PHP 8.3'; // Visibility error
+    public function increment(): void
+    {
+        [$major, $minor] = explode('.', $this->version);
+        $this->version = "$major." . (intval($minor) + 1);
```
TimWolla, line 192: Just on my phone and can't make an actual suggestion, but would make $minor++ as a separate statement to keep the complexity of this line low and a simple interpolation.

TimWolla (commented):
Example is clearly an improvement.

releases/8.4/release.inc
theodorejb: Updated!

saundefined (approved):

PR #1145, opened 2024-11-21
Update security.txt

adoy:
Updated security.txt expiration date as suggested by the release process following the announcement of the initial stable version of 8.4.

Cf: https://github.com/php/php-src/blob/master/docs/release-process.md#preparing-for-the-initial-stable-version-php-xy0


saundefined (approved):
Looks good, thanks a lot!

sy-records (approved):

PR #1148, opened 2024-11-21
Improve description for new `BcMath\Number` object

theodorejb:
Simplified and corrected wording - BCMath isn't only for float numbers.


saundefined (approved):
Thanks!

PR #1154, opened 2024-11-21
php8.4: Fix the title and RFC link of PDO driver specific subclasses

takaram:
PHP 8.4 has accepted two RFCs with similar titles, and they seem to be mixed up.

- https://wiki.php.net/rfc/pdo_driver_specific_subclasses
- https://wiki.php.net/rfc/pdo_driver_specific_parsers

This PR also adds `Pdo\Pgsql` class in the list of new PDO subclasses.


saundefined (approved):
Thanks!

PR #1142, opened 2024-11-21
Add BCMath block to PHP 8.4 release page

pronskiy:
Fixes #1140

PR #1155, opened 2024-11-21
Update Dutch translation for BcMath

nielsdos:
Reference: https://github.com/php/web-php/pull/1142#issuecomment-2490812100
cc @saundefined


saundefined (approved):
Thank you!

PR #1159, opened 2024-11-22
Turkish translation of PHP 8.4 Announcement Page

emir:
👋👋


releases/8.4/languages/tr.php
```diff
@@ -0,0 +1,57 @@
...
+    'bc_standard' => '<code>stream_bucket_make_writeable()</code> ve <code>stream_bucket_new()</code> artık <code>stdClass</code> yerine bir <code>StreamBucket</code> objesi döndürüyor.',
+    'bc_core' => '<a href="/manual/en/migration84.incompatible.php#migration84.incompatible.core.exit"><code>exit()</code> davranış değişikliği</a>.',
+    'bc_warnings' => '<code>E_STRICT</code> sabiti artık kullanımdan kaldırılmıştır.',
+
+    'footer_title' => 'Daha iyi performans, daha temiz sözdizimi, gelişmiş tür güvenliği.',
+    'footer_description' => '<p>PHP 8.4 kaynak dosyalarını indirmek için <a href="/downloads">indirme sayfasını</a> ziyaret edin. Windows için çalıştırılabilir dosyaları <a href="https://windows.php.net/download">PHP for Windows</a> sitesinden indirebilirsiniz. Tüm değişikliklerin listesi <a href="/ChangeLog-8.php#PHP_8_4">ChangeLog</a> içinde kayıtlıdır.</p>
+        <p>Detaylı bilgi için <a href="/manual/en/migration84.php">göç rehberine</a> göz atabilirsiniz.</p>',
```
saundefined, line 56:
```suggestion
        <p>Detaylı bilgi için <a href="/manual/tr/migration84.php">göç rehberine</a> göz atabilirsiniz.</p>',
```

Hey!

Would you like to provide links `/manual/tr/*` instead of `/manual/en/*` here and elsewhere?

emir: @saundefined Since my whole life has been spent with English documents, I’ve gotten so used to /en links that they didn’t even seem odd to me. Great catch. Thank you!

emir:
@saundefined This isn’t the topic of this PR, but we could also sort the languages alphabetically in common.php. What do you think?

saundefined:
> @saundefined This isn’t the topic of this PR, but we could also sort the languages alphabetically in common.php. What do you think?

It could be done, but I don't think it would be especially helpful...

Many languages do not use Latin, and correct sorting is probably possible only by keys,
and also the page tries to guess the user's language and show him the right version at once

saundefined (approved):
Thanks!

PR #1161, opened 2024-11-23
8.4: Changed announced version to 8.4 from 8.4.0

jorgsowa:
The version officially announced was 8.4.1, so it's better to skip the patch version in the page header.


sy-records (approved):

saundefined (approved):
Thanks!

PR #1172, opened 2024-12-03
Why use PHP

saundefined:
Draft version of Why use PHP page.

/cc @pronskiy


lwlwilliam:
I feel this is very suitable as the homepage of the official website. At present, the release log on the homepage only needs to display the title of the latest article in highlighted font. In addition, there should be an archive entrance in the navigation bar, which will look much more comfortable.

pronskiy:
@lwlwilliam, @AndriusBartulis, thanks for your feedback!
The plan is to launch it as a separate page, polish, analyse analytics data (#1183) and then plan how to update the main page.

jacekkarczmarczyk:
I think it's worth mentioning somewhere at the top:
- standarized dependency management (composer)
- built-in strong typing and external type checking and linting tools (phpstan, psalm etc)
- mature testing libraries (phpunit, pest)
- standards (PHP-FIG packages like Container, Cache etc., coding standards, PER, PSR)

I think these things are important to show that PHP is not only for simple pages or WP blogs (as many people may see PHP), but also for enterprise applications

why-use-php.php
```diff
@@ -0,0 +1,402 @@
...
+        <?php highlight_php_trimmed(
+            <<<'PHP'
+class HomeController
+{
+    #[Route(path: '/', method: HttpMethod::GET)]
+    public function index(): Response
+    {
+        return new Response(
+            headers: ['Content-Type' => 'text/html'],
+            body: '<h1>Welcome to the Modern PHP Application!</h1>'
```
mnapoli, line 42:
```suggestion
            body: '<h1>Welcome to Modern PHP!</h1>'
```

Not sure what the best exact wording is, but I believe the current sentence is incorrect, right?

why-use-php.php
```diff
@@ -0,0 +1,402 @@
...
+                        <li>
+                            Content Management Systems:
+                            <a href="https://wordpress.com/">WordPress</a>, <a href="https://www.drupal.org/">Drupal</a>,
+                            <a href="https://www.joomla.org/">Joomla!</a>, <a href="https://typo3.org/">TYPO3</a>.
+                        </li>
+                        <li>Learning Management Systems: <a href="https://moodle.org/">Moodle</a>.</li>
+                        <li>
+                            E-commerce platforms: <a href="https://woocommerce.com/">WooCommerce</a>,
+                            <a href="https://www.shopware.com/">Shopware</a>,
+                            <a href="https://prestashop.com/">PrestaShop</a>.
```
Pierstoval, line 203: 😉
```suggestion
                            <a href="https://prestashop.com/">PrestaShop</a>,
                            <a href="https://sylius.com/">Sylius</a>
```

pawelkudzia:
Great job, I really like idea about website like that! I can suggest adding info about the Symfony framework and it's ecosystem. If you can add it, then it would be amazing :)

godismyjudge95:
Statamic is a pretty popular CMS built on Laravel.  It has almost as many Github stars as Joomla and is consistently growing in packagist downloads.  Might be worth including them in the CMS list?

andrerom:
+100

Been working last 4 years i C#/Asp.Net, besides some [opinionated/unbalanced advocacy arguments](https://www.ibexa.co/blog-archive/why-php-and-symfony-and-not-java) written 10 years ago, here are some strong PHP arguments that I can think of today:
- Instant development feedback cycle, no compile time
- Fully dynamic at its core, but with lots of options added over the years for stricter type safety
- Fastest dynamic language, great for web pages, api's, prototyping, and more
- Large ecosystem of frameworks for web/api/++, many that can rival Asp.Net and Java/Scala options in features and UX. Example: Symfony and others have much better debugging tools/toolbars that gives instant insight into your whole stack (cache, req/res/mvc, sql, ...), this is much more fragmented with for instance Asp.NET across IDE and external tools.
- Lower hosting & deployment cost

---
Very very side topic: Only thing really missing from Asp.Net is a native template language like razor, so you can instantly know if something is wrong type or using wrong args without having to test whole application manually or with 100% coverage functional tests to detect easy mistakes.

xperia-query:
When will it be added?

PR #1177, opened 2024-12-03
Allow "/" when editing code

soyuka:
It was reported that you can't hit "/" when editing code, this fixes it. I was suprised that this part of the code doesn't use `mousetrap` as in `js/common.js`.

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -295,7 +295,7 @@ const initSearchModal = () => {
     // Open when / is pressed
     document.addEventListener("keydown", (event) => {
         if (event.target.contentEditable === 'true') {
-          return;
+            return;
         }
 
         if (event.key === "/") {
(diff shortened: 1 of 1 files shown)


js/search.js
```diff
@@ -294,6 +294,10 @@ const initSearchModal = () => {
 
     // Open when / is pressed
     document.addEventListener("keydown", (event) => {
+        if (event.target.contentEditable) {
+          return;
+        }
+
```
sy-records, line 300: Changing it this way will make it impossible to use the `/` popup search input box

js/search.js
```diff
@@ -294,6 +294,10 @@ const initSearchModal = () => {
 
     // Open when / is pressed
     document.addEventListener("keydown", (event) => {
+        if (event.target.contentEditable) {
```
sy-records, line 297:
```suggestion
        if (event.target.contentEditable === 'true') {
```

sy-records (changes requested):

sy-records (approved):

soyuka:
Thanks @sy-records !

PR #1180, opened 2024-12-05
Reuse existing output

Girgias:
@soyuka my JS skills are very rusty, so there might be a better and less tag dependent way on doing this.

But this allows reusing the existing output for examples if it exists.

### DIFF
--- a/js/interactive-examples.js
+++ b/js/interactive-examples.js
@@ -4,7 +4,7 @@ function createOutput(output) {
   const container = document.createElement("div");
   container.classList.add("screen", "example-contents");
   const title = document.createElement("p");
-  title.innerText = "Output (PHP "+ PHP.version +"):";
+  title.innerText = "Output of the above example in PHP "+ PHP.version +":";
   container.appendChild(title);
   const div = document.createElement("div");
   div.classList.add("examplescode");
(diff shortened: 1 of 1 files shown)


manual/en/function.strpos.php
```diff
@@ -197,4 +197,4 @@ function.</p></div>
  </div>
 
 
-</div><?php manual_footer(); ?>
+</div><?php manual_footer([]); ?>
```
soyuka, line 200: did you meant to commit the manual files?

js/interactive-examples.js
```diff
@@ -68,8 +78,13 @@ async function main() {
 
       const runPhp = await PHP.loadPhp();
       runPhp(phpcode.innerText);
-      lastOutput = createOutput(PHP.buffer.join(""));
-      phpcode.parentNode.appendChild(lastOutput);
+      if (exampleScreenPreElement !== null) {
+        exampleTitleParagraphElement.innerText = "Output of the above example in PHP "+ PHP.version +":";
```
soyuka, line 82: could you maybe store this string in a variable as its used twice?

soyuka (commented):
definitely a good improvement imo, thanks @Girgias

manual/en/function.strpos.php
```diff
@@ -26,13 +26,13 @@
...
-    1 => 'strpbrk',
+      0 => 'ref.strings.php',
+      1 => 'String Functions',
   ),
   'next' =>
   array (
-    0 => 'function.strrchr.php',
-    1 => 'strrchr',
+    0 => 'function.rtrim.php',
+    1 => 'rtrim',
```
sy-records, line 35: There's a problem with indentation here.

sy-records:
@soyuka Whenever I click `Run Code` the console outputs:

image


image

Girgias:
@sy-records which page are you encountering this? As I don't have this issue from a quick glance.

sy-records:
https://www.php.net/manual/en/function.array-chunk.php

Girgias:
Still nothing on Firefox, even when using the ZH language on the live manual :/

sy-records:
I can reproduce it in both Chrome and Firefox.

Firefox

Girgias:
I don't really understand why there is JS error. @soyuka can you have a look?

sy-records (approved):

saundefined (approved):

PR #1183, opened 2024-12-07
Add analytics

saundefined:
RFC: https://wiki.php.net/rfc/phpnet-analytics

/cc @pronskiy @crell

### DIFF
--- a/include/header.inc
+++ b/include/header.inc
@@ -91,6 +91,23 @@ if (!isset($config["languages"])) {
 
 <?php if (isset($config['meta_tags'])) { echo $config['meta_tags']; } ?>
 
+<!-- Matomo -->
+<script>
+    var _paq = window._paq = window._paq || [];
+    /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+    _paq.push(["setDoNotTrack", true]);
+    _paq.push(["disableCookies"]);
+    _paq.push(['trackPageView']);
+    _paq.push(['enableLinkTracking']);
+    (function() {
+        var u="//analytics.php.net/";
+        _paq.push(['setTrackerUrl', u+'matomo.php']);
+        _paq.push(['setSiteId', '1']);
+        var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+        g.async=true; g.src=u+'matomo.js'; s.parentNode.insertBefore(g,s);
+    })();
+</script>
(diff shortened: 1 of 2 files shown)


include/header.inc
```diff
@@ -91,6 +91,23 @@ if (!isset($config["languages"])) {
...
+<!-- Matomo -->
+<script>
+    var _paq = window._paq = window._paq || [];
+    /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+    _paq.push(["setDoNotTrack", true]);
+    _paq.push(["disableCookies"]);
+    _paq.push(['trackPageView']);
+    _paq.push(['enableLinkTracking']);
+    (function() {
+        var u="//analytics.php.net/";
```
saundefined, line 103:
```suggestion
        var u="https://analytics.php.net/";
```

sy-records (approved):

include/header.inc
```diff
@@ -91,6 +91,23 @@ if (!isset($config["languages"])) {
 
 <?php if (isset($config['meta_tags'])) { echo $config['meta_tags']; } ?>
 
+<!-- Matomo -->
```
sy-records, line 94: Should only `php.net` be allowed? exclude `*.preview.thephp.foundation`

saundefined: We have enabled "Only track visits and actions when the action URL starts with https://www.php.net" in Matomo,

but I guess we can add this one for double-check 🤔

saundefined: @sy-records updated

sy-records (approved):

leonardolara:
That's just great! I was wondering if there were any page access statistics for php.net.
It will be a great source of information for translators.
Any idea on how the info will be made available?

leonardolara:
I would like to see, for example, the page access counts (in English and Portuguese translation) from a geographic location (specially those inside Portuguese-speaking countries - Brazil, Portugal, Angola, Mozambique, etc). These stats will help me to prioritize the pt_BR translation efforts.

leonardolara:
> I would like to see, for example, the page access counts (in English and Portuguese translation) from a geographic location (specially those inside Portuguese-speaking countries - Brazil, Portugal, Angola, Mozambique, etc). These stats will help me to prioritize the pt_BR translation efforts.

@pronskiy, have you gathered enough data for these statistics?

PR #1189, opened 2024-12-23
Create Devcontainer config

garrettw:
Picks up where #995 left off, since the idea didn't seem objectionable.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,7 +3,7 @@
 {
     "name": "PHP",
     // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-    "image": "mcr.microsoft.com/devcontainers/php:1-8.2-bullseye",
+    "image": "mcr.microsoft.com/devcontainers/php:1-8.4-bullseye",
 
     // Features to add to the dev container. More info: https://containers.dev/features.
     "features": {
@@ -16,7 +16,7 @@
         // Configure properties specific to VS Code.
         "vscode": {
             "extensions": [
-                "ikappas.composer",
+                "DEVSENSE.composer-php-vscode",
                 "editorconfig.editorconfig",
                 "junstyle.php-cs-fixer",
                 "ms-vscode.makefile-tools",
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,38 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/php
+{
+    "name": "PHP",
+    // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+    "image": "mcr.microsoft.com/devcontainers/php:1-8.2-bullseye",
+
+    // Features to add to the dev container. More info: https://containers.dev/features.
+    "features": {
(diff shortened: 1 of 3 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,38 @@
...
+    "customizations": {
+        // Configure properties specific to VS Code.
+        "vscode": {
+            "extensions": [
+                "ikappas.composer",
+                "editorconfig.editorconfig",
+                "junstyle.php-cs-fixer",
+                "ms-vscode.makefile-tools",
+                "recca0120.vscode-phpunit"
+            ]
```
TimWolla: Not sure how I feel about this list of extensions, because of https://wiki.php.net/rfc/third-party-code.

TimWolla (commented):
Overall this makes sense to me. Thank you!

.devcontainer/devcontainer.json
garrettw: - `editorconfig.editorconfig` is the definitive, first-party extension to support the config file we have.
- Similarly, I would consider`ms-vscode.makefile-tools` to be first-party in the sense that it is made by VS Code, for VS Code.
- The other three are arguably more third-party, but might also meet the standard for inclusion.
- ~~I do see that at least one of these bears revisiting.~~ _Updated the Composer extension to a different one which has since taken over the "de facto" position._

TimWolla: `ikappas.composer` looks pretty unmaintained, having the last commit over 3 years ago with the maintainer not having any GitHub activitity at all. `recca0120.vscode-phpunit` has a very low number of installs and for PHP CS Fixer there are multiple extensions, where it's not clear that this specific one is the one we should endorse.

garrettw: I've already replaced the Composer extension. I can take the other two out.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,40 @@
...
+    "forwardPorts": [
+        8080
+    ],
+
+    // Use 'postCreateCommand' to run commands after the container is created.
+    "postCreateCommand": "sudo .devcontainer/post-create.sh && sudo chmod a+x \"$(pwd)\" && sudo ln -s \"$(pwd)\" /var/www/html",
+    "waitFor": "postCreateCommand",
+	"postAttachCommand": {
+		"Server": "apache2ctl start"
+	}
```
TimWolla: Please fix the indentation.

TimWolla (approved):
Contents LGTM.

.devcontainer/post-create.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+test -f composer.json && composer install
+
+sed -i 's/Listen 80$//' /etc/apache2/ports.conf \
+    && sed -i 's/<VirtualHost \*:80>/ServerName 127.0.0.1\n<VirtualHost \*:8080>/' /etc/apache2/sites-enabled/000-default.conf \
+    && rm -rf /var/www/html
```
derickr, line 7: Why is this removing `/var/www/html` ?

derickr (commented):
LGTM, but I have a question.

.devcontainer/post-create.sh
garrettw: The folder already exists in the image but we need it to be a symlink to the repo root, so this is a prep step before the symlink is created.
The alternative would be another `sed` command to change the `DocumentRoot` to point at the repo root.

PR #1193, opened 2024-12-26
Fix loadIndexWithFallback failed

### LINKED ISSUE #1122: Search panel makes infinite loop requests in main page
On the main page request to "https://www.php.net/js/search-index.php?lang=ru" is too many repeating and looks like infinite loop.

telegram-cloud-photo-size-2-5300784139655572273-y

I have cleared browser cache and refresh the main page, it repeated.

UPD:  The same in incognito mode (even just not typed something into search form)

sy-records:
Fix #1122

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -137,11 +137,11 @@ const initPHPSearch = async (language) => {
      */
     const loadIndexWithFallback = async () => {
         try {
-            const searchItems = await loadIndex();
-            return searchItems;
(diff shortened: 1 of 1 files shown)


saundefined (approved):
Thanks!

lhsazevedo:
Oops, sorry about that. Initially I had the language as parameter, but removed it in one of the development iterations.

PR #1213, opened 2025-01-27
Exclude input and textarea from "/" key handler

### LINKED ISSUE #1211: Search modal popup interferes with submitting user notes.
The `/` key is used as a shortcut to bring up the documentation search modal popup. This is fine virtually everywhere, but the one place where it is not is the "add a note" feature: when typing in the form-note text area you'll trigger the popup if you try typing something like `$x = 1/5;`.

The workaround is to write the entire note elsewhere, and then copy-paste.

[Browser in question: FF134, OS Win10]

sy-records:
Fix #1211 

Updated the keydown event listener to exclude <input> and <textarea> elements in addition to contentEditable elements. This ensures that the "/" shortcut does not interfere with user input in form fields or editable areas.

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -294,7 +294,13 @@ const initSearchModal = () => {
 
     // Open when / is pressed
     document.addEventListener("keydown", (event) => {
-        if (event.target.contentEditable === "true") {
+        const target = event.target;
+
+        if (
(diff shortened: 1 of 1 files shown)


saundefined (approved):
Thanks!

PR #1214, opened 2025-01-27
Add scroll padding for anchor links

xPaw:
On pages with sticky breadcrumbs, this fixes the anchored link to scroll less to account for the stickied header.

This may not be enough if the sticky bar happens to occur more space than 64 pixels (4rem) I set it to.

Example link to try on: https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.blacklist-filename

Before:
image

After:
image


saundefined (approved):
Looks good, thanks!

PR #1217, opened 2025-01-31
feat: add conference PHP Velho Oeste 2025

marcosmarcolin:
Hello,

I would like to add to the website our PHP conference in Southern Brazil, called PHP Velho Oeste(Old West).


saundefined (approved):
Thanks!

PR #1238, opened 2025-03-05
Add "enum." to manual search sections

### LINKED ISSUE #1237: the shortcut for manual lookup does not work for enums
Accessing `https://php.net/PDOStatement` displays the documentation page of the `PDOStatement` class. `https://php.net/str_replace` also works to see the documentation of the `str_replace` function. However https://php.net/RoundingMode does not show the documentation of the `RoundingMode` enum, and the error page it displays only lists some functions (some of them being quite far from the `RoundingMode` text) but not the `RoundingMode` enum.

It would be nice if this lookup implementation could support enums as well.

sy-records:
Fix #1237

### DIFF
--- a/include/site.inc
+++ b/include/site.inc
@@ -189,7 +189,7 @@ function get_available_sqlites() {
 // Get all manual prefix search sections
 function get_manual_search_sections() {
     return [
-        "", "book.", "ref.", "function.", "class.",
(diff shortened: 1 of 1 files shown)


saundefined (approved):
Thanks!

PR #1243, opened 2025-03-13
FIX supported-versions MOBIL table

xerc:
Bildschirmfoto 2025-03-13 um 18 01 17

Bildschirmfoto 2025-03-13 um 18 10 51


xerc:
jQ does not like this syntax => working on it

js/common.js
```diff
@@ -826,16 +826,13 @@ function getLanguage() {
      * "data-label" with the text of the th.
      */
     $(document).ready(function () {
-        $('table').each(function () {
```
sy-records, line 829: It's broken: https://web-php-pr-1243.preview.thephp.foundation/eol.php

sy-records (changes requested):

js/common.js
xerc: was more broken before ;)

sy-records (approved):

styles/supported-versions.css
```diff
@@ -19,10 +19,6 @@ table.standard tr.stable td:first-child {
 	}
 
 	table.standard td.collapse-phone {
-		padding: 0;
-	}
-
-	table.standard td.collapse-phone * {
 		display: none;
 	}
```
sy-records, line 23:
```suggestion
	}

	table.standard:nth-of-type(2) td::before {
		content: unset;
	}
```

PR #1252, opened 2025-04-07
Fix search result descriptions are double-escaped

### LINKED ISSUE #1239: [Bug] Search result descriptions are double-escaped
Here is what happens when searching for an enum in the French version of the website (using French is relevant because the French description contains accents):

Image

sy-records:
Fix #1239

The data is already escaped.

image

### DIFF
--- a/js/search-index.php
+++ b/js/search-index.php
@@ -59,7 +59,7 @@ $js = json_decode($s, true);
 
 foreach ($js as $k => $item) {
     if ($item && isset($index[$k])) {
-        $index[$k][1] = $item;
+        $index[$k][1] = html_entity_decode($item);
     }
 }
 
--- a/js/search.js
+++ b/js/search.js
@@ -418,7 +418,7 @@ const initSearchUI = ({ searchCallback, language, limit = 30 }) => {
                             id="search-modal__result-description-${i}"
                             class="search-modal__result-description"
                         >
-                            ${description}
+                            ${escape(description)}
                         </div>
                     </div>
                 </a>
--- a/js/search.js
+++ b/js/search.js
@@ -418,7 +418,7 @@ const initSearchUI = ({ searchCallback, language, limit = 30 }) => {
                             id="search-modal__result-description-${i}"
                             class="search-modal__result-description"
                         >
-                            ${escape(description)}
+                            ${description}
                         </div>
                     </div>
                 </a>


stof:
Returned escaped HTML in the JSON is quite uncommon though. It is generally recommended to perform the escaping at the point of displaying. Maybe the JSON should be changed instead.

lhsazevedo:
>It is generally recommended to perform the escaping at the point of displaying

Agreed. A simple solution would be to decode the HTML entities before escaping. This prevents double escaping while still protecting against XSS:

```js
const decodeHtmlEntities = (str) => {
  const textarea = document.createElement('textarea');
  textarea.innerHTML = str;
  return textarea.value;
};
```

```diff
- ${escape(description)}
+ ${escape(decodeHtmlEntities(description))}
```

```
Authenticate as &quot;none&quot; <script>alert(123)</script> &#x3C;script&#x3E;alert(123)&#x3C;/script&#x3E;
```

image

stof:
> Agreed. A simple solution would be to decode the HTML entities before escaping.

To me, a better solution is to remove the escaping performed when generating the Json response.

sy-records:
> Returned escaped HTML in the JSON is quite uncommon though.

Since we are using the json format, if the string contains `"`, it will be escaped.

```xml
  <refname>ssh2_auth_none</refname>
  <refpurpose>Authenticate as "none"</refpurpose>
```

sy-records:
Yes, but no HTML will appear in the current scene.

lhsazevedo:
I think @stof meant that HTML entities are not needed for escaping quotes just for JSON transport. We can just use standard JSON escaping `\"`:

```json
{
  "example": "Authenticate as \"none\""
}
```

PHP `json_encode` and JavaScript  `JSON.parse` should already handle this for us, and it should be transparent for our front-end code.

js/search-index.php
```diff
@@ -59,7 +59,7 @@
 
 foreach ($js as $k => $item) {
     if ($item && isset($index[$k])) {
-        $index[$k][1] = $item;
+        $index[$k][1] = html_entity_decode($item);
```
stof, line 62: Why do we need to decode it here ? Shouldn't the code generating `search-description.json` be updated instead ?

sy-records: https://github.com/php/phd/pull/196

PR #1259, opened 2025-04-23
Make table header sticky within doctable class

sy-records:
https://www.php.net/eol.php

image


saundefined (approved):
Looks good, thanks!

PR #1262, opened 2025-04-24
Add API for QA

saundefined:
- [x] Add API endpoint `releases/release-candidates.php`
- [x] Add schema to `openapi.yml`


tianon:
Would this be an appropriate time/place to consider unifying this API with https://www.php.net/releases/index.php ? (which we use for GA releases)

To be clear, I don't have a strong opinion, and will happily consume the data from wherever you want to maintain it, but given there's already an API here I figured I'd raise it just in case it's a helpful reminder. :+1: :heart:

saundefined:
@tianon for stable branches, there is https://www.php.net/releases/branches.php endpoint.

If it's better to do everything in one place and add, maybe, a filter in the query parameters, let's do that.
If it's more convenient to have different endpoints - also ok :)

In any case, I will be glad to hear your opinion.

tianon:
Oh, that one's way cleaner than what we use now, which is https://www.php.net/releases/index.php?json&max=100&version=8.4 (replacing 8.4 as appropriate, of course).  IMO unifying pre-releases into those APIs makes sense, but I'm not the maintainer, just a humble user. :bow: :heart:

tianon:
Ah, `branches.php` doesn't include the artifact checksums, which we do make use of, but it does include other really useful data like whether a given branch is EOL. :thinking:

derickr (commented):
LGTM, but also perhaps add a link to these from `/release-candidates.php`'s sidebar?

PR #1266, opened 2025-05-05
Open the Foundation website in a new tab on the home page

marcosmarcolin:
Hello, to maintain the standard of the social media link and also for a better experience.

### DIFF
--- a/index.php
+++ b/index.php
@@ -181,7 +181,7 @@ $SIDEBAR = <<<SIDEBAR_DATA
       <a href='https://thephp.foundation/' class='headline'>The PHP Foundation</a>
       <div class='body'>
         <p>The PHP Foundation is a collective of people and organizations, united in the mission to ensure the long-term prosperity of the PHP language.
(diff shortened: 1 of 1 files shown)


index.php
```diff
@@ -181,7 +181,7 @@
       <a href='https://thephp.foundation/' class='headline'>The PHP Foundation</a>
       <div class='body'>
         <p>The PHP Foundation is a collective of people and organizations, united in the mission to ensure the long-term prosperity of the PHP language.
-        <p><a href='https://thephp.foundation/donate/' class='btn btn-primary'>Donate</a></p>
+        <p><a href='https://thephp.foundation/donate/' target="_blank" class='btn btn-primary'>Donate</a></p>
```
saundefined, line 184:
```suggestion
        <p><a href='https://thephp.foundation/donate/' target="_blank" rel="noopener noreferrer" class='btn btn-primary'>Donate</a></p>
```

Thanks! LGTM, but I'd like to add also `rel="noopener noreferrer"`

marcosmarcolin: Done.

I thought about it, but I didn't add it at first, because unlike social networks, Foundation is a PHP Group website.

PR #1270, opened 2025-05-28
fix: language pt_BR option.

marcosmarcolin:
Hi,

_Brazilian Portuguese_ are already Brazilian words.

In English, the correct way to use _Brazilian Portuguese_ is as follows: https://github.com/php/web-php/blob/master/src/I18n/Languages.php#L23

Releases page:
image

Docs page:
image

(I'm Brazilian).


sy-records (approved):
Thanks!

PR #1277, opened 2025-06-10
First stab at a prototype for revamping the download instructions page

derickr:


downloads-get-instructions.php
```diff
@@ -0,0 +1,97 @@
...
+        echo <<<ENDAPT
+<p>
+    On the command line shell, enter:
+</p>
+<div class="example"><div class="example-contents screen"><pre>
+sudo apt -y install software-properties-common
+sudo add-apt-repository ppa:ondrej/php
+sudo apt update
+sudo apt install php{$version}
+</pre></div></div>
```
TimWolla, line 87: Using PPAs is Ubuntu-specific. For Debian the repository should be set up as per: https://packages.sury.org/php/README.txt

derickr: Yeah, we need to work on all of these. And probably also stick them in files.

downloads-get-instructions.php
```diff
@@ -0,0 +1,97 @@
...
+        default => ''
+    };
+
+    $versionDir = match($_GET['version']) {
+        'php84' => '8.4',
+        'php83' => '8.3',
+        'php82' => '8.2',
+        'php81' => '8.1',
+        default => '8.4'
+    };
```
derickr, line 22: This "routing" needs to be more clever, likely.

downloads.php
```diff
@@ -37,15 +35,85 @@
...
+<select id="osvariant" name="osvariant">
+	<?= option('osvariant', 'linux-deb-buster', 'Debian Buster'); ?>
+	<?= option('osvariant', 'linux-deb-bullseye', 'Debian Bullseye'); ?>
+	<?= option('osvariant', 'linux-deb-bookworm', 'Debian Bookworm'); ?>
+	<?= option('osvariant', 'linux-rpm-fedora41', 'Fedora 41'); ?>
+	<?= option('osvariant', 'linux-rpm-fedora42', 'Fedora 42'); ?>
+	<?= option('osvariant', 'linux-rpm-redhat', 'RedHat'); ?>
+	<?= option('osvariant', 'osx-latest', 'Latest'); ?>
+	<?= option('osvariant', 'windows-wsl', 'with WSL'); ?>
+	<?= option('osvariant', 'windows-normal', 'without WSL'); ?>
```
derickr, line 76: The prefixes are important here for disabling/enabling based on OS.

derickr:
Superseded by #1287

Issue #1265, opened 2025-05-05
Add PHP History timeline


pronskiy:
Hey folks,

Back in 2020, we at JetBrains put together this timeline of PHP’s history: https://jb.gg/php25.
The event data is open source: https://github.com/JetBrains/php-timeline. But since the site isn’t dynamic, it hasn’t really been updated since.

I was thinking it might make sense to move it over to php.net, maybe under something like https://www.php.net/history.

@mkurzeja kindly offered to help with the implementation.

@saundefined @samdark @sebastianbergmann — what do you think?

saundefined:
Sounds great to me

mkurzeja:
Hi, it took a bit of time, but we tested two different approaches. 

The main problem is how to combine the existing design and content approach with something that will look attractive.

The initial test was based on DocBook, but the outcome was disappointing, and I doubt anyone would read the page. It was looking really boring. Another issue with DocBook was it would not support feeding the events from "database file", like JSON.

Due to that, we decided to just add a page to the main web repo. It is built based on a JSON file, so adding a new event is just adding it to the JSON. We adjusted the styles a bit to make it look better, but still in line with the overall page styles. The main downside of this is that it will be only in English, and we will need to maintain it in the web-php repo, I think both are OK.

It currently looks like this: https://drop.accesto.com/v/f0cATp

We will now work on adding the events from years 2020-2025.

pronskiy:
@mkurzeja, thanks for sharing, this looks awesome! 

Do you think it makes sense to implement the draft PR _before_ adding 2020-2025 data? Because otherwise I assume there will be more discussions on the events rather than the implementation/design part, and that might slow down the progress.

sgolemon:
While I like the idea of having a timeline, is this the best place to maintain it?  Any notable event becomes a commit in the web-php repo requiring review.  I feel like a wiki (either wiki.php.net or just wikipedia.com) makes more sense here as it provides an easier path for contributors and lower friction for the project.  Especially if we defer it to wikipedia, which I consider a trustworthy external repository and the easiest flow for new contributors.

mkurzeja:
@sgolemon Thanks for the thoughtful suggestion about using a wiki! We actually considered that approach along with other formats, but found they didn't offer enough flexibility for our specific formatting needs. We were aiming for a timeline view initially, but adapted to a visual list with priority levels to better match the php.net design.

[...]

PR #1284, opened 2025-06-23
Add PHP History Timeline

mkurzeja:
Initial implementation of #1265

This adds two main changes:

1. New view of PHP History Timeline as a separate page. Events are loaded from a .json file

image

https://drop.accesto.com/v/f0cATp

2. Adds link to the new page to the right side menu:
image

Events from years 2020-2025 are not included in this PR, I will create a separate one.


saundefined:
Looks great, thanks!

Please note [Sara's comment](https://github.com/php/web-php/issues/1265#issuecomment-3000639191) (I suppose we can wait for others to weigh in as well).

IMO: I don't see a problem with web-php review for changes like this, as it's no more work than user conference review, but I'd be happy to hear from others.

mkurzeja:
Thanks, I get the point with the wiki idea, I just struggle a bit with some flows around it. Will summarize everything and reply to the comment directly.

Something strange happened with the visual regression test. Screen was updated, yet if failed again. I will check this again.

mkurzeja:
@saundefined we are now working on running the visual tests locally, so we can check why it fails.

saundefined:
@mkurzeja the last visual test looks awesome! :tada:

PR #1287, opened 2025-07-03
First stab at a prototype for revamping the download instructions page

derickr:


include/download-instructions/windows-native.php
```diff
@@ -2,5 +2,5 @@
 On the command line, run the following commands:
 </p>
 <pre><code class="language-powershell line-numbers">
-powershell -c "irm https://php.net/downloads/install.ps1 <?= $version; ?> | iex"
+powershell -c "& ([ScriptBlock]::Create((irm 'https://php.net/install/windows.ps1'))) -Version <?= $version; ?>"
```
derickr: This should be `https://www.php.net/install/windows.ps1` to avoid redirects. I'm also thinking that instead of `install`, the directory should be `include/download-instructions`, like the other files.

include/download-instructions/linux-debian-cli-community.php
```diff
@@ -2,11 +2,11 @@
...
-sudo apt-get update
-sudo apt-get install -y lsb-release ca-certificates apt-transport-https curl
-sudo curl -sSLo /tmp/debsuryorg-archive-keyring.deb https://packages.sury.org/debsuryorg-archive-keyring.deb
-sudo dpkg -i /tmp/debsuryorg-archive-keyring.deb
-sudo sh -c 'echo "deb [signed-by=/usr/share/keyrings/debsuryorg-archive-keyring.gpg] https://packages.sury.org/php/ $(lsb_release -sc) main" &gt; /etc/apt/sources.list.d/php.list'
-sudo apt-get update
+# Add the packages.sury.org/php repository.
+sudo apt update
+sudo apt install -y curl
+curl -sSL https://packages.sury.org/php/README.txt | sudo bash -x
```
derickr: I don't think we should tell users to `curl | bash` external sources. It's not a great way to teach devs security practises.

include/download-instructions/windows-native.php
derickr: I've fixed the domain part of this now.

include/download-instructions/linux-debian-cli-community.php
shivammathur: Reverted back to the full instructions.

include/download-instructions/windows-native.php
shivammathur: Yes, we can change the directory to include/download-instructions, my thought on adding the install directory was it will be short and would have recall value.

derickr:
This was merged after a rebase.

PR #1292, opened 2025-07-10
Add alt text to Elephants images

adamcable:
Fixing elements which do not have alt attributes

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -426,6 +426,7 @@ $(document).ready(function () {
                             link.attr('href', photo.url);
                             link.attr('title', photo.title);
                             image = $('<img>');
+                            image.attr('alt', photo.title);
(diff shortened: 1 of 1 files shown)


cmb69:
I'm not sure that just reusing the titles (which are apparently the file names) as alt text makes sense; that might actually be worse than having no alt texts at all. As is, a screen reader likely ignores the images, but now it may read: "DSC_0031" …

adamcable:
Yes - setting alt text to empty sounds better. And just makes sure it keeps to image alt text best practices

sy-records:
@cmb69 Hi, This doesn't seem to be a file name?

image

adamcable:
On live some are...

image

sy-records (approved):

saundefined (approved):
Looks good, thanks!

PR #1294, opened 2025-07-14
Optimize CSS for Google CSE search

### LINKED ISSUE #1113: Search results page does not look good
The results page that is displayed if you enter something in the search input and press enter.

https://www.php.net/search.php?lang=en&q=sdf

Tried both Firefox and Chromium.

sy-records:
Close #1113


saundefined (approved):
Awesome!
Thanks a lot

PR #1303, opened 2025-08-01
Normalize language code case in LangChooser

### LINKED ISSUE #1302: Language variant redirect does not respect priorities
When I open links like https://www.php.net/stat, it opens chinese version of documentation. 

It is caused by header `Accept-Language: cs,en;q=0.9,fr;q=0.8,zh-CN;q=0.7,zh;q=0.6`
When I change it to `Accept-Language: cs,en;q=0.9,fr;q=0.8` (remove chinese), it sends me to english version
That is not expected, because language should be chosen by priority - czech (no q parameter is implicitly 1, highest), english, french and then chinese. Czech is not supported by php.net, therefore english version should open.

This is proper implementation that returns supported language that has the highest priority: https://github.com/nette/http/blob/a02f3d131f8a5dbd573244a6813ed316cde8d6ad/src/Http/Request.php#L308-L335
Here is proof: https://3v4l.org/UJnKb#vnull

Possibly related to https://github.com/php/web-php/issues/435

I have fresh Windows 11 and Opera installation and for some reason french and chinese were added without me doing that and seeing chinese docs when my system is set to czech and english was really confusing.

I don't feel quite confident with the website codebase, so this will hopefully allow you to fix the current behavior. Thank you

sy-records:
Fix #1302

In `cs,en;q=0.9,fr;q=0.8,zh-CN;q=0.7,zh;q=0.6`, `zh-CN` is not converted to lowercase, so the match is `zh-` with a quality of 1. After converting `zh-` to zh, quality is greater than 0.9 for en.

### DIFF
--- a/src/LangChooser.php
+++ b/src/LangChooser.php
@@ -97,7 +97,7 @@ class LangChooser
                 // The language part is either a code or a code with a quality
                 // We cannot do anything with a * code, so it is skipped
                 // If the quality is missing, it is assumed to be 1 according to the RFC
-                if (preg_match("!([a-z-]+)(;q=([0-9\\.]+))?!", trim($value), $found)) {
+                if (preg_match("!([a-z-]+)(;q=([0-9\\.]+))?!", strtolower(trim($value)), $found)) {
                     $quality = (isset($found[3]) ? (float) $found[3] : 1.0);
                     $browser_langs[] = [$found[1], $quality];
                 }


sy-records:
@mabar Can you try it? https://web-php-pr-1303.preview.thephp.foundation/releases/8.4/index.php

mabar:
It works, thank you

PR #1308, opened 2025-08-09
Update prepend.inc add human logic function

ahilespelid:
it is known that 'is_scalar' does not skip 'null', but the documentation indicates that some resources may be skipped in the future, and since hypothetically a resource can return 'null', the code uses an excessive check for null '$logic_parametr ?? 0'. If you look under the hood of filter_var FILTER_VALIDATE_BOOLEAN, it is known that null returns false, so this construction is '$logic_parametr ?? 0' is redundant and must be replaced by '$logic_parametr'

### DIFF
--- a/include/prepend.inc
+++ b/include/prepend.inc
@@ -121,3 +121,11 @@ function google_cse(): void {
 
     echo $cse_snippet;
 }
+// Human bool function
+if(!function_exists('bool'))
+{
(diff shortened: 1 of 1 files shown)


sy-records (changes requested):
Thank you for your contribution, but where should the method added here be used?

derickr:
I think we can just close this now.

PR #1312, opened 2025-08-13
Update downloads content without full page reload

saundefined:


cmb69:
If there are two changes, there are two asynchronous requests. If the first response arrives after the second, the instructions would not match the latest user selection. I'm not sure how to solve that. A *mitigation* could be to delay the request for some fraction of a second (what might be a good idea anyway).

Also, error handling should be improved; logging to the console won't usually be seen by the user, but worse, if the server reponds with an error status, the script would try to go on. There should be a check for `response.ok` in addition to the `.catch()`.

saundefined:
@cmb69 thanks!

> If there are two changes, there are two asynchronous requests. If the first response arrives after the second, the instructions would not match the latest user selection. I'm not sure how to solve that. A _mitigation_ could be to delay the request for some fraction of a second (what might be a good idea anyway).

What do you think about [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)? Just tested locally, it solves this problem.
I'll send a commit with it now, if that option doesn't work, I'll roll it back.

cmb69:
> What do you think about [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)?

Interesting! I wasn't aware of that class, but it looks like a sensible way to solve the issue.

downloads.php
```diff
@@ -217,11 +217,47 @@ function option(string $value, string $desc, $attributes = []): string
...
+
+                        if (newForm) form.innerHTML = newForm.innerHTML;
+                        if (newInstructions) instructions.innerHTML = newInstructions.innerHTML;
+
+                        if (window.Prism && typeof Prism.highlightAll === 'function') {
+                            Prism.highlightAll();
+                        }
+
+                        const newUrl = form.action.split('?')[0] + '?' + params;
+                        history.pushState(null, '', newUrl);
```
shivammathur: Since we push history, might be a good idea to handle back and forward navigation using `popstate`. Currently if you make a change and press the back button afterwards, the URL changes back, but the instructions don't.

shivammathur (changes requested):

derickr:
Will this all still work if javascript is disabled too?

cmb69:
If JS is not supported, it's just a normal form with an "Update Instructions" button. It might be better, though, to actually remove the button when JS is available (i.e. the concrete JS is supported by the browser), instead of relying on `<noscript>`, since JS might not be available even when supported by the browser (e.g. some old browsers may not support `fetch()`, and in that case the user can't use the form).

downloads.php
```diff
@@ -217,13 +215,77 @@ function option(string $value, string $desc, $attributes = []): string
...
+                    console.error('Request failed:', err)
+                    if (!instructions.querySelector('.error')) {
+                        instructions.innerHTML = `<p class="error">An unexpected error occurred.</p>`
+                    }
+                });
         }
+
+        function removeUpdateInstructionsButton() {
+            const btn = document.querySelector('#instructions-form button[type="submit"]')
+            if (btn) {
```
shivammathur: I think we should check if the browser supports `fetch` and  `AbortController` as @cmb69 mentioned.

```suggestion
            if (btn && window.fetch && window.AbortController) {
```

shivammathur (approved):

PR #1313, opened 2025-08-15
Added the basic bug template

derickr:
The idea is to have one for downloads.php, so that we can link to it directly.


sy-records:
Can we use the yml format? https://github.com/php/doc-en/pull/4811

The md format lacks mandatory field validation, which may result in some invalid submissions.

derickr:
@sy-records We can, this is just what the GH UI created. Would you mind making a PR for two templates then: normal bugs, and one for the downloads page (as a feature request), which sets the "Page: Downloads" label?

derickr:
I'll close this then.

sy-records:
I made some changes to your branch.  I'm not sure about the content of the normal bugs template. Most of it may be related to the document content?

derickr:
I think we can just use their standard template, but add a field for the URL?

Issue #1314, opened 2025-08-15
Download Instructions Improvement Request re: multiple versions on MacOS with MacPorts


beryllium:
As an avid MacPorts user, I have a couple of things that I would like to suggest for this sections of code that decides which file to include for which OS / variant / etc.

https://github.com/php/web-php/blob/10ff79838802fa9c6e8163cfc7a1a0d9e77efa33/downloads-get-instructions.php#L48-L60

Once a filename is decided on for what multiversion should look like, it could include these MacPorts instructions:

MacPorts versioning usually looks like this:

```bash
sudo port install php84
sudo port install php83 # Not sure how these templates would best represent "previous version"

# Changing between versions
sudo port select php php84
sudo port select php php83
```

Now, that's when PHP is installed & working normally. However, the MacPorts project is having a volunteer shortage and as a result there are some issues with installing PHP 8.4 and higher using it. Certain important extensions are not available.

There is a manual workaround that can be applied by modifying the related Portfiles in the ports tree. A user is meticulously maintaining [a custom commit/PR that contains the patch data](https://github.com/macports/macports-ports/pull/28140). However, the project is [not willing to accept the PR as a monolith, along with some additional concerns](https://github.com/macports/macports-ports/pull/28140#issuecomment-2804138782).

Since I have personal objections to using Homebrew, I'm kind of in a limbo state at the moment. But it would be nice to see the download instructions updated to provide more context to the user. Let me know if I can help! (I would just need a decision on how to specify the filename to react to the checkbox, since multiversion is not exactly the same as 'community' ...)

shivammathur:
@beryllium 

I think the instructions to switch to a version can be added as the last command, like we have the `brew link` step for `brew`, we can have the `port select` command for macports. Feel free to create a PR to add that to the instructions file.

I do not think the php downloads page would be the right place to mention the maintenance issues for macports.

beryllium:
I think my PR adequately addresses my concern here, so I'm closing this issue. Thanks for the guidance!

PR #1316, opened 2025-08-16
Add MacPorts instructions for switching PHP versions

beryllium:
Per issue #1314 , this PR adds some instructions to the MacOS MacPorts installation documentation for changing between PHP versions.

### DIFF
--- a/include/download-instructions/osx-macports.php
+++ b/include/download-instructions/osx-macports.php
@@ -5,7 +5,9 @@ On the command line, run the following commands:
 # Please refer to https://guide.macports.org/chunked/installing.macports.html for installing MacPorts.
 
 # Install PHP <?= $version; ?>
+
 sudo port install php<?= $versionNoDot; ?>
+
 
 # Switch Active PHP <?= $version; ?> version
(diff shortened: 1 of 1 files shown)


include/download-instructions/osx-macports.php
```diff
@@ -4,4 +4,10 @@
 <pre><code class="language-bash line-numbers">
 # Please refer to https://guide.macports.org/chunked/installing.macports.html for installing MacPorts.
 sudo port install php<?= $versionNoDot; ?>
+
+# To install a previous version:
+sudo port install php<?= $priorVersionNoDot; ?>
+
+# To switch to a different version:
+sudo port select php php<?= $priorVersionNoDot; ?>
```
shivammathur: @beryllium 

Currently for PHP 8.3, it is weird.

Adding, the `port select` command for the chosen version after the `port install` command should be sufficient to highlight how to switch.

```zsh
# Install PHP 8.4
sudo port install php84

# To switch to PHP 8.4
sudo port select php php84
```

shivammathur (changes requested):

PR #1329, opened 2025-08-18
Format arrow functions on 8.4 release page consistently with other docs pages

theodorejb:
Compare the examples on https://www.php.net/manual/en/functions.arrow.php.

[According to the author of the Arrow Functions RFC](https://www.reddit.com/r/PHP/comments/fctb18/comment/fjd9u6a/), they are intended to be written in the format `fn($x) => $x` (without a space after the `fn` keyword). This is also the formatting required by the popular [PER Coding Style](https://www.php-fig.org/per/coding-style/#71-short-closures).


sy-records (approved):
Thanks!

PR #1347, opened 2025-08-23
Tweak Longhorn PHP conference announcement title, remove Blind Bird pricing mention

iansltx:
The Blind Bird pricing call to action is about to be outdated (as of committing this), so dropping it to avoid being misleading.


iansltx:
This can wait 'til Sunday afternoon US time to be merged. Just a matter of "blind bird pricing will no longer be valid as of Monday"

iansltx:
Sorry, one more review here, and revising my statement on timing because we got some of the metadata swrong (which explains why it wasn't showing up on the php.net homepage).

sy-records (approved):

Issue #1160, opened 2024-11-22
Many calendar events point to defunct user groups


inxilpro:
I was just looking at the official [PHP events calendar](https://www.php.net/cal.php) and it seems like the vast majority of events listed are for groups that no longer exist. I reviewed the first 7 days of the November 2024 calendar, and only 5 of the 27 events listed are actually active:

- Page/domain not found: 14
- Link works + event still exists: 5
- Meetup exists but PHP calendar is broken: 1
- Last meetup > 5 years ago: 3
- Last meetup > 2 years ago: 1
- Ads for training: 3

I'd be happy to help. I can potentially PR a script that could at least check that event links resolve. I'd also be happy to manually review existing recurring events and remove the ones that no longer exist.

Let me know if you'd be open to accepting a PR for some automated clean up, and maybe point me in the direction of how I might apply for some sort of moderation access to the calendar?

Thanks for all the hard work the existing php.net volunteers put into to keeping the site updated and working!

heiglandreas:
There us a list of PHP usergroups at https://php.ug.

That List once was also renxered on php.net.

That list does include the events that the listed usergroups publish via an ICS file (meetup dies that out of the box).

The main issue is as always keeping the list of usergrouls up to date. I am doing my best but help from others is much appreciated. I wrote a script that checks regularily for inactive groups but it might miss some.

The events from the usergrouls are also exported as ICS file from php.ug

Would that be helpful?

inxilpro:
@heiglandreas I've been meaning to reach out to you, because I want to figure out how [phpx.world](https://phpx.world) and [php.ug](https://php.ug) can collaborate! I'm building PHP× as a way for groups to host a simple homepage and handle announment lists/event rsvps/etc easily. In an ideal world, maybe we can set it up so that PHP× events are automatically added to php.ug?

Either way, I would love to see the events calendar on php.net get cleaned up. I did a little digging today, and it seems like new events get inserted into a db and then emailed to the php-webmaster@ email address for approval. Then, approved events are written periodically to a CSV file that's read by `cal.php`.

We might be able to modernize that system a bit and make it a little easier for folks to manage/keep up to date. I'm happy to work on a PR, but I'd love to get a thumbs-up from someone who has the ability to merge PR's so that I don't spend a couple weeks on something that isn't the direction y'all want to go…

heiglandreas:
[...]

PR #1356, opened 2025-08-26
Revamp support page structure

pronskiy:
- Add relevant resources
- Move outdated to archive

### DIFF
--- a/support.php
+++ b/support.php
@@ -40,22 +40,23 @@ site_header(
  For day-to-day help and troubleshooting, the broader PHP community is very active in a few places:
 </p>
 <ul>
-  <li>
-      <p><strong>Reddit:</strong> <a href="https://www.reddit.com/r/PHP/">r/PHP</a> — look for the Weekly help threads that are pinned to the top for beginner and general questions.</p>
-  </li>
-  <li>
-    <p><strong>Stack Overflow:</strong> browse or ask questions using the <a href="https://stackoverflow.com/questions/tagged/php">php tag</a>.</p>
-  </li>
-  <li>
-    <strong>Discord:</strong> there are many community-ran servers for chat-based help and discussion. Here are a couple of popular ones:
-      <ul>
-          <li><a href="https://phpc.chat/">phpc.chat</a></li>
-          <li><a href="https://discord.gg/php-developers-484326318851358720">PHP Developers</a></li>
-      </ul>
-  </li>
-  <li>
-    <strong>IRC:</strong> real-time chat at <a href="https://libera.chat/">Libera.Chat</a> in <strong>#phpc</strong>. This channel is also bridged with Discord via <a href="https://phpc.chat/">phpc.chat</a>.
-  </li>
+ <li>
+  <p><strong>Reddit:</strong> <a href="https://www.reddit.com/r/PHP/">r/PHP</a> — look for the Weekly help threads that are pinned to the top for beginner and general questions.</p>
+ </li>
+ <li>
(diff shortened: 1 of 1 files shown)


saundefined (approved):

support.php
```diff
@@ -8,11 +8,10 @@
   <div class="body">
     <ul>
       <li><a href="#documentation">Documentation</a></li>
-      <li><a href="#mailing-lists">Mailing Lists</a></li>
```
derickr, line 11: You should not move a link to the mailing lists to "legacy", as it is still the primary way of discussing new features and technical details for PHP.

The support mailing lists are indeed less well used (if at all), but calling all lists "legacy" is also not accurate.

support.php
```diff
@@ -95,12 +81,29 @@
...
+ Please note that these are legacy resources and primarily useful for historical reference.
+</p>
+
 <h2 id="webmasters">PHP.net webmasters</h2>
 
 <p>
  If you have a problem or suggestion <em>in connection with the PHP.net
  website or mirror sites</em>, <a href="/contact.php">please
- contact the webmasters</a>. If you have problems setting up PHP
+ contact the webmasters</a>. <br> If you have problems setting up PHP
```
derickr: Instead, I would suggest to use:
```
</p>
<p>
```

support.php
```diff
@@ -36,50 +35,37 @@
...
- sessions) is included in the right-hand column of the front page, and
- on the <a href="/cal.php">event calendar page</a>. If you want to list
- an upcoming event, just fill out the form <a
- href="/submit-event.php">on this page</a>.
+ Connect with local and regional PHP User Groups and find upcoming meetups, conferences, and training sessions:
 </p>
+<ul>
+  <li>Find a user group near you on <a href="http://php.ug/">PHP.ug</a>.</li>
+  <li>Browse upcoming events on the <a href="/cal.php">event calendar</a>.</li>
+  <li>Want to list an event? Submit it <a href="/submit-event.php">here</a>.</li>
```
derickr: The calendar events are totally out of date, and serve almost no purpose. Instead, it would be better to integrate the resources from php.ug and also not allow submissions directly. /cc @heiglandreas

support.php
```diff
@@ -95,12 +81,29 @@
  <a href="http://oftc.net">OFTC</a>).
```
derickr: I think the only active group is `#phpc` on the libera network. So it would be best to mention this one specifically. It is also bridged with the Discord (linked from phpc.chat, and invites through https://discord.gg/php-developers-484326318851358720).

support.php
```diff
@@ -95,12 +81,29 @@
...
+<h3 id="mailing-lists">Mailing Lists</h3>
+<p>
+ Historically, a number of mailing lists were devoted to talking about PHP and related projects. While many are no longer active, you can still find their descriptions and archives on <a href="mailing-lists.php">this page</a>.
+</p>
+
+<h3 id="newsgroups">Newsgroups</h3>
+<p>
+ The PHP language newsgroup was comp.lang.php. Many of our mailing lists were also reflected onto the news server at
+ <a href="news://news.php.net">news://news.php.net/</a>. The server also has a read only web interface at
+ <a href="https://news-web.php.net/">https://news-web.php.net/</a>.
```
derickr, line 86: You don't need to use past tense here. They are still available on news-web and NNTP, and will continue to do so.

support.php
pronskiy: This is out of scope if this PR. Should be handled within https://github.com/php/web-php/issues/1160.

support.php
pronskiy: Support mailing lists are legacy, so they should not be mentioned here as active.

PR #1357, opened 2025-08-26
fix: adds trailing space

ismaelmartinezc:
I added a trailing space to fix this:
image

Please take a look at: https://www.php.net/releases/index.php


DanielEScherzer (approved):
Good catch, thanks!

PR #1372, opened 2025-08-28
Update title and descriptions for manual pages

derickr:
This makes use of the new information that Phd will display once https://github.com/php/phd/pull/201 is merged.

It changes the title on the home page also to just "PHP" (removing the `: Hypertext Preprocessor` part).

It also changes the meta information for manual pages, which is used for social media cards.

### DIFF
--- a/include/layout.inc
+++ b/include/layout.inc
@@ -427,7 +427,7 @@ function site_header(?string $title = NULL, array $config = []): void
 
     $title = $title ? "PHP: {$title}" : 'PHP';
     $meta_image_path = $MYSITE . 'images/meta-image.png';
-    $meta_description = $config['description'] ?? $title;
+    $meta_description = $config['description'] ?? "PHP is a popular general-purpose scripting language that powers everything from your blog to the most popular websites in the world.";
 
     $defaults = [
         "lang" => $LANG,


include/layout.inc
```diff
@@ -421,12 +421,13 @@ EOT;
...
 
-function site_header(string $title = 'Hypertext Preprocessor', array $config = []): void
+function site_header(?string $title = NULL, array $config = []): void
 {
     global $MYSITE, $LANG;
 
+    $title = $title ? "PHP: {$title}" : 'PHP';
     $meta_image_path = $MYSITE . 'images/meta-image.png';
-    $meta_description = "PHP is a popular general-purpose scripting language that powers everything from your blog to the most popular websites in the world.";
+    $meta_description = $config['description'] ?? $title;
```
TimWolla: That doesn't seem right. We should not use the title if there is no description given.

include/shared-manual.inc
```diff
@@ -325,6 +325,7 @@ function manual_setup($setup): void {
         "prev" => $setup["prev"],
         "next" => $setup["next"],
         "cache" => $lastmod,
+        "description" => $setup["this"][2] ?? null,
     ];
     site_header($setup["this"][1] . " - Manual ", $config);
```
sy-records:
```suggestion
    site_header($setup["this"][1] . " - Manual", $config);
```

index.php
```diff
@@ -108,7 +108,7 @@
   </div>
 EOF;
 
-site_header("Hypertext Preprocessor",
```
sy-records, line 111: The title and description of the modified homepage are both `PHP`.

include/layout.inc
derickr: What would you like to see instead?

TimWolla: The old generic description ("PHP is a popular general-purpose scripting language that powers everything from your blog to the most popular websites in the world.") would work (or just nothing).

sy-records:
```suggestion
    $default_description = "PHP is a popular general-purpose scripting language that powers everything from your blog to the most popular websites in the world.";
    $meta_description = $config['description'] ?? $default_description;
```

sy-records: I agree with @TimWolla. We should keep a default description rather than making it identical to the title.

derickr: I did too. I had made the change locally, but had pushed to the wrong repository.

TimWolla (approved):

PR #1390, opened 2025-09-06
Fix GPG keys link to use the current request URI

### LINKED ISSUE #1389: Links to GPG keys no longer work
### Summary

GPG keys should be properly linked from the source tarball page

### Details

The links on the source page for PHP 8.3's can be seen here: https://www.php.net/downloads.php?source=Y 

If you scroll down to php 8.3.25, there is a link to "GPG Keys for PHP 8.3". This takes the user to this page:

https://www.php.net/downloads.php#gpg-8.3

The page does not contain the GPG keys, so users who choose to verify the code cannot do so. 

### Additional context

_No response_

sy-records:
Close #1389

### DIFF
--- a/include/version.inc
+++ b/include/version.inc
@@ -153,7 +153,7 @@ function show_source_releases()
               </li>
             </ul>
 
-            <a href="#gpg-<?php echo $mver; ?>">GPG Keys for PHP <?php echo $mver; ?></a>
+            <a href="<?php echo $_SERVER['REQUEST_URI']; ?>#gpg-<?php echo $mver; ?>">GPG Keys for PHP <?php echo $mver; ?></a>
           </div>
         <?php endforeach; ?>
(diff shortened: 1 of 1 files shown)


saundefined (approved):
Looks good, thanks!

include/version.inc
```diff
@@ -153,7 +153,7 @@ function show_source_releases()
               </li>
             </ul>
 
-            <a href="#gpg-<?php echo $mver; ?>">GPG Keys for PHP <?php echo $mver; ?></a>
+            <a href="<?php echo $_SERVER['REQUEST_URI']; ?>#gpg-<?php echo $mver; ?>">GPG Keys for PHP <?php echo $mver; ?></a>
```
derickr, line 156: This introduces an XSS issue, wouldn't it?

cmb69: Good catch!

PR #1397, opened 2025-09-07
Add docker to the downloads page.

shivammathur:


include/download-instructions/linux-docker-cli-default.php
```diff
@@ -0,0 +1,10 @@
+<p>
+On the command line, run the following commands:
+</p>
+<pre><code class="language-bash line-numbers">
+# Pull the PHP Docker image.
+docker pull php:cli
+
+# Launch a container with PHP.
+docker run --rm -it --entrypoint bash php:cli
```
derickr: I think that should be `php-cli` instead of `php:cli` ?

include/download-instructions/linux-docker-web-default.php
```diff
@@ -0,0 +1,10 @@
+<p>
+On the command line, run the following commands:
+</p>
+<pre><code class="language-bash line-numbers">
+# Pull the PHP Docker image.
+docker pull php:fpm
+
+# Launch a container with PHP.
+docker run --rm -it --entrypoint bash php:fpm
```
derickr: And this ought to be `php-fpm` then, I think?

include/download-instructions/osx-docker.php
```diff
@@ -0,0 +1,10 @@
+<p>
+On the command line, run the following commands:
+</p>
+<pre><code class="language-bash line-numbers">
+# Pull the PHP Docker image.
+docker pull php:<?= $version; ?>-cli
+
+# Launch a container with PHP.
+docker run --rm -it --entrypoint bash php:<?= $version; ?>-cli
+</code></pre>
```
derickr: This seems to be the same as include/download-instructions/linux-docker-cli-community.php, so make it a symlink (like we already use in other places)

include/download-instructions/windows-docker.php
```diff
@@ -0,0 +1,10 @@
+<p>
+On the command line, run the following commands:
+</p>
+<pre><code class="language-bash line-numbers">
+# Pull the PHP Docker image.
+docker pull php:<?= $version; ?>-cli
+
+# Launch a container with PHP.
+docker run --rm -it --entrypoint bash php:<?= $version; ?>-cli
+</code></pre>
```
derickr: This seems to be the same as include/download-instructions/linux-docker-cli-community.php, so make it a symlink (like we already use in other places)

include/download-instructions/linux-docker-cli-default.php
shivammathur: Fixed

include/download-instructions/osx-docker.php
shivammathur: Converted to a symlink

include/download-instructions/windows-docker.php
shivammathur: Converted to a symlink

include/download-instructions/linux-docker-web-default.php
shivammathur: Fixed

PR #1410, opened 2025-09-10
Fix docker instructions for web in downloads page

shivammathur:
This PR fixes the instructions to use the fpm image when web-development is selected on Windows and macOS.


sy-records:
Using `Squash and merge` during merging is preferable. 😄

shivammathur:
@sy-records Will use that next time.

Issue #1448, opened 2025-09-23
Add instructions for Mezzio and Windows WSL AlmaLinux


arhimede:
### Summary

Add a new section for Mezzio and AlmaLinux WSL

### Details

I want to add 2 new items on the download page:
- How to start using Mezzio ( similar to Symfony instructions ) 
- How to install AlmaLinux on Windows WSL

cmb69:
Consider to submit a pull request.

arhimede:
> Consider to submit a pull request.

I will submit a PR, yes, is already prepared.

ibraheemaibraheem1202-gif:
> ### Summary
> Add a new section for Mezzio and AlmaLinux WSL
> 
> ### Details
> I want to add 2 new items on the download page:
> 
> * How to start using Mezzio ( similar to Symfony instructions )
> * How to install AlmaLinux on Windows WSL
> 
> ### Additional context
> _No response_

By using Chrome

derickr:
As explained in the PR (#1452) we are not doing this for now.

PR #1452, opened 2025-09-24
Add Mezzio and AlmaLinux WSL2 to download page

arhimede:
This PR addresses the issue https://github.com/php/web-php/issues/1448

### DIFF
--- a/include/download-instructions/windows-wsl-almalinux.php
+++ b/include/download-instructions/windows-wsl-almalinux.php
@@ -1,0 +1,6 @@
+<p>
+Instructions for installing PHP for AlmaLinux 9 and 10 on Windows WSL2 can be found on:
+</p>
+<p>
+» <a href='https://docs.dotkernel.org/development/'>https://docs.dotkernel.org/development/</a>
+</p>
--- a/downloads-get-instructions.php
+++ b/downloads-get-instructions.php
@@ -26,7 +26,7 @@ if ($options['os'] === 'osx' || $options['os'] === 'windows') {
     }
 }
 
-if (in_array($options['usage'], ['fw-drupal', 'fw-laravel', 'fw-symfony', 'fw-wordpress', 'fw-joomla'])) {
+if (in_array($options['usage'], ['fw-drupal', 'fw-laravel', 'fw-symfony', 'fw-wordpress', 'fw-joomla', 'fw-mezzio'])) {
     $file = "{$options['usage']}";
     $options['os'] = null;
 }
--- a/downloads.php
+++ b/downloads.php
@@ -57,6 +57,7 @@ $usage = [
     'fw-drupal' => 'Drupal Development',
     'fw-joomla' => 'Joomla Development',
     'fw-laravel' => 'Laravel Development',
+     'fw-mezzio' => 'Mezzio Development',
     'fw-symfony' => 'Symfony Development',
     'fw-wordpress' => 'WordPress Development',
 ];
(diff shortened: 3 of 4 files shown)


include/download-instructions/windows-wsl-almalinux.php
```diff
@@ -0,0 +1,6 @@
+<p>
+Instructions for installing PHP for AlmaLinux 9 and 10 on Windows WSL2 can be found on:
+</p>
+<p>
+» <a href='https://docs.dotkernel.org/development/'>https://docs.dotkernel.org/development/</a>
+</p>
```
shivammathur: While this is a great resource, it will be a good idea to have actual commands to setup PHP packages similar to how we have it for fedora and Red Hat.

Also, please add linux instructions for AlmaLinux and you can symlink the same for WSL.

shivammathur (changes requested):

include/download-instructions/windows-wsl-almalinux.php
arhimede: > While this is a great resource, it will be a good idea to have actual commands to setup PHP packages similar to how we have it for fedora and Red Hat.
> 
> Also, please add linux instructions for AlmaLinux and you can symlink the same for WSL.

Done, and also I reorganised the things a bit

shivammathur:
@arhimede 

Hi!

Thanks for the PR. Right now, we're not going to merge this.

We’re in the process of defining policy for adding new frameworks and operating systems. Right now, this PR would also make Alma the default for Linux, and that is probably not useful.

Guidelines that we are considering, include:

- The page it links to needs to have in detail how to set up all the parts to get to a working set-up, including database, web server, etc. The link for Mezzio would fall foul of this, as it refers back to the PHP documentation. This is also already an issue with the other frameworks, and we will be working with them to see to improve this documentation.
- Notoriety — we can't really expand the list to include every framework or OS variant, it would become too unwieldy.

arhimede:
> We’re in the process of defining policy for adding new frameworks and operating systems. Right now, this PR would also make Alma the default for Linux, and that is probably not useful.

Well, I can add AlmaLinux at a bottom of that drop-down, to not be the first one in the list. :-) 
AlmaLinux have a reasonable good market share as a RedHat clone and it replaced Centos on dedicated servers in past years. 
We may want to avoid the vendor lock-in situation . 

> Guidelines that we are considering, include:
> 
>    The page it links to needs to have in detail how to set up all the parts to get to a working set-up, including database, web server, etc. The link for Mezzio would fall foul of this, as it refers back to the PHP documentation. This is also already an issue with the other frameworks, and we will be working with them to see to improve this documentation.

That's correct, most of the frameworks that are already listed does not have proper instructions for PHP  installation.
 I simply followed them. 
I am waiting then for the guidelines, meantime we will extend Mezzio documentation.

>    Notoriety — we can't really expand the list to include every framework or OS variant, it would become too unwieldy.

My bad, this PR address  2 separate items ( Mezzio and AlmaLinux) , it was better to have 2 separate PR's , and discuss each items individually.

arhimede:
1. I removed instructions for Mezzio until we will extend that to show how to complet setup a development environment. Will be a separate PR . 
2. I moved AlmaLinux lower in the Linux drop-down , to avoid being the default one. 
3. AlmaLinux instructions are one of the very few that contains instructions how to setup a complete dev env.

**Note:**
There are 2 major server control panels on the market right now : CPanel and Plesk.
CPanel supports only Ubuntu and RedHat clones ( AlmaLinux, Rocky Linux, etc)
Plesk have support only for Ubuntu, Debian, AlmaLinux, RedHat 
I consider that ALmaLinux is one of the important OS for servers.

derickr:
It seems we never actually closed this PR, so doing so.

PR #1460, opened 2025-09-27
Fix Docker image in linux-docker-web-default.php

TimWolla:


include/download-instructions/linux-docker-web-default.php
```diff
@@ -3,8 +3,8 @@
 </p>
 <pre><code class="language-bash line-numbers">
 # Pull the PHP Docker image.
-docker pull php-fpm
+docker pull php:fpm
```
TimWolla, line 7: The explicit `docker pull` is not even necessary. `docker run` will pull implicitly (but will not update when already pulled).

sy-records (approved):

PR #1463, opened 2025-09-28
interactive-examples.js: Enable for informal examples

AllenJB:
eg. as found on https://www.php.net/DatePeriod

FAO @derickr


AllenJB:
~~Looking into as issue with replacing the output (fails to show generated output at all)~~
Fixed output replacement for informal examples and checked it still works with non-informal examples.

PR #1501, opened 2025-10-09
Fix example code block padding for PHP 8.3+

ElfSundae:
This pull request fixes the missing padding in example code blocks introduced by changes in PHP 8.3+.  
Since `highlight_string()` now returns `<pre><code>...</code></pre>` instead of `<code>...</code>`, the existing CSS rule targeting only `<code>` no longer applied.  

截屏2025-10-09 19 46 40

The fix adds `<pre>` to the selector so both cases are covered:

```diff
+.docs .example-contents > .phpcode > pre,
 .docs .example-contents > .phpcode > code {
     padding: .75rem;
 }
```

This ensures that the padding style is correctly applied to all example code blocks.

### DIFF
--- a/styles/theme-base.css
+++ b/styles/theme-base.css
@@ -948,7 +948,7 @@ div.tip p:first-child {
     overflow-x: auto;
 }
 
-.docs .example-contents > .phpcode > pre,
+.docs .example-contents > .phpcode > pre > code,
 .docs .example-contents > .phpcode > code {
     padding: .75rem;
 }


styles/theme-base.css
```diff
@@ -948,6 +948,7 @@ div.tip p:first-child {
     overflow-x: auto;
 }
 
+.docs .example-contents > .phpcode > pre,
```
sy-records:
```suggestion
.docs .example-contents > .phpcode > pre > code,
```

sy-records (changes requested):

styles/theme-base.css
ElfSundae: Thanks 👍

sy-records (approved):

sy-records:
Thanks!

PR #1525, opened 2025-10-19
(Docs) CSS stylesheets: Fix for PHP 8.3+ highlight_string changes

AllenJB:
Related: https://github.com/php/phd/issues/208

When generating docs with PHP 8.3+, the examples have lost the padding due to changes in the code generated by highlight_string.

This currently only affects users generating their own docs, but will affect the web docs when the PHP version used to generate those is updated.

I tested these changes locally and with the current live web docs (using stylish) and did not find anywhere it interfered with the current styles.

### DIFF
--- a/styles/theme-base.css
+++ b/styles/theme-base.css
@@ -948,7 +948,7 @@ div.tip p:first-child {
     overflow-x: auto;
 }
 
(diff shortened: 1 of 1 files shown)


AllenJB:
Duplicate of #1501

PR #1586, opened 2025-11-03
Update search to use new pre-combined indexes + local docs support

AllenJB:
This PR is related to https://github.com/php/phd/pull/212 and https://github.com/php/phd/pull/204

**IMPORTANT:** At least one of the above 2 PRs to phd MUST be merged before this one.

These changes modify search.js to use the new pre-combined search indexes (significantly simplifying the code) added by both the above mentioned PRs.

Additionally the changes required for [supporting the search on local docs](https://github.com/php/phd/pull/204) (which set the language to `local` to trigger the alternative behavior)

The version of `search.js` used here is the same as that included in https://github.com/php/phd/pull/204 . After this PR has been merged, phd can be modified to pull this copy of `search.js` instead of duplicating it in the phd repository, the same as it does for stylesheets.

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -20,7 +20,7 @@ const initPHPSearch = async (language) => {
         const key = `search2-${language}`;
         const cache = window.localStorage.getItem(key);
 
-        if (!cache) {
+        if ((!cache) || (language === 'local')) {
             return null;
         }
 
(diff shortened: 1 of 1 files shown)


theodorejb:
It seems like this broke the visual snapshot tests, since they are failing on all PRs after this was merged. Actual image:
actual

Expected image:
expected

derickr:
@AllenJB I hadn't noticed this made the tests fail. Can you have a look?

AllenJB:
@derickr I don't believe my changes caused these test failures.

Based on manual testing, the code is doing what it always does and correctly sending the user to `/search.php?lang={language}&q={query}`, but that page then changes the URL. I believe this is likely caused by the Google CSE scripts.

You can see this happen on php.net itself (in addition to testing with the `php -S` local setup)

My guess is that there's been a change in the Google CSE scripts causing this (hence why the tests passed when this PR was merged).

Based on the [PR Preview Action Logs](https://github.com/php/web-php/actions/workflows/pr-preview.yml) this occurred some time on Nov 4th between 11am and 8pm GMT (between runs for PRs 665 and 666)

My guess is the test needs to be updated to account for the Google CSE changes, but I'm not sure the best way to do that here.

sy-records:
@saundefined Could you take a look? I don't have server permissions. I suspect it might be an issue with the script? https://github.com/php/web-php/blob/7d87155d121e27b4b4e4d7485c217799e86e2397/.github/workflows/pr-preview.yml#L21-L22

The `search-combined.json` file wasn't copied?

saundefined:
@sy-records the `search-combined.json` is [generated](https://web-php-pr-1616.preview.thephp.foundation/manual/en/search-combined.json), and the search works on the preview boxes.
I'll try to see what's wrong with the tests 🤔

Issue #1593, opened 2025-11-05
Search: Index Caching Issues


AllenJB:
The way the search caches indexes is, in my opinion, not ideal.

# Background

The search indexes are ~2.1 MB (.json file size) and localStorage quota is (generally) 5MB according to https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria

The current search uses localStorage and caches the indexes with a key for each specific language.

This means that if a user uses the manual in multiple languages, they can store at most 2 before hitting the localStorage quota.

This doesn't stop search working in further languages - the indexes of further visited languages just don't get cached.

However, if a user visits a language, then never returns to it, the indexes for that language are never considered for deletion, and data in localStorage never expires otherwise (unless the user manually clears their data, or the browser decides to delete data from a site not recently visited - I don't know the specifics of browser behavior in this regard)

This means that if the user regularly uses php.net, that (old) unused data remains in localStorage forever, preventing the storage of further data. At the moment, based on a quick search of the web-php repo, this is just other search indexes, but could be used for other functionality in the future.

# Alternative Options

I'm not very familiar with client-side, but based on [some (relatively quick) research](https://web.dev/articles/storage-for-the-web)  I think that if this functionality is to be retained, it should be rewritten to use either Cache API or OPFS. These seem to be designed to do exactly what the search is using localStorage for, and the storage quotas are _much_ larger. 

An advantage of going with OPFS would be that it would be trivial to manage cached indexes for all languages (listing them in a directory) and deleting unused ones, if that's desirable.

The other alternative would be to get rid of JS caching altogether and just let the browser handle caching as with any other web request.

Does anyone else have thoughts on this?

haszi:
If I understand the function you linked to correctly, we are re-creating the functionality of the `Expires` HTTP header. If that is so, wouldn't it be easier/simpler to use that header and let the browser handle the caching?

Issue #1589, opened 2025-11-04
Search is broken


php4fan:
### Affected Page URL

https://www.php.net/

### Describe the bug

The search is broken

### Steps to reproduce

Type 'array_unique' or 'shuffle' (or any other function that exists) in the search box. Do NOT hit Enter.

### Expected behavior

Should show results

### Screenshots

<img width="1411" height="923" alt="Image" src="https://github.com/user-attachments/assets/1c6a6b2a-68cc-469e-8a93-5f541e876591" />

<img width="1051" height="900" alt="Image" src="https://github.com/user-attachments/assets/bd9c187e-c3a4-4588-8447-8dd2fcadf53c" />

### Additional context

Errors in JavaScript console:

<img width="687" height="199" alt="Image" src="https://github.com/user-attachments/assets/28dac3c9-ff5f-4ed5-86dd-cfbf415e2ef8" />

sy-records:
Clearing the browser's local storage and reacquiring it can resolve the issue.

PR #1594, opened 2025-11-05
Search: Clear any old caches to free up quota

AllenJB:
Related to #1589 

This change removes any old caches before storing new data. This reduces the chances of hitting quota.

(This won't happen immediately for all users - this check is only done when either there is no current cached indexes or the cache expires and the script wants to fetch new data, which currently happens every 14 days.

I've also changed the message when quota is exceeded to an info message, since it doesn't actually affect functionality.

Note that it's still possible for users to hit quota even disregarding old cached values if they use the manual in multiple languages.

The search indexes are ~2.1 MB and localStorage quota is (generally) 5MB according to https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria

Based on this (any my own testing) users can store the indexes for 2 languages at most.

I further explore this issue in #1593

### DIFF
--- a/js/search.js
+++ b/js/search.js
@@ -35,12 +35,26 @@ const initPHPSearch = async (language) => {
         return data;
     };
 
+    const cleanupOld = async() => {
+        // Previously used cache key
+        const key = `search-${language}`;
+        const oldCache = window.localStorage.getItem(key);
+        if (!oldCache) {
+            return;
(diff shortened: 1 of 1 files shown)


derickr (approved):
LGTM, but I don't really know how this works.

PR #1607, opened 2025-11-14
Add windows QA builds

shivammathur:
This PR
- Adds windows QA builds
- Renames `release-candidates.php` to `pre-release-builds.php`.
- Adds an entry to `error.php` to redirect after the rename.
- Adds support for redirecting with query string in `error.php` as `release-candidates.php` was also an API with query strings.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -94,7 +94,7 @@ paths:
 
     "/pre-release-builds.php":
         get:
-            summary: "Currently Pre-release versions of PHP."
+            summary: "Current pre-release versions of PHP."
             parameters:
                 - in: query
                   name: format


openapi.yml
```diff
@@ -92,9 +92,9 @@ paths:
                                             type: string
                                             format: date-time
 
-    "/release-candidates.php":
+    "/pre-release-builds.php":
         get:
-            summary: "Currently RC versions of PHP."
+            summary: "Currently Pre-release versions of PHP."
```
derickr: This probably should say "Current" instead of "Currently".

shivammathur: Fixed

shivammathur:
@derickr Test failures are related to the changes in #1586. Merging this for now.

PR #1612, opened 2025-11-16
Render feature comparisons with standard function

theodorejb:
This refactor greatly simplifies maintaining consistent styling and markup for the feature comparisons.

No content has been changed, except for fixing unnecessary blank lines at the start & end of the "Command line linter supports multiple files" code blocks on the PHP 8.3 release page.


derickr:
Is this still relevant, as the PHP 8.5 (and I suppose later) release announcements don't use this any more, and the existing ones are supposed to be static?

theodorejb:
It simplifies maintaining the pages in a consistent way. If we want to update the whole site to a new design, this will make it much easier.

PR #1616, opened 2025-11-17
Update security.txt for another year

DanielEScherzer:
Updated security.txt expiration date as suggested by the release process, doing this a few days before the announcement of the initial stable version of 8.5 to having another thing to worry about.


sy-records (approved):

adoy (approved):

DanielEScherzer:
rebased, now merging locally so that the merged commit is also GPG signed with my key

PR #1618, opened 2025-11-17
Fix Playwright search tests with static copy of new search indexes

AllenJB:


sy-records:
I think we should include a test file instead of the full JSON. Or download before testing

AllenJB:
I took another look at the failures and realized I had focused on the wrong failure. This restores the visual tests to include search results (rather than empty results).

I've simply done what the tests previously did and included a static copy of the indexes here. I updated the snapshot image as the results differ from before due to my fixes to include missing entries in the search results.

If the indexes were to be downloaded for the tests, what would be the best way (and command) to do that? (Are there any preferences / restrictions?) I would suggest it should be in the Makefile (rather than part of the GH Actions) so that it works for local testing. I had a brief look at the Makefile but couldn't see anywhere that's currently doing anything like this.

PR #1625, opened 2025-11-19
Update foundation note on the release page

pronskiy:


TimWolla (approved):
This is better. The previous text was solely focused on the Foundation, which is not correct for php.net.

PR #1127, opened 2024-11-19
Add Ukrainian translation of PHP 8.4 announcement page

bkosun:
Ukrainian translation of PHP 8.4 announcement page

PR #1633, opened 2025-11-20
Improve Ukrainian translation of PHP 8.5 announcement page

bkosun:
Ukrainian translation of PHP 85 announcement page


releases/8.5/languages/uk.php
```diff
@@ -1,79 +1,79 @@
...
+    'bc_casting_non_int_floats' => 'Приведення чисел з плаваючою крапкою (або рядків, що їх представляють) до цілого типу тепер викликає попередження, якщо їх неможливо представити як ціле число.',
 
-    'footer_title' => 'Кращий синтаксис, вища продуктивність і краща типобезпечність.',
-    'footer_description' => '<p class="first-paragraph">Повний список змін опубліковано у <a href="/ChangeLog-8.php#PHP_8_5" target="_blank">ChangeLog</a>.</p><p>Будь ласка, ознайомтеся з <a href="/manual/en/migration85.php" target="_blank">посібником з міграції</a> для детального переліку нових можливостей і несумісностей.</p>',
+    'footer_title' => 'Кращий синтаксис, покращена продуктивність і безпека типів.',
+    'footer_description' => '<p class="first-paragraph">Повний перелік змін описано на сторінці <a href="/ChangeLog-8.php#PHP_8_5" target="_blank">ChangeLog</a>.</p><p>Будь ласка, ознайомтеся з <a href="/manual/uk/migration85.php" target="_blank">посібником з міграції</a>, щоб отримати детальніший список нових функцій і несумісних змін.</p>',
```
saundefined, line 78: https://www.php.net/manual/uk/migration85.php

Migration guide returns 404.
We can leave it as is for now, and once the documentation is updated, replace it with `uk`.

What do you think?

Otherwise looks good, thanks!

bkosun: It looks like an bug on the php.net side, since the link for previous versions of PHP works despite the fact that there is no translation:

https://www.php.net/manual/uk/migration84.php

We also decided earlier to use `uk` instead of `en` despite the actual lack of documentation:
https://github.com/php/web-php/pull/1127#issuecomment-2486727053

PR #1649, opened 2025-11-21
php/php-src#17725 add info about substituting static with self in Cha…

rekmixa:
Added implement of feature described in this issue: https://github.com/php/php-src/issues/17725


rekmixa:
@DanielEScherzer any review?

derickr:
This is not in the NEWS file, which is probably why it wasn't picked up. I'll leave it to @php/release-managers-85 to decide whether this should be here, as it would cause an inconsistency.

edorian (approved):

edorian:
This feel through the cracks. Sorry for that. Merging so we have it in one list at least

PR #1731, opened 2026-01-01
Fase out http links to https links where possible

jordikroon:
When I was planning to do this I did not expect this commit to be so long. I am sorry for that. 
That said I was also very surprised with the amount of links still on http. Even on non-archive pages. So my goal was to update those in order to medernize the PHP.net website.

I updated all *.php.net links to it's respectful https link. Excluded XML namespaces that point to `http://php.net/ns`.
There are some subdomains that are not available anymore, which could still be relevant. Those I will send through the mailing list. 

Additionally some external links have also been replaced. Those have been manually verified before changed and by far not all the links have been updated due to the manual labor it requires.

And lastly; docs.php.net links that have been updated to www.php.net


kamil-tekiela:
I have a deja vu. I believe someone proposed it in the past, and we finally declined the PR. Unless the link is broken, don't change the http to https. 

For broken URLs, you should open a single PR for each type of link. Don't put it all in a single PR because someone needs to review it. Nobody is going to read through 12000 lines of code to verify that you replaced the broken links properly. Open a PR and in the description, put which regex you used to replace the broken links. 

I see that you have also changed php.net to www.php.net. <s>This site doesn't use that subdomain and we have an automatic redirection set up to the root domain. Therefore, do not put www in the URLs.</s> It seems it's actually the other way around, but in that case there is no reason to add www to working URLs either. As long as both point to the same, it's fine.

jordikroon:
Your dejavu is right. Actually it has been proposed a few times and I should have checked that before creating this PR.
 - https://github.com/php/web-php/pull/287
 - https://github.com/php/web-php/pull/623
 
 I admit too it is hard to review. So I will create different PR's for the various URLs. That way we also have more control over legacy URLs as some URLs do not exist anymore. It would be nice to preserve history in order to keep the archive in better shape.

PR #1739, opened 2026-01-01
Update qa.php.net links to https

jordikroon:
Replace http://qa.php.net URLs with https://qa.php.net.

Side notes:
 - The URL http://qa.php.net/~derick/ has been changed to http://downloads.php.net/~derick/ even though both URLs don't include 4.4.0RC2. My assumption is that if we ever restore old RC's it would be there.
 - http://qa.php.net/testfest.php has been changed to https://phptestfest.org.

### DIFF
--- a/archive/2008.php
+++ b/archive/2008.php
@@ -442,7 +442,7 @@ site_header("News Archive - 2008", ["cache" => true]);
     <div class="entry-content description">
         <abbr class="published newsdate" title="2008-08-01T08:50:37+02:00">[01-Aug-2008]</abbr>
         <div>
-        <p>The PHP development team is proud to announce the <a href="https://qa.php.net/">first alpha release</a> of the upcoming minor version update of PHP. <a href="http://windows.php.net/download/">Windows binaries</a> will be available starting with alpha2 (intermediate snapshots available at <a href="http://snaps.php.net">snaps.php.net</a>). The new version PHP 5.3 is expected to improve stability and performance as well as add new language syntax and extensions. Several new features have already been documented in the <a href="http://php.net/docs.php">official documentation</a>, others are listed on the <a href="https://wiki.php.net/doc/scratchpad/upgrade/53">wiki</a> in preparation of getting documented. Please also review the <a href="http://php.net/php5news">NEWS</a> file.</p>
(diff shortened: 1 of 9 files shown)


archive/2008.php
```diff
@@ -482,7 +482,7 @@
...
+        <p>Overall 158 tests have been submitted as part of <a href="https://phptestfest.org">TestFest 2008</a> since the launch of the <a href="http://testfest.php.net">TestFest submission site</a> by 30 different people from people all over the world. Actually this is not counting the various submissions by existing core developers, who also took this opportunity to add some more tests. This has actually increased total <a href="http://gcov.php.net/" target="_new">test coverage</a> for ext/reflection, ext/dom and ext/exif by about 10% each. While the organization of the TestFest was a bit adhoc, there were numerous TestFest events in local user groups. So the number of people exposed to the PHP test framework is much greater. Hopefully this will lead to more people submitting bug reports with an accompanying <a href="https://qa.php.net/write-test.php">phpt</a> test file!</p>
```
derickr: This test fest is for a different one (2017) and not 2008. We also don't control the domain, so I don't think we should be linking to phptestfest.org any more.

archive/2008.php
```diff
@@ -442,7 +442,7 @@
...
+        <p>The PHP development team is proud to announce the <a href="https://qa.php.net/">first alpha release</a> of the upcoming minor version update of PHP. <a href="http://windows.php.net/download/">Windows binaries</a> will be available starting with alpha2 (intermediate snapshots available at <a href="http://snaps.php.net">snaps.php.net</a>). The new version PHP 5.3 is expected to improve stability and performance as well as add new language syntax and extensions. Several new features have already been documented in the <a href="http://php.net/docs.php">official documentation</a>, others are listed on the <a href="http://wiki.php.net/doc/scratchpad/upgrade/53">wiki</a> in preparation of getting documented. Please also review the <a href="http://php.net/php5news">NEWS</a> file.</p>
```
derickr: snaps.php.net no longer exists either.

There are links to `http://php.net` instead of `https://php.net`.

derickr (changes requested):

derickr:
@jordikroon Are you going to fix the few mistakes in here?

jordikroon:
Sorry for the delay. I got a little side tracked (plus holidays). I have addressed the issues. 
Links to php.net have intentionally not been touched as I was planning to create a separate PR for that.

archive/2008.php
jordikroon: I didn't know. I have removed those links all together now since the old URL gave a 404 anyway.

derickr:
I've merged this, but qa.php.net is no longer something we run. However, we did set up redirects so the links aren't dead.

PR #1741, opened 2026-01-02
Remove error suppression operator (@) and add proper error handling

jordikroon:
At some places the `@` symbol was still used. This PR changes that behaviour with proper handling.

Since those checks are always (correct me if I am wrong) for internal paths. A simple file_exists & is_readable check is sufficient. If not we could override the error handler instead.

### DIFF
--- a/cal.php
+++ b/cal.php
@@ -253,8 +253,11 @@ function display_events_for_day($day, $events): void
 // Find a single event in the events file by ID
 function load_event($id)
 {
+    $path = "backend/events.csv";
+    if (!file_exists($path) || !is_readable($path)) { return false; }
+
     // Open events CSV file, return on error
-    $fp = @fopen("backend/events.csv",'r');
+    $fp = fopen($path,'r');
     if (!$fp) { return false; }
 
     // Read as we can, event by event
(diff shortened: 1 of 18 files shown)


include/file.inc
```diff
@@ -0,0 +1,34 @@
...
+    return filemtime($filename);
+}
+
+function file_get_contents_if_exists(string $filename): string|false
+{
+    if (!file_exists($filename) || !is_readable($filename)) {
+        return false;
+    }
+
+    return file_get_contents($filename);
```
derickr, line 33: There is a race condition here, if the file gets removed between the `file_exists`/`is_readable`, and the `file_get_contents`. This is quite possible, as the rsync process to put new data on the servers is not atomic. I would recommend keeping the `@` here.

include/shared-manual.inc
```diff
@@ -101,11 +101,11 @@ function manual_notes_load(string $id): array
 
     // Open the note file for reading and get the data (12KB)
     // ..if it exists
-    if (!file_exists($notes_file)) {
+//    if (!file_exists($notes_file) || !is_readable($notes_file)) {
```
derickr, line 104: This introduces a parse error, as the `{` is now gone matching the `}` on line 106.

cal.php
```diff
@@ -253,8 +253,11 @@ function display_events_for_day($day, $events): void
 // Find a single event in the events file by ID
 function load_event($id)
 {
+    $path = "backend/events.csv";
+    if (!file_exists($path) || !is_readable($path)) { return false; }
+
     // Open events CSV file, return on error
-    $fp = @fopen("backend/events.csv",'r');
+    $fp = fopen($path,'r');
```
derickr, line 260: There are race conditions here too for similar reasons.

derickr (changes requested):

derickr:
@jordikroon Would you have some time to look at my comments? Otherwise I think we should close this and just leave the `@` in place (or really, we need a better function in PHP for these).

jordikroon:
For now I will close this. Given the redesign it's not a good moment to push this through anyway.

PR #1762, opened 2026-01-11
license-update: update website with new license info

ramsey:
This PR encompasses the changes proposed in the [PHP License Update](https://wiki.php.net/rfc/php_license_update) RFC, for the PHP website.

Discussion is available on the [internals mailing list](https://news-web.php.net/php.internals/127984).

Changes proposed for php-src are available at https://github.com/php/php-src/pull/20904.

### DIFF
--- a/license/contrib-guidelines-code.php
+++ b/license/contrib-guidelines-code.php
@@ -1,32 +1,31 @@
 <?php
 $_SERVER['BASE_PAGE'] = 'license/contrib-guidelines-code.php';
 include_once __DIR__ . '/../include/prepend.inc';
-site_header("License Information", ["current" => "help"]);
+site_header("Contributor Guidelines", ["current" => "help"]);
 ?>
 <h2>PHP Contributor Guidelines for Code Developers</h2>
 
 <p>
-  Before you complete the form to request a Git account, you must understand and accept the principles under which PHP
+  Before you contribute code to PHP, you must understand and accept the principles under which PHP
   itself is developed. These are summarised in the next paragraph.
 </p>
 <p>
-  Any code contributed to PHP must be contributed under the PHP license (or a compatible BSD-style license).
-  PHP will be redistributed under the PHP license (currently
-  <a href="http://www.php.net/license/3_01.txt">version 3.01</a>).
+  Any code contributed to PHP must be contributed under the Modified BSD License or other, compatible permissive license.
+  PHP is distributed under the <a href="/license/#bsd-3-clause">Modified BSD License</a>.
   This includes implementation code, test cases, utility scripts and tools - that is, all code associated
   with the PHP project. If you contribute code that isn't entirely your own
   (for example it may be partially derived from other Open Source software) you are asked to add a comment
   into the source code to indicate the origin and the license of the original code.
-  If you are unsure of the license you are asked to confirm with the owner of the code that they are
-  happy for it to be contributed to PHP under the PHP license.
+  If you are unsure of the license, you are asked to confirm with the owner of the code that they
+  grant permission for it to be contributed to PHP under the
+  terms of the Modified BSD License or other compatible license.
 </p>
 <p>
-  Specifically regarding GPL and LGPL source code:
+  Specifically regarding source code licensed under the GNU General Public License (GPL):
 </p>
 <ul>
-  <li>GPL or LGPL licensed code cannot be used as a basis for any derived work contributed to PHP.</li>
-  <li>Extensions which link  GPL'd libraries will not be accepted.</li>
-  <li>Extensions which link to LGPL libraries  will be strongly discouraged.</li>
+  <li>GPL-licensed code cannot be used as a basis for any work contributed to PHP.</li>
+  <li>Extensions linking to GPL-licensed libraries will not be accepted.</li>
 </ul>
 
 
--- a/license/distrib-guidelines-code.php
+++ b/license/distrib-guidelines-code.php
@@ -1,14 +1,13 @@
 <?php
 $_SERVER['BASE_PAGE'] = 'license/distrib-guidelines-code.php';
 include_once __DIR__ . '/../include/prepend.inc';
-site_header("License Information", ["current" => "help"]);
+site_header("Distribution Guidelines", ["current" => "help"]);
 ?>
 <h2>PHP Distribution Guidelines</h2>
 
 <p>
-  The PHP code can be redistributed in source or binary form provided that
-  the terms of the PHP license are respected (see
-  <a href="http://www.php.net/license/3_01.txt">version 3.01</a>).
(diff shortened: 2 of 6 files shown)


license/contrib-guidelines-code.php
```diff
@@ -36,8 +35,8 @@
   For related projects, please refer to the Project websites:
 </p>
 <ul>
-  <li><a href="http://pecl.php.net/copyright.php">PECL</a></li>
-  <li><a href="http://pear.php.net/copyright.php">PEAR</a></li>
+  <li><a href="https://pecl.php.net/copyright.php">PECL</a></li>
+  <li><a href="https://pear.php.net/copyright.php">PEAR</a></li>
```
derickr, line 39: The PEAR server might possibly no longer exist by this Friday, and maybe we should just remove this section now? PECL also does no longer accept new packages either.

license/distrib-guidelines-code.php
```diff
@@ -1,65 +1,59 @@
...
-  A text file containing license and copyright information is sometimes
-  given the filename "Notices", and may be referred to as a "Notices file".
+    In plain English, this means that you have to include the full license text
+    in human-readable form with every distributed copy of PHP, whether in source
+    or binary form. One way of doing this is to put a copy of the license in a
+    text file that you include with the source or binary package before distribution.
+    This ensures that the license information can be read properly even when a
+    binary is shipped instead of source code. A text file containing license and
+    copyright information is sometimes given the filename "Notice" or "NOTICE,"
+    and may be referred to as a "notice file."
```
derickr: nit: The `,` in `"NOTICE,"` and the `.` in `"notice file."` should be outside the `"`'s

ramsey: Those silly American style guides! 😉

I generally follow the MLA style guide in writing, but the APA style guide and AP stylebook also agree that commas and periods should be placed inside closing quotation marks.

However, I'm aware that British English generally places them outside closing quotation marks. Is it the style of php.net to follow the British style?

license/2_01.txt
```diff
@@ -29,32 +30,32 @@ are met:
 
   3. The name "PHP" must not be used to endorse or promote products
      derived from this software without prior permission from the
-	 PHP Group.  This does not apply to add-on libraries or tools
```
DanielEScherzer, line 32: can we leave the whitespace changes to a separate PR (if needed at all)?

DanielEScherzer: actually, can all of the cleanup of the old licenses (commas, removal of "very", etc.) be done separate from the changes to add the new license?

ramsey: Yep. I can do that.

ramsey: I've removed the whitespace changes and other clean-up from the older license files.

license/contrib-guidelines-code.php
```diff
@@ -1,43 +1,32 @@
...
+    includes implementation code, test cases, utility scripts and tools&mdash;that is, all code associated with the PHP
+    project. If you contribute code that isn't entirely your own (for example, it may be partially derived from other
+    Open Source software), you should add a comment to the source code indicating the origin and license of the original
+    code. If you are unsure of the license, you must confirm with the owner that they grant permission to contribute it
+    to PHP under the terms of the Modified BSD License or other compatible license.
 </p>
-<ul>
-  <li>GPL or LGPL licensed code cannot be used as a basis for any derived work contributed to PHP.</li>
-  <li>Extensions which link  GPL'd libraries will not be accepted.</li>
-  <li>Extensions which link to LGPL libraries  will be strongly discouraged.</li>
```
DanielEScherzer, line 29: the LGPL library entry was removed - was this intentional?

also there are still whitespace changes that make things harder to review, e.g. the paragraphs above got indented more

ramsey: These changes are intentional. There's no legal reason code within the PHP source cannot link against the LGPL, and we shouldn't be in the business of discouraging people from using LGPL.

Here's a list (probably not exhaustive) of extensions currently in php-src that link against LGPL code (and, in some cases, bundle LGPL code within php-src):

- bcmath--we bundle libbcmath (LGPL) within php-src
- enchant
- gettext
- gmp
- iconv
- mbstring--we bundle libmbfl (LGPL) within php-src

Special mention:

readline links against the readline library, which is licensed under the GPL. However, you can also tell it to use libedit, which is BSD-licensed.

ramsey: I can try to refactor this to minimize whitespace differences, but it'll be a while.

ramsey:
~~I have attempted to reduce the whitespace changes and text changes as much as possible, as requested, but this brings this PR into a state that I disagree with and which was not part of my proposal. I'm not sure what to do with it at this point. I'll leave it here for others to merge, if they wish.~~

I'll ~~try to~~ fix it this week.

ramsey:
The most recent changes are my latest attempt to reduce white space changes and get the diff into a state that makes it easier to see the changes. I would still prefer to re-write all of this copy for better flow and correctness, but I decided to remove most of my original edits and stick primarily to changes to satisfy the RFC.

license/distrib-guidelines-code.php
ramsey: I removed my edits and fell back to using the British style.

license/contrib-guidelines-code.php
ramsey: I removed most of my edits and have gotten rid of all superfluous white space changes.

TimWolla:
> I would still prefer to re-write all of this copy for better flow and correctness, but I decided to remove most of my original edits and stick primarily to changes to satisfy the RFC.

That makes sense. Anything more can happen in a follow-up.

license/index.php
```diff
@@ -25,34 +25,107 @@
...
-    restrictions associated with GPL.
-   </li>
-  </ul>
+  Redistributions of source code must retain the above copyright notice,
+  this list of conditions and the following disclaimer.
  </li>
  <li>
-  Some files have been contributed under other (compatible) licenses and carry
-  additional requirements and copyright information.<br>
-  This is indicated in the license + copyright comment block at the top of the source file.
```
derickr, line 46: I don't think we can remove these three lines, as there is some code that comes from somewhere else, like:

- https://github.com/php/php-src/blob/master/ext/hash/xxhash/xxhash.h
- https://github.com/php/php-src/blob/master/ext/standard/crypt_blowfish.c
- https://github.com/php/php-src/blob/master/ext/bcmath/libbcmath/src/bcmath.h

(and there a whole bunch more)

ramsey: The diff is showing these lines as removed, but they were moved above (within a `<p>` tag instead of `<li>`).

derickr (approved):

PR #1775, opened 2026-01-18
Change language code to lowercase for translation file

leonardolara:
For languages like 'pt_BR', the filename is 'pt_br.ini' in lowercase.

The headers and footers of pt_BR pages are not being translated.


saundefined (approved):
Looks good, thanks!

PR #1781, opened 2026-01-21
Fix changelog file reference in error.php

sy-records:
searching for PHP changelog on google - the top result is: https://www.php.net/changelog-8.php  which doth not exist because it should be ChangeLog-8.php instead.


saundefined (approved):
Looks good, thanks!

PR #1812, opened 2026-02-10
Display conference dates on conferences page and homepage

### LINKED ISSUE #1606: Prominently display dates and locations of conferences
The list of conferences is difficult to scan for anything that may actually be relevant for you, since there is no easy way to identify when or where an event will take place. Some events include this info in their descriptions, some do not.

It would be great to just have date and location listed under the headline on the conferences page, and also have date and location listed right on the start page under the titles of the conferences. Then you can quickly see if the events may actually be relevant to you.

The homepage could also do with some indication that there may be more than the 3 events listed, as it's not clear the headline is actually a link to viewing more conferences.

lacatoire:
## Summary
- Fixes #1606
- **Conferences page**: displays `finalTeaserDate` prominently under each conference title and in the sidebar panel
- **Homepage**: replaces the outdated `$CONF_TEASER` array (stuck on 2015 data) with `NewsHandler::getConferences()` to access date information, and shows dates next to conference titles in the sidebar
- Adds a "View all" link when more than 4 conferences exist in a category


derickr (changes requested):
I don't mind adding the date, but I have to note:

- It is hard to see the difference between the title, and the date. Some sort of visual mark in between would be useful, I suppose.
- There are now conflicts due to code-reorganisation — that needs fixing first.
- I am not sure whether it currently makes sense to make changes to the content on the homepage, as there is a redesign under way.

PR #1838, opened 2026-02-26
Update the Installation Instructions for Debian

gdaffa:
Closes #1833
- Update Debian source list from `.list` to `.sources` format.
- Drop `apt-transport-https` package.

PR #1839, opened 2026-02-27
fix(css): Override `.token.operator` bg to `trasnparent`

gdaffa:
Fix visual error within the code block by removing `background` property from `.token.operator`.

Before:
bg-before-fix

After:
bg-after-fix

For anyone wondering `.token.operator` class is defined on `prism.css` line 3 column 1479.

### DIFF
--- a/styles/code-syntax.css
+++ b/styles/code-syntax.css
@@ -43,6 +43,10 @@ pre.line-numbers .line-numbers-rows > span:before {
     padding-right: 1em !important;
 }
 
+.token.operator {
+    background: transparent !important;
+}
+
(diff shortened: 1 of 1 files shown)


TimWolla:
This bundled file should not be modified, because this will make updates harder. Instead the CSS should be overridden in some other way.

gdaffa:
Oh im sorry i didnt know that, i also think that way but @shivammathur asked me to fix it on PR #1838 so i think its fine.

shivammathur:
@gdaffa Please use `styles/code-syntax.css` instead to override this.

gdaffa:
oh ok ill be right back

gdaffa:
done

PR #1856, opened 2026-03-09
Add FrankenPHP to downloads page and news

pronskiy:
As discussed previously with @thephpf, with full Windows support in makes sense to add FrankenPHP to downloads page as a n option and give it a short announcement in the news.

/cc @dunglas

### DIFF
--- a/include/download-instructions/frankenphp.php
+++ b/include/download-instructions/frankenphp.php
@@ -1,0 +1,38 @@
+<?php if ($options['os'] === 'linux'): ?>
+<p>
+On the command line, run the following commands:
+</p>
+<pre><code class="language-bash line-numbers">
+# Download and install FrankenPHP.
+curl https://frankenphp.dev/install.sh | sh
+
+# Serve the current directory.
+frankenphp php-server
+</code></pre>
+<?php elseif ($options['os'] === 'osx'): ?>
+<p>
+On the command line, run the following commands:
+</p>
+<pre><code class="language-bash line-numbers">
+# Install FrankenPHP using Homebrew.
+brew install dunglas/frankenphp/frankenphp
+
+# Serve the current directory.
+frankenphp php-server
+</code></pre>
+<?php elseif ($options['os'] === 'windows'): ?>
+<p>
+In PowerShell, run the following commands:
+</p>
+<pre><code class="language-powershell line-numbers">
+# Download and install FrankenPHP.
+irm https://frankenphp.dev/install.ps1 | iex
+
+# Serve the current directory.
+frankenphp php-server
+</code></pre>
+<?php endif; ?>
+
+<p>
+For more information, see the <a href="https://frankenphp.dev/docs/">FrankenPHP documentation</a>.
+</p>
--- a/include/download-instructions/frankenphp.php
+++ b/include/download-instructions/frankenphp.php
@@ -9,6 +9,17 @@ curl https://frankenphp.dev/install.sh | sh
 # Serve the current directory.
 frankenphp php-server
 </code></pre>
+<div class="note">
+<p>
+	The install script uses your distribution's packages (deb or RPM) when
+	available, and those receive updates through the system package manager.
+	If a standalone binary is installed instead, FrankenPHP does not update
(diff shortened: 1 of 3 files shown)


archive/entries/2026-03-09-1.xml
```diff
@@ -0,0 +1,29 @@
...
+            <p>
+                In 2025, <a href="https://frankenphp.dev/">FrankenPHP</a>, the modern PHP application server built on Caddy,
+                joined the official PHP organization on GitHub at
+                <a href="https://github.com/php/frankenphp">php/frankenphp</a>,
+                with support from <a href="https://thephp.foundation/blog/2025/05/15/frankenphp/">The PHP Foundation</a>.
+            </p>
+            <p>
+                The latest release brings
+                <a href="https://dunglas.dev/2026/03/windows-support-for-frankenphp-its-finally-alive/">Windows support</a>,
+                making FrankenPHP available on Linux, macOS, and Windows.
```
dunglas:
```suggestion
                making FrankenPHP available on Linux, macOS, FreeBSD, and Windows.
```

include/download-instructions/frankenphp.php
```diff
@@ -0,0 +1,38 @@
...
+On the command line, run the following commands:
+</p>
+<pre><code class="language-bash line-numbers">
+# Download and install FrankenPHP.
+curl https://frankenphp.dev/install.sh | sh
+
+# Serve the current directory.
+frankenphp php-server
+</code></pre>
+<?php elseif ($options['os'] === 'osx'): ?>
```
dunglas: You could also just run the same commands on a on Linux, if Homebrew is available it will be automatically used by the script.

dunglas (approved):

pronskiy:
Hey @derickr @saundefined @sgolemon @sy-records @bukka, friendly ping on this one. We discussed adding FrankenPHP to the downloads page at the PHP Foundation dev meeting and agreed to move forward with it. Would appreciate your reviews when you get a chance. Happy to address any feedback!

bukka:
The downloads make sense (someone needs to review the actual code though) but not sure about the archive and if it should go to front page as we put there only releases. If it's archive that it's not visible there (like conferences) then it's fine but for homepage we might need to make FrankenPHP more official and have some wider discussion. Currently it's like we would put there pecl extension release (which si the status that we gave FrankenPHP when moving to PHP org)...

sy-records (approved):

iluuu1994:
FrankenPHP does not auto-update, FWIU. I think it's important to note that the user will be responsible for doing so. This is especially true because dependencies are statically baked into the binary. For example, users will need to update FrankenPHP when there's an OpenSSL vulnerability, even if the system-provided OpenSSL receives the fix.

dunglas:
It depends on how users install it. The script will try to use the packages for the distribution (deb, RPM or homebrew) if possible. These versions auto-update.

iluuu1994:
@dunglas That's good to know, thanks! I'd still feel better if there was a note for the cases where a pre-compiled binary is used.

ramsey (approved):

ramsey:
> [force-pushed](https://github.com/php/web-php/compare/d18e8ce2c4399b6d53e38d60849cf89b8cae33e1..cf7c6d2b3ea48b157b8991a85fea010db77fb5ed) the master  branch from [d18e8ce](https://github.com/php/web-php/commit/d18e8ce2c4399b6d53e38d60849cf89b8cae33e1) to [cf7c6d2](https://github.com/php/web-php/commit/cf7c6d2b3ea48b157b8991a85fea010db77fb5ed)

👀 What?

pronskiy:
@ramsey, sorry, my bad, wrong git command 🙈

ramsey:
It might be easier if you create a topic/feature branch and do the PR from that, instead of doing the PR from your `master` branch.

pronskiy:
Folks, changes:
- Removed news entry for now as @bukka requested (will follow up after internals discussion)
- Added note regarding auto-update as @iluuu1994 suggested

PR #1873, opened 2026-03-24
Split preview CI

saundefined:
Split preview CI due security reasons.
Any suggestions and comments are welcome.

Refs:
- https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/


derickr:
I've heard back from the reporter as well, and they've given their blessing too.

PR #1876, opened 2026-03-31
Remove the 'I want to use PHP for' line from download instruction generation

derickr:
There was a discussion among the PHP Foundation developers about the usefulness of this line. Many of the framework-specific instructions linked back to this page to explain how to install PHP, defeating the purpose of linking to their specific instructions altogether.


derickr:
> That might mean we'll have to close #1856 as well.

That's fine. That wasn't suitable for merging any way.

PR #1890, opened 2026-04-21
Fix the ReferenceError when using keyboard navigation

esnard:
`curr` was undefined - I guess it was meant to be `current`, but it also looks like the whole instruction isn't needed anymore.


sy-records (approved):

sy-records:
Thank you

PR #1891, opened 2026-04-22
Adding PHPTek 2026

svpernova09:
Adding PHPTek 2026. Full disclosure: I work for PHP Architect, LLC.

Conference site: https://phptek.io


svpernova09:
PHPTek 2026 is already in the list, but it's very far below. Should the order be updated so the open CFP is shown first, then the upcoming by closest date?

PR #1902, opened 2026-05-13
Grammar edits, move paragraph, removed div in docs.php

pwright64:
Swapped order of the "translation disclaimer/contribution" paragraph and the list of languages to make page easier to navigate. Also, cleaned up language in the paragraph. Also, removed unnecessary div surrounding first two paragraphs. Styles are identical, could not determine any reason for them to be in their own div.

### DIFF
--- a/urlhowto.php
+++ b/urlhowto.php
@@ -32,17 +32,17 @@ function a($href): void {
 
 ?>
 
-<h1>Navigation tips&tricks</h1>
+<h1>Navigation tips & tricks</h1>
 
 <p>
- When using the PHP.net website, there is even no need to get to
+ When using the PHP.net website, there is no need to use
  a search box to access the content you would like to see quickly.
- You can use short PHP.net URLs to access pages directly.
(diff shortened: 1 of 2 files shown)


urlhowto.php
```diff
@@ -32,17 +32,17 @@ function a($href): void {
...
  a search box to access the content you would like to see quickly.
- You can use short PHP.net URLs to access pages directly.
+ You can instead use short PHP.net URLs to access pages directly.
 </p>
 
 <p>
- Note, that these shortcuts are expected to work on all mirror
- sites, not just at the main site. If you find that some of these
+ Note that these shortcuts are expected to work on all mirror
+ sites, not just on the main site. If you find that some of these
```
derickr: We don't have mirror sites any more, so this can be dropped.

derickr (changes requested):

urlhowto.php
pwright64: Paragraph removed.

PR #1907, opened 2026-05-28
Improve sidebar scroll behavior

Botato300:
This PR fixes a bug and adds a feature to the docs sidebar.

**Feature added:** When you visit a page such as https://www.php.net/manual/es/function.strlen.php, the side menu scrolls all the way to the top, which is awkward from a UX perspective. What I do is detect the active element (in this example, it would be strlen), and automatically scroll to that item. If I didn’t explain that clearly, here’s a video comparison:

https://github.com/user-attachments/assets/943a0e23-55e8-4938-9154-3866c7955ebe

**Bug fixed:** When using the J and K shortcuts to navigate through the items in the side menu, the scroll bar did not move. This PR fixes that!

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -104,7 +104,7 @@ Mousetrap.bind("b o r k", function () {
 function cycle(to, from) {
     from.removeClass("current");
     to.addClass("current");
-    $.scrollTo(to.offset().top);
+    to[0].scrollIntoView({ block: "center" });
 }
 
 function getNextOrPreviousSibling(node, forward) {
--- a/js/common.js
+++ b/js/common.js
@@ -145,7 +145,7 @@ function cycleHeaders(matches, forward) {
 
 function scrollToActiveItem() {
     const sidebar = document.querySelector(".layout-menu");
-    const activeItem = sidebar.querySelector(".current");
+    const activeItem = sidebar?.querySelector(".current");
 
     if (sidebar && activeItem) {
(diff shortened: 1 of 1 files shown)


sy-records:
Thanks for your PR.
When I visited the link, there was a noticeable scroll-down after the changes. Was that intentional?

preview link https://web-php-pr-1907.preview.thephp.foundation/manual/en/function.strlen.php

Botato300:
> Thanks for your PR. When I visited the link, there was a noticeable scroll-down after the changes. Was that intentional?
> 
> preview link https://web-php-pr-1907.preview.thephp.foundation/manual/en/function.strlen.php

Exactly, it's completely intentional. The idea is to make the item visible in the side menu (automatically when the page loads).

sy-records:
I don't think the top navigation bar should be affected.

Botato300:
> I don't think the top navigation bar should be affected.

Oh, you mean the sidebar in the window, right? That's caused by the `scrollIntoView()` function, which affects the parent scrollbars (whenever necessary). Would you prefer an implementation that only affects the sidebar menu's scrollbar? I can get started on that 😁

Botato300:
Done! Now the scroll doesn’t affect the window scroll or any other ancestor scroll.
@sy-records

sy-records (approved):
LGTM. Waiting for others

derickr (approved):

PR #1913, opened 2026-06-03
Future Website Concept

markrandall:
Prototype to demonstrate proposed changes to php.net as detailed in https://wiki.php.net/rfc/website_revamp_2026

### DIFF
--- a/ChangeLog-8.php
+++ b/ChangeLog-8.php
@@ -1030,6 +1030,47 @@ changelog_header(8, $MINOR_VERSIONS);
 
 <a id="PHP_8_4"></a>
 
+<section class="version" id="8.4.22"><!-- {{{ 8.4.22 -->
+<h3>Version 8.4.22</h3>
+<b><?php release_date('04-Jun-2026'); ?></b>
+<ul><li>Date:
+<ul>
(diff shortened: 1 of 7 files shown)


.docker/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+# We are using Node.js 22 LTS as a reliable, modern standard
+RUN apt-get update && apt-get install -y \
+    curl gnupg \
+    && curl -fsSL https://deb.nodesource.com/setup_22.x | bash - \
+    && apt-get install -y nodejs \
+    && apt-get clean \
+    && rm -rf /var/lib/lists/*
+
+# 3. Globally install TypeScript and the Socket Dev CLI
+RUN npm install -g typescript socket tsx esbuild
```
derickr: I don't see this being used anywhere. Do we need to add this Docker file?

markrandall: That's a leftover from hacking together a local dev environment - I strongly prefer to run all such projects (and their toolchains) inside containers as a standard security precaution.

I'll either polish it up for broader use, or remove it, prior to marking the PR as ready.

PR #1914, opened 2026-06-08
Avoid wrapping PHP 8.5 release code blocks

theodorejb:
This improves readability (particularly on smaller screens), since code blocks with overflow can be easily scrolled horizontally to read longer lines, rather than trying to visually parse syntax with line breaks in unintended places (which can even appear invalid when a line break is in the middle of a token).

Before:
before 2

After:
after 2

### DIFF
--- a/releases/8.5/release.inc
+++ b/releases/8.5/release.inc
@@ -686,7 +686,7 @@ PHP
         <div class="release-notes-grid-container">
             <div class="release-notes-grid">
                 <div>
-                    <h2><?= message('new_classes_title', $lang) ?></h2>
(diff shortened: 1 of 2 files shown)


saundefined:
Preview: https://web-php-pr-1914.preview.thephp.foundation/releases/8.5/en.php

theodorejb:
I also added a commit to allow linking to the "Additional features and improvements" and "Deprecations and backward compatibility breaks" headings. This works on the other release pages, but for some reason wasn't implemented on the PHP 8.5 release page.

Issue #1917, opened 2026-06-10
Documentation pages requesting assets from new.www.internal.php.net and timing out


kriswillis:
### Affected Page URL

https://www.php.net/docs.php

### Describe the bug

The documentation pages are failing to load due to requesting assets from the new.www.internal.php.net domain.

### Steps to reproduce

1. Visit https://www.php.net/docs.php

### Expected behavior

Page and all assets load

derickr:
Hi,

This was a problem for a little while (the `<base>` HTML element had the wrong value). But I can not reproduce this now, and this should no longer be occurring.

cheers,
Derick

kriswillis:
Working again here, too, thanks!

PR #1918, opened 2026-06-11
Fix broken relative links on manual pages

### LINKED ISSUE #395: The language setting (Japanese) is not maintained when clicking links in the manual
From manual page: https://php.net/class.ziparchive

---

### Description
When browsing the PHP manual in Japanese, clicking any link within the page redirects to the English version (`/manual/en/...`) instead of staying on the Japanese version (`/manual/ja/...`). 

Even though the Japanese translation for the linked page exists and can be viewed by manually switching the language back to Japanese, the language preference is not being maintained across page transitions.

### Steps to Reproduce
1. Open any PHP manual page in Japanese (e.g., `https://www.php.net/manual/ja/class.ziparchive.php`).
2. Click any link to a method or function within that page.
3. The browser navigates to the English version of the page (`/manual/en/...`).

### Expected Behavior
The manual should maintain the selected language (Japanese) and navigate to the `/manual/ja/...` URL when a link is clicked.

KentarouTakeda:
f8821252 dropped the page path from the `<base>` href, so relative links on
manual pages now resolve against the site root: in-page links on localized
pages land on the English version. Fixes php/doc-ja#395.

image

This restores the path part only, mirroring the canonical link below it; the
host stays hardcoded, keeping the #1917 fix. On production the output is
identical to pre f8821252.

Manual pages appear to be served through a CDN, so the fix may not be
visible everywhere until the cache is purged.

### DIFF
--- a/include/header.inc
+++ b/include/header.inc
@@ -61,7 +61,7 @@ if (!isset($config["languages"])) {
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
-  <base href="https://www.php.net">
+  <base href="https://www.php.net/<?php echo $_SERVER['BASE_PAGE'] ?? '' ?>">
 
   <title><?php echo $title ?></title>
 


sy-records:
🚀 Preview for commit 21e90449bae24424567423ca727da2291a856fc6 is available at https://web-php-pr-1918.preview.thephp.foundation/

include/header.inc
```diff
@@ -61,7 +61,7 @@ if (!isset($config["languages"])) {
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
 
-  <base href="https://www.php.net">
+  <base href="https://www.php.net/<?php echo $_SERVER['BASE_PAGE'] ?? '' ?>">
```
sy-records, line 64: LGTM. But I think the old way was better. Now, when I click on it during the preview test, it redirects to php.net.

```suggestion
  <?php if (!empty($_SERVER["BASE_HREF"])): ?>
   <base href="<?php echo $_SERVER["BASE_HREF"]; ?>">
  <?php endif ?>
```

sy-records (approved):

include/header.inc
KentarouTakeda: Thanks for the review! That's a side effect of pinning the host, and it's intentional: the suggested code derives the href from `SERVER_NAME` — that's how `new.www.internal.php.net` ended up in the `<base>` and broke asset loading (#1917). This PR restores only the path part, using the same pattern as the canonical link below.

derickr: @KentarouTakeda Not *quite* — I hadn't set the original host to be fed through to the CDN. The commit that I made yesterday can be reverted. I will now do that (and hence, I can close this PR too).

derickr:
Solved through 742ae5c2269d7c6df25e2716b5d143970e4962e5

PR #1916, opened 2026-06-10
Unify design of all release pages

theodorejb:
Updates the design of the PHP 8.0-8.4 release pages to match PHP 8.5, with dark mode support, syntax highlighting of variables in code blocks, buttons to copy code blocks, and other design improvements. Release pages prior to 8.5 still lack the Key Features section.

Also moved to functions for rendering the hero section, feature comparisons, and footer, to avoid duplicating huge amounts of markup and simplify maintaining consistent styling.

No content has been changed, except for correcting an English grammar typo on the PHP 8.1 page, fixing inconsistent i18n language spelling in a few dropdowns, and correctly styling code in PHP 8.5 feature headings (as was already done on previous release pages).

Builds on top of #1914, so that should be merged first.

PR #1920, opened 2026-06-12
Update download link to point to downloads.php

vincent-bru:


sy-records (approved):

sy-records:
Normally, when I visit /downloads, I should be redirected to /downloads.php. It works fine when I test it locally, but for some reason it doesn't work on php.net.

https://github.com/php/web-php/blob/03ab57cbb2bf44b284ba6adc4c4b09dc8ae09a79/error.php#L188-L194

Could you take a look at this? @derickr

vincent-bru:
> Normally, when I visit /downloads, I should be redirected to /downloads.php. It works fine when I test it locally, but for some reason it doesn't work on php.net.
> 
> Could you take a look at this? @derickr

It means it is a server config thing. (htaccess not hitting the right spot?)

derickr:
I can't reproduce this, https://www.php.net/downloads works for me.

vincent-bru:
> I can't reproduce this, https://www.php.net/downloads works for me.

go on php.net, click on the php8.5 logo from the menu
image
now click on the "Upgrade to php8.5" button in the banner
image

derickr:
@vincent-bru I did that. Works for me.

vincent-bru:
> @vincent-bru I did that. Works for me.

Ok, on firefox, edge, chrome on my pc and laptop it doesn't work but when I go private it works (no extensions in both test)

sy-records:
The page loads normally, but shouldn't it redirect to downloads.php?

derickr:
No, we didn't tend to redirect these. I wonder what redirects it in your local setup then @sy-records . Can you trace that?

sy-records:
https://github.com/php/web-php/blob/03ab57cbb2bf44b284ba6adc4c4b09dc8ae09a79/error.php#L188-L194

It's redirected here, and I think the redirect is correct; 

most of the references in the repository point to /downloads.php. https://github.com/search?q=repo%3Aphp%2Fweb-php+%2Fdownloads.php&type=code

Meanwhile, /downloads was introduced in #1916. This should be standardized to use /downloads.php

derickr:
Hmm, this is odd. You're right, I missed the `!` in those few lines. I first thought this was something on the CDN (I saw a setting about it), but it is `off` meaning:

```Follow redirects
If enabled, bunny.net will follow the redirects returned by the origin.
If disabled, the redirect responses will be cached on the CDN and returned to the end-user. 
```

derickr:
Ah, it should be redirected somewhere else already: https://github.com/php/infrastructure/blob/main/roles/properties/www/templates/www.php.net.conf#L36 — I'll have a look why that doesn't trigger.

Please do fix that link to `/downloads` in a PR (the one introduced in #1916).

derickr:
Fixed the configuration too: https://github.com/php/infrastructure/commit/a700be0ae4a11a6d1326ce4703d441c9f9168f2d

PR #892, opened 2023-12-08
Enhancement: Run `vimeo/psalm` on GitHub Actions

localheinz:
This pull request

- [x] runs a static code analysis with `vimeo/psalm` on GitHub Actions

Follows #559.

PR #1922, opened 2026-06-13
Setup PHPStan

staabm:
adding a minimal PHPStan setup. this is the starting point so iterating on the PHPStan level can be achieved in future PRs.

as requested in https://github.com/php/web-php/pull/892#issuecomment-4693509965

### DIFF
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -115,7 +115,7 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "8.2"
+          - "8.4"
 
         dependencies:
           - "locked"
--- a/.github/workflows/integrate.yaml
+++ b/.github/workflows/integrate.yaml
@@ -107,6 +107,49 @@ jobs:
       - name: "Validate XML files"
         run: "for a in $(find . -name '*.xml'); do xmllint --quiet --noout $a; done"
 
+  static-analysis:
+    name: "Static Analysis"
+
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        php-version:
+          - "8.2"
+
+        dependencies:
+          - "locked"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v3"
+
+      - name: "Set up PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          extensions: "none, curl, dom, json, mbstring, tokenizer, xml, xmlwriter, iconv"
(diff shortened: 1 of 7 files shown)


phpstan-baseline.neon
```diff
@@ -0,0 +1,295 @@
...
+			message: '#^Variable \$version might not be defined\.$#'
+			identifier: variable.undefined
+			count: 1
+			path: include/download-instructions/linux-debian-cli-community.php
+
+		-
+			message: '#^Variable \$version might not be defined\.$#'
+			identifier: variable.undefined
+			count: 1
+			path: include/download-instructions/linux-debian-web-community.php
```
staabm, line 235: one way to get rid of these errors in the future would be declaring variables expected to be passed in from global scope via

```php
<?php
/** @var string $version */
?>
```

at the top of the files.

---

alternatively a more aggressive approach could be asserting the variables instead of just making the tooling aware of it using e.g.
```php
<?php
assert(isset($version) && is_string($version));
?>
```

phpstan.neon.dist
```diff
@@ -0,0 +1,14 @@
...
+	level: 3
+	paths:
+		- .
+	fileExtensions:
+		- php
+		- inc
+
+	excludePaths:
+		analyse:
+			- vendor/
```
staabm, line 14: after getting more used to the project structure I think for now its best to scan all php files, but exclude only the `vendor/`. that way we don't miss anything.

initially I configured to only scan certain folders, but this would e.g. miss .php files in the repo root.

this may be fine tuned in future PRs of course

markrandall:
I'd be glad to have phpstan in, although FWIW as I'm rewriting pretty much everything I am aiming for level 8.

staabm:
having it more strict is great to have. but for a start and for all these legacy scripts we need to start somewhere :)

.github/workflows/integrate.yaml
```diff
@@ -107,6 +107,49 @@ jobs:
...
 
+  static-analysis:
+    name: "Static Analysis"
+
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        php-version:
+          - "8.2"
```
derickr: Should be `8.4` now.

derickr (commented):
I don't mind, but I couldn't see from the PR whether this would start failing things when phpstan finds problems in PR to this repo.

derickr:
My fault, but after this (https://github.com/php/web-php/pull/1929) there are now conflicts in `composer.{json,lock}`.

staabm:
conflict resovled

PR #1923, opened 2026-06-14
Remove duplicate definitions of `base`

sy-records:


derickr (changes requested):
I would probably leave the top one, and not the bottom one. 

I also don't think that whitespace changes are mixed in within the same *commit* — could you split that up?

PR #1929, opened 2026-06-16
Production now runs PHP 8.4

derickr:


sy-records (approved):

pronskiy:
How about updating composer.json entries as well?

saundefined (approved):

PR #1931, opened 2026-06-17
[Preview - Part 1] Eliminate globals for $RELASES / $OLDRELEASES / $BRANCHES

markrandall:
Attempts to eliminate the dependence on the $RELEASES, $OLDRELEASES, and $BRANCHES global variables that require injecting in at the top level scope to avoid breaking things.

Changes:
- releases.inc and versions.inc now return their values, as well as retaining the previous global-scope, allowing the new helper to reliably access their contents by requiring them again, and using the return value.
- Updated bumpRelease to retain returning logic.
- Added extra normalization to the data returned from the two data sources.
- Updated everywhere I could find outside of branches.inc that used those old helpers.
- Wrote unit tests for the new helpers.

Included:
- Removes requirements to annotate unit tests with code coverage


include/branches.inc
```diff
@@ -135,7 +108,8 @@ function get_all_branches() {
     return $branches;
 }
 
-function get_active_branches($include_recent_eols = true) {
+#[DEprecated('Use Branches::active()')]
```
DanielEScherzer:
```suggestion
#[Deprecated('Use Branches::active()')]
```

src/Releases/Branches.php
```diff
@@ -0,0 +1,462 @@
...
+
+        return null;
+    }
+
+    /**
+     * Finds the last release from a given branch
+     *
+     * @return NormalizedReleaseStruct|null
+     */
+
```
DanielEScherzer:
```suggestion
```

src/Releases/Branches.php
```diff
@@ -0,0 +1,434 @@
...
+{
+    /**
+     * @return array<int, array<string, NormalizedReleaseStruct>>
+     */
+    public static function getReleaseData(): array
+    {
+        static $cache = null;
+
+        /* there is no normalisation required here because it's all standard format */
+        return $cache ??= require __DIR__ . '/../../include/version.inc';
```
derickr, line 43: I wonder whether this "caching" is needed, considering the return happens straight from an opcache-cached oparray.

src/Releases/Branches.php
```diff
@@ -0,0 +1,434 @@
...
+    }
+
+    /**
+     * @return array<int, array<string, NormalizedReleaseStruct>>
+     */
+    public static function all(): array
+    {
+        static $cache = null;
+        return $cache ??= (function () {
+            $results = [];
```
derickr, line 99: same here, and the use of a closure seems odd to just wrap around a single return value, without it being reused.

src/Releases/Branches.php
```diff
@@ -0,0 +1,434 @@
...
+        })();
+    }
+
+    /**
+     * @return array<string, array{security?:string, stable?:string, date?:string}>
+     */
+    public static function getBranchOverrides(): array
+    {
+        static $cache = null;
+        return $cache ??= require __DIR__ . '/../../include/branch-overrides.inc';
```
derickr, line 89: same

src/Releases/Branches.php
markrandall: I find the style personally cleaner than a big if block or double return points.

src/Releases/Branches.php
markrandall: version.inc is the one for the most recent releases - as such it's optimized for updating hashes, and and uses an IIFE each time it is included to format the data into the normalized form.

include/branches.inc
```diff
@@ -274,6 +250,7 @@ function get_initial_release($branch) {
     return null;
 }
 
+#[Deprecated('Use Branches::getFinalRelease')]
 function get_final_release($branch) {
```
derickr, line 254: I think this is only [used in this file](https://github.com/php/web-php/pull/1931/changes#diff-43d06e91255c23eaf41c1487da11fc6d976173f61782484e5ad3ecf8f73005deR315), so perhaps remove it?

src/Releases/Branches.php
derickr: No, I specifically meant the `static $cache` part. I really don't think that's needed (nor wanted).

derickr (changes requested):

src/Releases/Branches.php
markrandall: Is your position that every time we call the function to get the release data, we should re-include the file, and reinvoke the IIFE that's already in there? Because the static is what prevents that from happening.

derickr:
Been thinking about this, and I am really not keen on including random files during static function calls, nor IIFE's, nor using `static` variables in static methods. Would it not be a better design to have an actual Branches *class* instead of a glorified namespace for static functions? That way you can in the constructor include the `.inc` file(s), and if needed, load certain properties lazily? If necessary, I think I would even prefer a singleton over this.

markrandall:
> Would it not be a better design to have an actual Branches class instead of a glorified namespace for static functions?

Yes it would, in fact that's what the domain objects based version did in the original build. This was to a) facilitate the front page information without having to rely on fragile global variables b) Provide a pathway for auto loading them without having to do the explicit global-scope include and c) the domain objects version is much more complex.

As for statics and IIFEs they're fully functional and high performance. I would prefer us not to get bogged down in coding styles.

derickr:
I have merge this now, but I would like in further PRs to see fewer of these "static member hacks".

PR #1932, opened 2026-06-18
Add phpstan type to NewsHandler

markrandall:
Defines shape of NewsHandler entry.

### DIFF
--- a/include/layout.inc
+++ b/include/layout.inc
@@ -542,7 +542,7 @@ function get_news_changes()
         return false;
     }
 
-    $date = date_create($lastNews["updated"]);
+    $date = date_create($lastNews["updated"] ?? $lastNews["published"]);
     if (isset($_COOKIE["LAST_NEWS"]) && $_COOKIE["LAST_NEWS"] >= $date->getTimestamp()) {
         return false;
     }
@@ -557,6 +557,8 @@ function get_news_changes()
 
     $date->modify("+1 week");
     if ($date->getTimestamp() > $_SERVER["REQUEST_TIME"]) {
+        assert(isset($lastNews["link"][0]["href"]));
+
         $link = preg_replace('~^(http://php.net/|https://www.php.net/)~', '/', $lastNews["link"][0]["href"]);
         $title = $lastNews["title"];
         return "<a href='{$link}'>{$title}</a>";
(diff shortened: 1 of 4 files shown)


src/News/NewsHandler.php
```diff
@@ -24,8 +50,14 @@ public function getLastestNews(): array|null
         return $news[0];
     }
 
-    /** @return list<array> */
-    public function getFrontPageNews(): array
+    /**
+     * Unused - Looks up generated news with frontpage tags.
+     *
+     * Currently unused as the front page no longer displays news.
```
derickr: That's not really true yet, right? That needs a whole lot more works than just this PR.

src/News/NewsHandler.php
```diff
@@ -71,11 +107,46 @@ public function getNewsByYear(int $year): array
...
 
+        /** @phpstan-ignore-next-line - pregen-news sets global variable */
         return is_array($NEWS_ENTRIES) ? $NEWS_ENTRIES : [];
     }
+
+
+    /**
+     * @param NewsEntryStruct $data
+     */
+    public static function parseTeaserCutoff(array $data): ?DateTimeImmutable
```
derickr: These two seem to be new functions? I can't see where they're used.

markrandall:
It turns out that `$entry['updated']` isn't universal, so I had to touch a few extra controllers to make them fall back to  `$entry['published']`

src/News/NewsHandler.php
markrandall: One removed (was used in later branch), the other applied.

include/layout.inc
```diff
@@ -557,6 +557,8 @@ function get_news_changes()
 
     $date->modify("+1 week");
     if ($date->getTimestamp() > $_SERVER["REQUEST_TIME"]) {
+        assert(isset($lastNews["link"][0]["href"]));
```
derickr: I don't think we should have assertions in user-facing code (ie, code that might end up rendering a page for a visitor to see). Because it will result in either a (cached) whitescreen page, or a mangled one.

derickr (changes requested):

include/layout.inc
markrandall: preg_match already expects a string so if the value isn't set, at the very least it would give a deprecation warning on the next line.

Any uncaught exceptions really should be returning a 500 error and the CDN not caching it. Am I to understand that is not already the case? If so it could be handled once we move to a front controller.

derickr: Exceptions didn't exist when this code was written, so there is no such thing as "uncaught exceptions" built-in. So nothing is handled for it either.

PR #1934, opened 2026-06-19
[Preview - Part 2] Basic Framework

markrandall:
Builds upon Part 1 - https://github.com/php/web-php/pull/1931

These core framework components all work together. While they could _technically_ be split up into individual PR I have chosen not to do so due to their tight dependence on each other to deliver real-world functionality. 

As we're only writing glue code for industry-standard libraries, rather than building fully custom implementations of routing and services, being able to see them all working together should provide better context for reviewing than if they were all self-contained with no meaningful bigger-picture.

### Build Process
Facilitates creating optimized data for the sake of enormously increased performance vs runtime reflection and data processing. 

The "build" logic itself is a a simple executor that auto-detects classes that implement the `BuildStep` interface, and invokes their static `build` method, passing in a basic context & logging object. The meat of the build process is contained in those function-specific classes that implement `BuildStep`.

To invoke build, run `php bin/build.php` and the compiled data will be generated in `var/`, to which reading and writing is encapsulated via `VarCache`.

**Coupling:** Used by both router and services to provide essential performance improvements vs runtime scanning and reflection.
**External Dependencies:** None 

### Routing
Decouples the physical path of the .php file from its controller code, allows using placeholders / wildcards within routing URLs instead of needing to create large numbers of template-generated files (as is the case with release announcements) e.g. `releases/8_4_1.php`.

Uses standard attribute-based `#[Route]`ing mechanism to detect classes, which are also used by the service layer to register anything with one as a service.

**Coupling:** Depends on build step to compile routing paths without incurring significant performance hit. Used by Kernel for core functionality.
**External Dependencies:** Symfony Router

### Service Layer
Allows registering classes as services, allows inversion of control (IoC) by autowiring service constructor arguments with other specified services.

**Coupling:** Depends on build step to compile service definitions, without incurring significant performance hit. Depends on routing layer to bootstrap container and create initial controllers. Used by Kernel for core functionality.
**External Dependencies:** Symfony Dependency Injection / Container Interface

### HTTP Kernel
Processes a HTTP request and converts it to a Response, and sends that response to the end-user. Currently because this is a workaround until we can get a fully supported front controller this kernel only gets invoked in the error.php handler, and if it does not match a route it continues on to let the other hundreds of rules in error.php take their  turn.

**Coupling:** Depends on Router, Service Layer
**External Dependencies:** None


src/Framework/IO/FS.php
```diff
@@ -0,0 +1,52 @@
...
+ * If any of these throw, then something has likely gone very wrong.
+ */
+class FS
+{
+    /**
+     * @throws RuntimeException
+     */
+    public static function readContents(string $path): string
+    {
+        $contents = file_get_contents($path);
```
jordikroon, line 21: This may throw an E_WARNING. We can use [thecodingmachine/safe](https://github.com/thecodingmachine/safe) to avoid that behaviour. Only checking for false is not sufficient.

src/Framework/Build/BuildTools.php
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     * Used for protecting developers from hurting themselves by relying on data which
+     * might have stale caches or otherwise be unready for use during build steps
+     */
+    public static bool $building = false;
+
+    /**
+     * @return array<string, ReflectionClass<object>>
+     */
+    public static function getReflectionClasses(): array
```
jordikroon, line 54: I would prefer that we define `$cache` as property. I am only really in favor of `static $var` within functions. It's used at some more places.

src/Framework/Build/BuildTools.php
```diff
@@ -0,0 +1,224 @@
...
+        foreach (BuildTools::parseClassHierarchy(ProjectGlobals::getProjectRoot() . '/src') as $class) {
+            if (!$class->isSubclassOf(BuildStep::class)) {
+                continue;
+            }
+
+            $buildSteps[] = $class->getName();
+        }
+
+        /* protect us against doing silly things */
+        BuildTools::$building = true;
```
jordikroon: Probably better to use static in case we ever change the class name.

src/Framework/Build/BuildTools.php
```diff
@@ -0,0 +1,224 @@
...
+            }
+        }
+
+        return $foundFiles;
+    }
+
+    public static function getLastModifiedTime(): int
+    {
+        /* in development mode we always assume that we've recently refreshed our data */
+        if (isset($_ENV['DEVELOPMENT'])) {
```
jordikroon: This will fail if `DEVELOPMENT=0`. It's better to define a global constant `const DEVELOPMENT = true` during initialisation.

As for the name. I prefer `IS_DEBUG`. But that's a minor nit.

src/Framework/Services/ServiceLocator.php
```diff
@@ -0,0 +1,131 @@
...
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Dumper\PhpDumper;
+use Symfony\Component\DependencyInjection\Reference;
+use function is_file;
+
+class ServiceLocator implements BuildStep
+{
+    public static function build(BuildContext $context): void
+    {
+        ini_set('memory_limit', '1G');
```
jordikroon: Is this really necessary? It's a big chunk of memory.

src/Framework/functions.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+/* no-class */
+
+/**
+ * Default escaping mechanism for unsafe HTML
+ */
+function safe(string $html): string
+{
+    return htmlspecialchars($html, encoding: 'UTF-8');
```
jordikroon, line 10: The function name `safe` feels very vague. Can we make this more explicit, like `safe_htmlspecialchars`.
Though this doesn't seem to be used now while I do see `htmlspecialchars` being used all over.

jordikroon (changes requested):

src/Framework/IO/FS.php
markrandall: It looks like checking against false is what safe does https://github.com/thecodingmachine/safe/blob/4fbc0088994d486b0012d67116b90825f17a9309/generated/8.1/filesystem.php#L165

markrandall: That said, we could potentially bring in League's Flysystem for this, that would give us a consistent API across any future filesystems we need to use.

src/Framework/Services/ServiceLocator.php
markrandall: What's the downside? Although that should be in the build initializer.

src/Framework/functions.php
markrandall: It is used extensively throughout other parts of the code that I had to pull out to make this PR. It's safe to output its contents straight into a template due to HTML encoding. As it would be used hundreds / thousands of times, there's no benefit IMO to making it more verbose than it needs to be.

src/Framework/Build/BuildTools.php
markrandall: There is no single point of initialization that applies to all entry points so far as I can tell.

Even if there were, there's no existing mechanism I have yet found for file-based site-level configuration to be used to set it, so we would have to either create one, or use ENV anyway. 

Moving the debug state to project globals with null coalesce would be a good change.

src/Framework/Services/ServiceLocator.php
jordikroon: The downside is that it implies 1G is needed. That's why I am wondering if this really needs 1G, or if a lower value is sufficient.

markrandall: As a build script, my feeling is it should be able to take as much memory as it needs, and spending time trying to fine tune it wouldn't be a good use of time. 1GB should be plenty for now and in future, if it goes beyond that then chances are something's gone majorly wrong.

I did originally hit the 128MB limit when I was building the full project.

src/ProjectGlobals.php
```diff
@@ -23,4 +23,9 @@ public static function getPublicRoot(): string
         return realpath(__DIR__ . '/../public')
             ?: throw new Error('Unable to locate public root');
     }
+
+    public static function isDevelopmentMode(): bool
+    {
+        return ($_ENV['DEVELOPMENT'] ?? '') === '1';
```
jordikroon: To also allow `true`, `on` or `yes`.

```suggestion
        return filter_var(($_ENV['DEVELOPMENT'] ?? ''), FILTER_VALIDATE_BOOL);
```

markrandall: Good idea.

PR #1942, opened 2026-06-24
Fix preserving query strings for error page redirects

shivammathur:


error.php
```diff
@@ -19,7 +19,8 @@
 // Get URI for this request, strip leading slash
 // See langchooser.inc for more info on STRIPPED_URI
 $URI = substr($_SERVER['STRIPPED_URI'], 1);
-$queryString = $_SERVER['QUERY_STRING'] ?? '';
+$uriParts = explode('?', $URI, 2);
+$queryString = isset($uriParts[1]) ? html_entity_decode($uriParts[1], ENT_QUOTES, 'UTF-8') : '';
```
derickr, line 24: This doesn't yet account for the fact that `$URI` is used as index into the `$url_aliases` array. If a query string for any of these entries is present, it would still not match.

I think `$URI` needs to be changed too, to `$uriParts[0]` after the explosion in 22.

derickr (changes requested):

PR #1957, opened 2026-07-06
Add a copy function for the SHA256 checksum

sy-records:
Close #1956

### DIFF
--- a/include/version.inc
+++ b/include/version.inc
@@ -139,8 +139,9 @@ function show_source_releases()
                   <?php download_link($rel['filename'], $rel['filename']); ?>
                   <span class="releasedate"><?php echo date('d M Y', strtotime($rel['date'])); ?></span>
                   <?php
-                    if (isset($rel['md5']))    echo '<span class="md5sum">', $rel['md5'], '</span>';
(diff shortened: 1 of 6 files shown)


derickr (changes requested):
This change is good, but it's not complete. There is a whole list of versions and SHA256 codes on `/releases/` as well.

include/version.inc
```diff
@@ -139,8 +139,12 @@ function show_source_releases()
                   <?php download_link($rel['filename'], $rel['filename']); ?>
                   <span class="releasedate"><?php echo date('d M Y', strtotime($rel['date'])); ?></span>
                   <?php
-                    if (isset($rel['md5']))    echo '<span class="md5sum">', $rel['md5'], '</span>';
-                    if (isset($rel['sha256'])) echo '<span class="sha256">', $rel['sha256'], '</span>';
+                    if (isset($rel['md5'])) {
```
sy-records: Should we remove the MD5 section?

derickr: We can do that, as I had a look, and can't actually find anywhere where this is still displayed.

PR #1960, opened 2026-07-06
migrate doc redirects to redirects.json sidecar from doc-en

jordikroon:
The goal for this PR is to migrate all doc related redirects out of web-php.
Instead we generate a `redirects.json` from `redirects.xml` which is then loaded in-place.

Redirects not related to docs are NOT migrated. However shortcuts are.

Depends on:

https://github.com/php/phd/pull/260
https://github.com/php/doc-en/pull/5658


sy-records:
Do other languages also need to generate a redirects.xml file?

error.php
```diff
@@ -198,155 +197,60 @@
...
-    } else {
-        mirror_redirect("/manual/{$match[1]}/{$parts[0]}.php#{$parts[1]}");
+// Doc-page redirect map generated by phd from doc-en/redirects.xml,
+// shipped as manual/en/redirects.json.
+//   'shortcut'  - bare URI  (php.net/{from})               -> /manual/{LANG}/{to}.php
+//   'in_manual' - manual URI (/manual/{lang}/{from}[.php]) -> /manual/{lang}/{to}.php
+//   'both'      - matches at either entry point
+
+$docRedirects = is_file(__DIR__ . '/manual/en/redirects.json')
+    ? json_decode(file_get_contents(__DIR__ . '/manual/en/redirects.json'), true)
```
derickr, line 208: `json_decode()` can also fail of the JSON file is corrupted for some reason. Please guard against that.

jordikroon:
No. While technically they can, we should only load doc-en. The file is explicitly marked as `<?do-not-translate?>` in doc-en and should not have a copy in doc-lang repos.

Though it will be very likely that we will add a variant of redirects.xml to doc-ext (third-party extension manual) in the future.

Having the redirects.xml in doc-en (not doc-base) allows us to manage redirects from one PR, rather than having PRs to depend on each other. It also allows a variant in doc-ext in later on.

derickr (changes requested):

Issue #1955, opened 2026-07-05
www.php.net homepage is blank


utf-4096:
### Affected Page URL

https://www.php.net

### Describe the bug

For some reason, my combination of browser settings makes https://www.php.net return a blank page every time, but other URLs and even https://www.php.net/?1 or https://www.php.net/index.php work, possibly indicating a bad cache?

### Steps to reproduce

1. Run `curl --verbose --header 'Accept-Language: en-US,en;q=0.9' 'https://www.php.net/'`
2. See that a blank page is returned

### Expected behavior

Homepage shows up, like on https://www.php.net/index.php.

### Screenshots

<img width="952" height="720" alt="Image" src="https://github.com/user-attachments/assets/502e6ad9-f474-40a6-872c-b432de327852" />

Krinkle:
Same here, from London.

<img width="2063" height="1781" alt="Image" src="https://github.com/user-attachments/assets/29a99382-7ac3-476e-855d-2f7483f16244" />

risharde:
Wow, it's still down 2 days later? That I did not expect!

R0bur:
I encoutered this issue in the Chrome browser. In my case, these URLs return empty response (HTTP 200, content-length: 0):
https://www.php.net/
https://www.php.net/downloads.php
I discovered that this problem is caused by the "accept-language" HTTP request header. The issue can be resolved by moving the "English (US)" option to the top of the preferred languages ​​list (chrome://settings/languages).

AllenJB:
Ping @derickr 

Based on a conversation today in the #php-webmaster channel on the PHPC Discord today I believe this issue may now be resolved.

svpernova09:
I did a full CDN Purge (after a URL purge of `https://www.php.net` did not clear the white page) for www.php.net, and it was reported working again. I wasn't able to reproduce the issue from the central US (Comcast).

utf-4096:
Homepage still blank on my end.

Full request/response, if it helps:

<details>

```
$ curl --verbose --header 'Accept-Language: en-US,en;q=0.9' 'https://www.php.net/'
* Host www.php.net:443 was resolved.
* IPv6: 2400:52e0:1e02::1273:1
* IPv4: 143.244.56.49
*   Trying [2400:52e0:1e02::1273:1]:443...
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* SSL Trust Anchors:
[...]

PR #1971, opened 2026-07-15
Fix blank pages from cached CORS rejection

philip:
Addresses issue #1955

Requests with an unrecognized Origin header hit exit(10), which returned an empty 200 with no Cache-Control. The CDN cached that blank page and served it from the affected nodes for 30 days (required manual purge to fix), even to requests sending no Origin at all. Now we skip the CORS headers and serve the page normally.


philip:
Issue was addressed via https://github.com/php/web-php/pull/1975

PR #1975, opened 2026-07-24
Do not abort cross-origin requests with an empty 200 (#1955)

svpernova09:
How to reproduce

```shell
$ tests/server start
$ git checkout master

$ curl -s -o /dev/null -w '%{http_code} %{size_download}\n' \
-H 'Origin: https://example.com' http://localhost:8080/
200 0

$ curl -s -o /dev/null -w '%{http_code} %{size_download}\n' \
http://localhost:8080/
200 53385
```

Same URL, same server. One request-header a client fully controls turns the page into a zero-byte 200 OK.

`exit(10)` at `include/prepend.inc:31` sets a process exit status and emits no body, so the response is a perfectly 
valid success with nothing in it.

That empty body then gets cached and served to everyone: Origin is not part of the CDN cache key and we send no Vary 
header. It also persists, because Cache-Control: public, max-age=300 is set later in include/header.inc:54 and never 
runs, so the CDN applies its own default TTL instead of five minutes.

On this branch:

```shell
$ git checkout prevent-returning-blank-responses

$ curl -s -o /dev/null -w '%{http_code} %{size_download}\n' \
-H 'Origin: https://example.com' http://localhost:8080/
200 53385
```

### DIFF
--- a/include/prepend.inc
+++ b/include/prepend.inc
@@ -21,20 +21,39 @@ header("Permissions-Policy: interest-cohort=()");
 
 /* Fix Silly Same Origin Policies */
 (function (): void {
+    // The presence of CORS headers below depends on the Origin request header,
+    // so shared caches must not reuse a response across different origins.
+    header("Vary: Origin", false);
+
     if (!isset($_SERVER["HTTP_ORIGIN"])) {
         return;
     }
 
+    // Browsers send a literal "null" origin for sandboxed iframes, documents
+    // loaded from data:/file: URLs and some cross origin redirects. parse_url()
+    // reports no host at all for those, so treat them as untrusted.
     $host = parse_url($_SERVER["HTTP_ORIGIN"]);
-    if (!preg_match('/^(.+\.)?php\.net$/', $host["host"])) {
-        if ($host["host"] != $_SERVER["SERVER_NAME"]) {
-            exit(10);
+    $hostname = $host["host"] ?? "";
+
+    $isTrustedOrigin = $hostname !== "" && (
+        preg_match('/^(.+\.)?php\.net$/', $hostname) === 1
+        || $hostname === ($_SERVER["SERVER_NAME"] ?? "")
+    );
+
+    if (!$isTrustedOrigin) {
+        // Serve the page normally, only without any CORS headers.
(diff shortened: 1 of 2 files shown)


include/prepend.inc
```diff
@@ -21,20 +21,39 @@ header("Permissions-Policy: interest-cohort=()");
...
+
+    $isTrustedOrigin = $hostname !== "" && (
+        preg_match('/^(.+\.)?php\.net$/', $hostname) === 1
+        || $hostname === ($_SERVER["SERVER_NAME"] ?? "")
+    );
+
+    if (!$isTrustedOrigin) {
+        // Serve the page normally, only without any CORS headers.
+        if (in_array($_SERVER["REQUEST_METHOD"] ?? "GET", ["GET", "HEAD", "OPTIONS"], true)) {
+            return;
```
derickr, line 46: I don't understand why you're allowing these three methods without the CORS origin server to match.

svpernova09: It's not an access gate — it only decides whether the response carries CORS headers. Safe methods are served as usual, because these pages are public anyway: drop the Origin header and you get the same bytes, on master too. What stops a foreign page reading the response is the missing Access-Control-Allow-Origin, not us refusing to answer.

Unsafe methods are the separate branch below, and those we still reject; with a 403, a body and no-store, rather than the empty 200 that was getting cached.

We could also do something like this if it makes the intentions clearer:

```php
$isPhpNetOrigin = $hostname !== "" && (
        preg_match('/^(.+\.)?php\.net$/', $hostname) === 1
        || $hostname === ($_SERVER["SERVER_NAME"] ?? "")
    );

if (!$isPhpNetOrigin) {
    // Safe methods are served as usual, only without CORS headers.
    if (in_array($_SERVER["REQUEST_METHOD"] ?? "GET", ["GET", "HEAD", "OPTIONS"], true)) {
        return;
    }

    // Unsafe methods are a different matter: a browser will not let a cross
    // site form forge or omit Origin
    header("Cache-Control: no-store");
    http_response_code(403);
    exit("Cross-origin requests are not allowed for this endpoint.\n");
}
```

PR #1980, opened 2026-07-30
Add PHP Playground

jordikroon:
I have been working on this as a side-project a while back. However as work was done on the redesign I decided not to postpone completing this as it would yet another thing.

Now that that has settled a bit, I decided to give it another go. As I think PHP could really benefit from having one.
Please give it a good look. The examples live in `js/playground.js`. Feel free to drop suggestions for those if you think something is missing.

image


jordikroon:
Moved this back to Draft until the preview branch is done. After that I am planning to rebase this and do some extra testing. Feel free to leave your initial thoughts regarding this PR.

PR #1981, opened 2026-08-03
Fix missing “Copy” button on the Windows download page

LamentXU123:
Add Copy buttons next to SHA256 checksums on download-related pages. Follow-up #1957. Now not only in the pre-realese page do we have this button after sha1 hash.

cc @sy-records

### DIFF
--- a/releases/index.php
+++ b/releases/index.php
@@ -233,20 +233,12 @@ function mk_rel(int $major,
             if (!isset($src["filename"])) {
                 continue;
             }
-            printf('<li><a href="https://museum.php.net/php%d/%s">%s</a>',
+            printf('<li><a href="https://museum.php.net/php%d/%s">%s</a></li>',
                    $major, $src["filename"], $src["name"]);
-            if (isset($src['sha256'])) {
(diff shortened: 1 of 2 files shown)


releases/index.php
```diff
@@ -233,12 +233,20 @@ function mk_rel(int $major,
             if (!isset($src["filename"])) {
                 continue;
             }
-            printf('<li><a href="https://museum.php.net/php%d/%s">%s</a></li>',
```
sy-records, line 236: These don't have SHA-256 hashes, only MD5, so the changes made here are unnecessary and need to be reverted.

sy-records (changes requested):

releases/index.php
LamentXU123: I dont see the reason why we only provide a button in case of SHA1 but not MD5.

sy-records: Most of them are SHA-256; I've already removed MD5 from my PR. These MD5 values may simply be legacy issues.

LamentXU123: Oh I see. Fair point :)

sy-records (approved):

PR #1985, opened 2026-08-12
Fix include paths in supported versions visualization

cookieguru:
Currently, https://www.php.net/supported-versions.php is broken: the versions table exists but the calendar and everything below it is missing.  Running locally, we can see it is the result of a fatal error:

Screenshot of broken page showing error text

With this patch, we can see the vizualization, body text, and footer render correctly:

Screenshot of working page with no errors

This was likely broken in b24d3106d45cc39e34ae0055386a1ae387019a20


sy-records:
Fixed via 26b50b74acf74e10646804f2e31866c199eb9bfd

PR #1992, opened 2026-08-19
Fix test failed

sy-records:


include/release-qa.php
```diff
@@ -174,10 +190,14 @@
     // Sorted information for later use
     // @todo need these?
     // $QA_RELEASES['releases']   : All current versions with active qa releases
```
jordikroon: The commented variable needs to be lowercased as well.

jordikroon (changes requested):

include/release-qa.php
sy-records: done

PR #2018, opened 2026-09-14
Set Cache-Control to no-store on paths containing %2F/%5C/%2E

svpernova09:
We currently have a CDN issue that can poison our caches because of how the CDN creates cache keys from paths, resulting in `/` on `www.php.net` being cached as a redirect.

This patch catches those requests and sets `Cache-Control: no-store` header so they won't be cached, but other legitimate redirects or 404s will be.

Once this lands and is tested, we can remove the current workaround edge rules, which set cache time to 0 on all 301, 302, and 404 response codes.

I have opened a support ticket with Bunny to address the issue.


include/prepend.inc
```diff
@@ -11,6 +11,19 @@ require_once __DIR__ . '/../src/autoload.php';
...
+// under a different URL (www.php.net "/" was poisoned this way in 2026-09).
+// Keep serving these requests, but never let the response be cached.
+(function (): void {
+    $path = explode('?', $_SERVER['REQUEST_URI'] ?? '/', 2)[0];
+    if (preg_match('~%(?:2f|5c|2e)|\\\\|(?:^|/)\.{1,2}(?:/|$)~i', $path)) {
+        header_register_callback(static function (): void {
+            header('Cache-Control: no-store');
+        });
+    }
+})();
```
derickr, line 25: I don't understand why there is a closure here that you then need to execute immediately, nor why the `header_register` is there, considering that literally the next line in the code also sets a header.

derickr (changes requested):

include/prepend.inc
svpernova09: It was an attempt at encapsulation, and I went a bit overboard. 

We *shouldn't* need this change if Bunny resolves the issue on their side. According to the support ticket, they've escalated it on their side, so this PR can wait until we see the outcome before moving forward.

svpernova09:
The is no longer needed; it's being addressed upstream, and our current solution is their recommendation until they resolve the parsing issue.

